Issue #5, opened 2016-08-28
Support kwargs for with_query


asvetlov:
`url.with_query(a='1', b='2')`

artyomboyko:
@asvetlov should query & kwargs be merged ? should one have priority over another ?

asvetlov:
No, just total replacement

asvetlov:
Fixed by #10

PR #10, opened 2016-09-08
add kwargs support for with_query

artyomboyko:
#5


tests/test_url.py
```diff
@@ -647,6 +647,11 @@ def test_with_query():
     assert str(url.with_query({'a': '1'})) == 'http://example.com/?a=1'
 
 
+def test_with_query_kwargs():
```
asvetlov: test should demonstrate query replacement, not adding a query to empty querystring.

yarl/__init__.py
```diff
@@ -557,16 +557,20 @@ def with_port(self, port):
                                                                port)),
                    encoded=True)
 
-    def with_query(self, query):
+    def with_query(self, query=None, **kwargs):
```
asvetlov: This prevents from using query as key: `url.with_query(query='val1', subquery='val2')`.
Please change signature to `def with_query(self, *args, **kwargs):` and check for `args` and `kwargs` mutual exclusiveness.
`len(args)` should be 1 if any.

docs/api.rst
```diff
@@ -431,6 +431,8 @@ section generates a new *URL* instance.
    :class:`~multidict.MultiDict` instances) or :class:`str`,
    autoencode the argument if needed.
 
+   It also can take an arbitrary number of keyword arguments.
+
```
asvetlov: No, the sentence is not enough.

Please change method's signature to
```
   .. method:: URL.with_query(query)
                     URL.with_query(**kwargs)
```

tests/test_url.py
```diff
@@ -647,6 +647,14 @@ def test_with_query():
...
+    url = URL('http://example.com')
+    assert str(url.with_query(
+        {'a': '2', 'b': '4'}, a='1')) == 'http://example.com/?a=1'
+    assert 'query2=1' in str(url.with_query(query='1', query2='1'))
```
asvetlov: Change checks to `str(...) == '...'` form.
`in` is too weak and not obvious comparing to all other tests for the library.

yarl/__init__.py
```diff
@@ -557,16 +557,26 @@ def with_port(self, port):
...
 
         """
         # N.B. doesn't cleanup query/fragment
+
+        if kwargs:
+            query = kwargs
+        elif len(args) == 1:
+            query = args[0]
+        else:
+            raise TypeError("Invalid query type")
```
asvetlov: It's `ValueError` actually.
Also please improve the message -- it should have a clue for proper format.

tests/test_url.py
artyomboyko: @asvetlov because kwargs doesn't have ordering, I'm not sure what order variables will be in the end. So I'm not sure how url will look like in the end. is this wrong ?

It looks like

```
http://example.com/?query2=1&query=1
```

But I'm not sure why, on other machines it can look ```http://example.com/?query=1&query2=1```

yarl/__init__.py
```diff
@@ -557,16 +557,26 @@ def with_port(self, port):
...
 
+        It also can take an arbitrary number of keyword arguments.
+
         Clear query if None is passed.
 
         """
         # N.B. doesn't cleanup query/fragment
+
+        if kwargs:
+            query = kwargs
```
asvetlov: Should check for empty args, generate an exception otherwise.

tests/test_url.py
asvetlov: Aaah, sorry.
I've forgot about hash randomization.

yarl/__init__.py
asvetlov: Say again, add explicit check for empty `args` here.
Otherwise user may call `url.with_query('a=b', c='d')` with confusing behaviour.

asvetlov:
Looks good (after fixing last comment) but please add tests for raising coverage back to 100%.

tests/test_url.py
```diff
@@ -649,8 +649,13 @@ def test_with_query():
 
 def test_with_query_kwargs():
     url = URL('http://example.com')
-    assert str(url.with_query(
-        {'a': '2', 'b': '4'}, a='1')) == 'http://example.com/?a=1'
+    with pytest.raises(ValueError):
```
asvetlov: Please split a test: extract every `with pytest.raises(...)` into separate test function.

tests/test_url.py
```diff
@@ -649,8 +649,13 @@ def test_with_query():
...
+        url.with_query(
+            {'a': '2', 'b': '4'}, a='1')
+    with pytest.raises(TypeError):
+        url.with_query(b=3)
+    with pytest.raises(ValueError):
+        url.with_query('a=1', 'a=b')
     assert 'query2=1' in str(url.with_query(query='1', query2='1'))
```
asvetlov: Hmm. Returning to hash randomization: you can check not `str(url)` but `url.query == {query='1', query2='1'}`.
It should work.
Please try this way.

artyomboyko: @asvetlov I'm still getting them in the wrong order..
```
(Pdb) url.with_query(query='1', query2='1').query
<MultiDictProxy('query2': '1', 'query': '1')>
```

asvetlov:
Thank you but what I required from tests is here: ea61d81

artyomboyko:
@asvetlov thank you. yes, it's much more beautiful this way

PR #11, opened 2016-09-17
Fix reference to method in docs

mgedmin:
This looks wrong (at http://yarl.readthedocs.io/en/latest/):
ekrano nuotrauka is 2016-09-17 10-12-22
and I hope I got the Sphinx syntax right to fix it.


asvetlov:
Thanks!

PR #30, opened 2016-11-25
Load pytest-runner only if mandatory

noplay:
This PR load pytest runner only if need. Without that you need an internet access to execute setup.py install even if the dependencies are already installed.

The code for detecting if require is from section Conditional Requirement:
https://pypi.python.org/pypi/pytest-runner


asvetlov:
thanks

PR #54, opened 2017-02-08
Drop spurious print

bersace:
Hi,

There is a spurious debug print. I suggest to drop it.

Thanks for yarl.


fafhrd91:
my mistake

fafhrd91:
0.9.2 is released

PR #57, opened 2017-02-16
Feature/kwargs constructor

mosquito:
kwargs constructor for create a new url instance.

```python
u = URL(
        scheme='http',
        host='127.0.0.1',
        user='foo',
        password='bar',
        port=8000,
        path='/index.html',
        query_string="arg=value1",
        fragment="top"
    )

assert str(u) == 'http://foo:bar@127.0.0.1:8000/index.html?arg=value1#top'
```

`scheme` and `host` is required in this case


mosquito:
@fafhrd91 please review.

dkhutornoy:
We want this basically to use when there are known domain and protocol: `url = URL(host=domain, scheme=protocol)`. With current API we do: `url = URL('://'.join([protocol, domain])).with_path(...)`, since `url = URL(domain).with_scheme(protocol)` raises error.

PR #58, opened 2017-02-16
build classmethod

mosquito:
kwargs constructor for create a new url instance.

```python
u = URL.build(
    scheme='http',
    host='127.0.0.1',
    user='foo',
    password='bar',
    port=8000,
    path='/index.html',
    query_string="arg=value1",
    fragment="top"
)

assert str(u) == 'http://foo:bar@127.0.0.1:8000/index.html?arg=value1#top'
```

`scheme` and `host` is required in this case

new version of #57


mosquito:
@asvetlov I totally rebase and reopen #57. Please review.

yarl/__init__.py
```diff
@@ -192,6 +192,23 @@ def __init__(self, val='', *, encoded=False, strict=False):
...
         self._cache = {}
 
+    @classmethod
+    def build(cls, scheme='', user='', password='', host='', port=None, path='',
+              query_string='', fragment='', *, strict=False, encoded=False):
+
+        assert scheme and host, "You must specify scheme and host"
+
+        return cls(
+            urlunsplit(
```
asvetlov: Implementation looks unnecessary slow and maybe error-prone.

I suggest just passing SplitResult without unsplitting.

mosquito: `SplitResult` has been removed.
`urlunsplit` implementation is so simple and clear (just string concatenation and condition checks)

asvetlov:
No.
Concatenating and parsing result back in `__init__` is very bad and slow idea.

mosquito:
@asvetlov I made tests for it.

First of all I rewrote it like this:
```python
@classmethod
def build(cls, scheme='', user='', password='', host='', port=None, path='',
          query_string='', fragment='', *, strict=False, encoded=False):

    assert scheme and host, "You must specify scheme and host"

    url = cls('%s://%s' % (scheme, host), strict=strict, encoded=encoded)

    if user:
        url = url.with_user(user)

    if password:
        url = url.with_password(password)

    if port:
        url = url.with_port(port)

    if path:
        url = url.with_path(path)

    if query_string:
        if isinstance(query_string, str):
            url = url.with_query(query_string)
        elif isinstance(query_string, dict):
            url = url.with_query(**query_string)
        else:
            url = url.with_query(*query_string)

    if fragment:
        url = url.with_fragment(fragment)

    return url
```

I run ipython and run `%timeit`

```python
In [8]: %timeit URL.build(host="example.net", scheme="http", user="foo", password="bar", fragment="test", query_string="aaa=1")
10000 loops, best of 3: 61.6 µs per loop
```

after that I return `urlunsplit` based implementaion:
```python
@classmethod
def build(cls, scheme='', user='', password='', host='', port=None, path='',
          query_string='', fragment='', *, strict=False, encoded=False):

    assert scheme and host, "You must specify scheme and host"

    return cls(
        urlunsplit((
            scheme, cls._make_netloc(user, password, host, port), path, query_string, fragment
        )),
        strict=strict,
        encoded=encoded
    )
```

ipython's `%timeit` results is:
```python
In [2]: %timeit URL.build(host="example.net", scheme="http", user="foo", password="bar", fragment="test", query_string="aaa=1")
10000 loops, best of 3: 28.7 µs per loop
```

mosquito:
@asvetlov please assist me, what should I do? What changes should be applied?

asvetlov:
Try constructing `SplitResult` as you did in first version and passing this object into `yarl.URL` with `encoded=True`.

mosquito:
@asvetlov you absolutely right. It's better.

```python
In [3]: %timeit yarl.URL.build(scheme="http", host="foo", path="/")
The slowest run took 34.88 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 2.8 µs per loop
```

asvetlov:
Cool!
Please mention new method in `api.rst`

mosquito:
@asvetlov done

docs/api.rst
```diff
@@ -429,7 +450,7 @@ section generates a new *URL* instance.
 
    .. doctest::
 
-      >>> URL('http://example.com/'.with_path('/path/to')
+      >>> URL('http://example.com/').with_path('/path/to')
```
mosquito: correction for doctests

asvetlov:
There are two questions.
1. Make arguments keyword-only maybe? `URL.build('https', 'guest', 'secret', 'example.com', 1443, '/admin/')` looks messy.
2. `query` parameter might make sense (should be mutually exclusive with `query_string`).

mosquito:
1. Done
2. Done

mosquito:
Was rebased with `master`

mosquito:
@asvetlov any comments?

yarl/__init__.py
```diff
@@ -192,6 +192,45 @@ def __init__(self, val='', *, encoded=False, strict=False):
         self._val = val
         self._cache = {}
 
+    @classmethod
+    def build(cls, *, scheme='', user='', password='', host='', port=None, path='',
+              query=None, query_string='', fragment='', strict=False):
+
+        """ Creates and returns a new URL
+
+        :type scheme: str
```
asvetlov: Pls remove these lines, we don't use sphinx autodoc for the project.

Docstring hould be just 
```
"""Creates and returns a new URL.
"""
```
(no leading space and period at the end).
Just look on other docstrings from the module.

yarl/__init__.py
```diff
@@ -192,6 +192,45 @@ def __init__(self, val='', *, encoded=False, strict=False):
...
+        :type password: str
+        :type host: str
+        :type port: int
+        :type path: str
+        :type query_string: str
+        :type fragment: str
+        :type strict: str
+        """
+        if host or scheme:
+            assert scheme, 'Can\'t build URL with "host" but without "scheme".'
```
asvetlov: Please replace assertions with raising `ValueError`.

yarl/__init__.py
```diff
@@ -192,6 +192,45 @@ def __init__(self, val='', *, encoded=False, strict=False):
...
+            assert host, 'Can\'t build URL with "scheme" but without "host".'
+
+        assert not (query and query_string), "Only one of \"query\" or \"query_string\" should be passed"
```
asvetlov: `assert` -> `ValueError`

mosquito:
@asvetlov can you merge it?

fafhrd91:
I would add quoting

something like https://github.com/fafhrd91/aiohttp/blob/master/aiohttp/http_message.py#L400

mosquito:
@fafhrd91 please add your example as a comment. Your link is obsolete.

fafhrd91:
you need to quote at least path and query string

fafhrd91:
here is proper link https://github.com/aio-libs/aiohttp/blob/master/aiohttp/http_writer.py#L316

fafhrd91:
could you add quoting for path and query, then I will merge

mosquito:
@fafhrd91 done. Actually not needed apply quoting to query string, but if you want I already made it.

fafhrd91:
why do you think we should not quote query string?

mosquito:
@fafhrd91 because I forgot replace `query_sting`. Sorry and thank you.

fafhrd91:
I think `quote` is misleading name

mosquito:
@fafhrd91 should I make any changes?

fafhrd91:
I am not sure. it is in the wild, it would be backward incompatible

PR #74, opened 2017-05-03
Do not quote or unquote + if not a query string.

ericsnowcurrently:
"+" is entirely legal in URL paths.  Unfortunately unquote() does the wrong thing.  This patch fixes that.  It also adds a test for quote(qs=False) for "+", to be sure it continues to do the right thing.


fafhrd91:
thanks

PR #75, opened 2017-05-04
Unexpected hash behaviour

hellysmile:


fafhrd91:
could you add entry to changelog

fafhrd91:
thanks!

PR #83, opened 2017-06-12
Prevent double unquoting

Deepwalker:
In current version of yarl we have problem with query args unquoted twice. So If you will pass URL as argument, it will be broken. This PR fixes the problem.


asvetlov:
Thanks!

Issue #91, opened 2017-07-13
Encode semicolon in query string


asvetlov:
aio-libs/aiohttp#1919 aio-libs/aiohttp#1901

asvetlov:
Fixed by #104

PR #104, opened 2017-08-24
Unsafe encoding for QS fixed. Encode `;` char in value param

pfreixes:
Current behaviour of YARL does not encode the `;` parameter when it is present in the QS param value. The `;` is not a safe parameter and has to be encoded [1].

The following snippet shows the error (raised via aiohttp):

```python
from yarl import URL
url = URL("http://www.google.com").with_fragment(None)
params = {'foo': 'a;b'}
url2 = url.with_query(params)
url2.query
```

This snippet gets as an output an "invalid" MultiDict

```
<MultiDictProxy('foo': 'a', 'b': '')>
``` 

[1] https://tools.ietf.org/html/rfc3986#section-2.2

### DIFF
--- a/tests/test_update_query.py
+++ b/tests/test_update_query.py
@@ -113,10 +113,10 @@ def test_with_multidict_with_spaces_and_non_ascii():
 
 def test_with_query_multidict_with_unsafe():
     url = URL('http://example.com/path')
-    url2 = url.with_query({'a+b': '?=+&'})
-    assert url2.raw_query_string == 'a%2Bb=?%3D%2B%26'
-    assert url2.query_string == 'a%2Bb=?%3D%2B%26'
-    assert url2.query == {'a+b': '?=+&'}
+    url2 = url.with_query({'a+b': '?=+&;'})
+    assert url2.raw_query_string == 'a%2Bb=?%3D%2B%26%3B'
+    assert url2.query_string == 'a%2Bb=?%3D%2B%26%3B'
+    assert url2.query == {'a+b': '?=+&;'}
 
 
 def test_with_query_None():
--- a/yarl/__init__.py
+++ b/yarl/__init__.py
@@ -760,7 +760,7 @@ class URL:
                 else:
                     raise TypeError("Invalid variable type: mapping value "
                                     "should be str or int, got {!r}".format(v))
-                lst.append(quoter(k, safe='/?:@')+'='+quoter(v, safe='/?:@;'))
+                lst.append(quoter(k, safe='/?:@')+'='+quoter(v, safe='/?:@'))
                 query = '&'.join(lst)
         elif isinstance(query, str):
             query = _quote(query, safe='/?:@',


pfreixes:
@asvetlov can you take a look to it? if the bug is confirmed this would require a fix release ASAP.

asvetlov:
The change is a step in right direction but it's not enough.

`URL.query` uses `parse_qsl`: https://github.com/aio-libs/yarl/blob/master/yarl/__init__.py#L474
which in turn uses `;` as separator along with `&`.

pfreixes:
Hi @asvetlov, 

That's the idea, if `parse_qsl` uses the `;` as a separator, dont make this character safe. From my understanding right now with this change, the behaviour is consistent.

it's up to the user give the raw URL properly encoded :

```python
>>> url = URL("http://www.google.com?foo=a%3bb")
>>> url.query
<MultiDictProxy('foo': 'a;b')>
```

Or it's up to the YARL module to encode the `;` when the params are given explicitly by the user to build a URL:

```python
>>> URL("http://www.google.com").with_query({'foo': 'a;b'}).query
<MultiDictProxy('foo': 'a;b')>
```

am I missing something?

pfreixes:
More info.

The default implementation of the `urlencode` uses the `quote` function putting the `;` as a not safe parameter, having as a result:

```python
from urllib.parse import urlencode
urlencode({'foo':'a;b'})
'foo=a%3Bb' 
```

From my understanding, the `urllib` is working consistently. Having, as a result, a bi-directional operationality:

```python
>>> from urllib.parse import parse_qs
>>> parse_qs('foo=a%3Bb')
{'foo': ['a;b']}
```

pfreixes:
Tries to address the #91 issue.

asvetlov:
Hi. Honestly I don't know how to process it better.
`;` was discouraged by W3C in last edition (2014 IIRC).
Perhaps we shouldn't process it as parameters separator, it is a very weird inheritance from SGML era.
Replacing `parse_qsl` with custom implementation sounds reasonable to me, python's version does many   dances around str/bytes etc BTW.

Regarding to `;` quoting: RFC 3986 allows this symbol in query part without percent encoding. I think Python/requests doesn't behave right in this case.

asvetlov:
Well, it should be encoded in cases like `url.with_query({'foo': 'a;b'})` but I not sure for `url.with_query('foo=a;b')`.

pfreixes:
Thanks! any chance to have a patched release?

Regarding the RFC and W3C emphasis, well explained here [1] (for future readers)

Regarding the inconsistency with `url.with_query({'foo': 'a;b'})` vs `url.with_query('foo=a;b')`, from my understanding the second one must not encode the `;`.

[1] https://www.w3.org/TR/1999/REC-html401-19991224/appendix/notes.html#h-B.2.2

asvetlov:
I'd like review other yarl-related issued in aiohttp bug tracker and publish a new release just after checking that all mentioned problems has fixed.

PR #117, opened 2017-10-11
Fix make commands and flake errors, hotfix CI

alxpy:
FYI @kxepal

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,19 +3,21 @@ MD = $(shell python -c "import multidict; print(multidict.__path__[0])")
 all: test
 
 
-.develop: $(shell find yarl -type f)
-	@pip install -e .
+.develop: requirements-dev.txt
+	pip install -U -r requirements-dev.txt
 	@touch .develop
 
-test: .develop mypy
-	pytest ./tests --flake8
 
-vtest: .develop mypy
-	pytest ./tests --flake8 -v
+test: .develop
+	pytest ./tests ./yarl --flake8
 
 
-cov: .develop mypy
-	pytest --cov yarl --cov-report html --cov-report term ./tests/ --flake8
+vtest: .develop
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -7,9 +7,15 @@ all: test
 	@pip install -e .
 	@touch .develop
 
+
+flake: .develop
+	flake8 yarl
```
kxepal: I think this should be run as a part of `make test` to simplify development process. And no need to intorduce another target since pytest uses flake8 and can run it against regular sources.

kxepal: mmm...by looking on `.travis.yml` and `Makefile` it seems that there is need to make some refactoring to achieve that.

 @asvetlov, what do you think?

alxpy: During the tests run we check flake errors only at tests folder:
```
test: .develop
	pytest ./tests --flake8
```
Maybe it would be better:
```
test: .develop
	flake8 yarl tests
	pytest ./tests
```

kxepal: I think it would be fine to just run `pytest --flake8` - it will run both tests and flake8 on everything in directory: yarl, setup.py, tests and the rest bits. That shouldn't hurt. If you use `pytest-sugar`, the result report will be pretty nice.

alxpy:
I updated the PR:
- fix flake errors
- fix make commad
- [hotfix] add flake check to CI

Makefile
alxpy: I did not notice your message =(

asvetlov:
@kxepal I think the PR is ready for merging

kxepal:
Yes, it is.

alxpy:
@kxepal  can you merge it?)

kxepal:
🎉 
@asvetlov I hope you don't mind (;

Issue #111, opened 2017-09-27
No way to replace path without replacing query


Lucretiel:
The following is unexpected behavior:

```
url1 = URL("http://www.example.com/foo?bar=10")
url2 = url1.with_path("/bar")

assert url1.query = url2.query  # Assertion fails
```

While I understand the design decision to treat the query and fragment components as part of the path, this has left us without an (efficient) way to preserve the query and fragment components when replacing just the path. The only way to do it is:

```
url2 = url1.with_path(new_path).with_query(url1.query_string).with_fragment(url1.fragment)
```

This is a total of 4 extra method calls, at least of which does a lot of unnecessary processing. I'm assuming that there's code out there that depends on this weird behavior, so I'd like to propose and, with permission, implement a new method, `with_just_path`, which replaces the path but leaves the fragment and query the same.

asvetlov:
Did you try `URL.build`?

Lucretiel:
I mean, that would work, but why even have the `with_*` methods in the first place? This seems like it'd fit in well with the other `with` methods.

asvetlov:
Honestly I don't want to blow up URL class with new methods.
Also I suspect your need is very rare: usually query/fragment is described by path, different paths have different allowed query params.
The only exception is passing in query some constant information like security token, it could be done by `url.update_query()` now.

Lucretiel:
My need is in use for a web framework, which is why I'm interested in reducing the amount of unnecessary work implied by `with_query`. In particular, I'm working on django-style path routing where the front of a path is cut off when it's passed to child route views. It's not exactly a very common use case, but I can imagine it'd be common for creating web frameworks based on yarl.

asvetlov:
Thus do you need cutting first path symbols from relative URL?
Make sense.
Maybe create a special method for this case?

eternal-sorrow:
Then `URL.with_host()` should cut off the path, query and fragment altogether. Why the inconsistency? It doesn't make sense.
```python
>>> URL('https://foo.tld/path?a=1#fragment').with_host('bar.tld')
URL('https://bar.tld/path?a=1#fragment')
```

apalala:
It is totally unexpected that `.with_path()` will remove `query`.

JP-Ellis:
Working with this library, and got this unexpected behaviour as well.

Is there any appetite for a PR?

trim21:
is this consider a bug or by design？

asvetlov:
It is a design decision.
[...]

PR #119, opened 2017-10-19
Allow subclassing by removing hard calls to URL()

twidi:
One of the power of yarl is to easily create a new object based on
another.

This feature was not usable by subclasses because all methods doing this
returned a `URL` object, hard-coded.

This commit use `self.__class__` instead, to return a new object of the
same class as the original one.


asvetlov:
It's easy change but hard design question.
Now, when I sure nobody subclass from `URL`, I can do deep refactoring without fear to freak some code.

What is your use case for subclassing?

twidi:
Hi, thanks for your answer.

In a recent comment about an issue, you said

> Honestly I don't want to blow up URL class with new methods. Also I suspect your need is very rare

It's COMPLETELY justified, and it was a really good answer. We expect this in libraries, it's part of the game.

But then, you must let users play with the library, at their own risk. You provide a public API, and as long as you don't break this (thanks semver), you should not worry about refactoring internals. Using `self.__class__` instead of hard-coded `URL` should not change that.

If the user uses internal methods, ie not the documented API, it's its choice, and it can use things like [raincoat](https://pypi.python.org/pypi/raincoat) to be notified.

About my use case, it's simply an override of `update_query` (with more things than in your branch on the subject), but it could be anything, really.

Also, the code to actually make it work is a lot more likely to break, because I have to override every method returning `URL`.

In conclusion: you cannot (and, imho, you should not) prevent user to play as they want with your lib because no matter what, they'll do it. The lib can just be a little helpful in this path.

But maybe you see another way for that?

asvetlov:
`yarl.URL` is not intended to be inherited from. In C# it should be `sealed`, in Java -- `final`.
That's final decision.
Like you should never try to inherit from `pathlib.Path` or `asyncio._SelectorSslTransport`.
I'd like to keep full control over the class.
Maybe later URL will be rewritten with Cython for sake of speed. Inheritance support prevents many future optimizations.

If you need a wierd/custom `url.update_query()` -- please make `update_query(query)` function, not modify `yarl`.
If `update_query()` is buggy -- please report and fix it in the library itself.

twidi:
I think it's a bad decision, but it's your call.

Note that i'm not modifying `yarl`, I override it. It's a common practice in the python world.

And about sealing... we're talking about a class to manage urls, here, not to control a nuclear power plant :) 

Keep up the good work! I still love this library. I may fork it and release it with the sublcassing being possible, though.

kxepal:
But still. What's your case, @twidi ? Except of having possibility to do subclass.

twidi:
As i said, I updated `update_query` to handle multiple keys, but it's not as simple as its sounds:

- if a key already exists, should a replace the value(s) with the new one(s) or add them
- if I add them, what about a value that already exists
- how to remove a key

So I created my own way to do this.

One could also want to create its own method to handle #111 for example.

asvetlov:
aiohttp creates `yarl.URL` in many cases.
If part of processed URLs will have patched of `update_query()` but other will not -- the overall design is very subtle and error prone.

The same for #111 In practice caller should always check `hasattr(url, 'new_method')` before calling it. 

BTW I have intention for fixing `update_query` behavior to support multiple keys but should fix `MultiDict.update()` first.

twidi:
I know about `update_query`, I saw your work.

About `aiohttp`, one could want to use `yarl` outside of the `aiohttp` context. If it is not allowed, it should be clearly mentioned.

And even inside the `aiohttp` context, like I do, I may want to compose my URL the way I want.

And when you said 

> If part of processed URLs will have patched of update_query() but other will not

I cannot imagine why `aiohttp` would create another url based on the one I give us. If it's a internally created object, it would be a `URL`, not one of my own subclass, so no problem here.

Finally I wanted to say this: Let's be clear: I totally understand your point of view. I just don't share it :) Let's agree to disagree ;)

And good lock about `update` and `update_query`, it's a more complex subject than it sounds at first view :) I'll follow your work on this subject, it's interesting.

asvetlov:
Ok, do clone if you want -- it's Open Source!
I agree with your disagree :)

PR #120, opened 2017-10-19
[typing] with_query & update_query can accept ints

twidi:
In `__init__.pyi`, in the versions of `with_query` and `update_query`
accepting a `Mapping`, a `Sequence[Tuple]` or `**kwargs`, the
expected values for the querystring are marked as being a `str`, but
the code indicate clearly that also `int` are accepted.

This commit changes that by replacing `str` by `Union[str, int]`


jettify:
Changes make sense, quick question are you running mypy in your project with yarl or just noticed those issues in stubs?

twidi:
I use mypy in my project with yarl :) But I saw the problem even before that, with pycharm which works great with python annotations!

jettify:
Can you please rebase this branch?

twidi:
done!

PR #121, opened 2017-10-19
[typing] Correct wrong arg name for __truediv__

twidi:
In the file `__init__.pyi`, the argument wass named `str` but it's `name`
in the code.

This commits fixes that.


jettify:
thanks!

PR #131, opened 2017-11-09
Percent-encoding for userinfo in _make_netloc classmethod

nuno-andre:
According to [RFC 3986](https://tools.ietf.org/html/rfc3986#section-3.2.1) (percent-encoding) and [RFC 2617](https://tools.ietf.org/html/rfc2617#section-2) (colon is safe only for username).

### DIFF
--- a/yarl/__init__.py
+++ b/yarl/__init__.py
@@ -597,6 +597,8 @@ class URL:
         if password:
             if not user:
                 user = ''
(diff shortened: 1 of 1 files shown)


asvetlov:
Please fix broken tests and add a new one for your case.

asvetlov:
It is implemented already, user and password are percent encoded.

Issue #141, opened 2017-12-07
Using `with_query` with an empty mapping causes `query_string` to raise a `TypeError`


CCP-Aporia:
Python Version: 3.6.1
OS: Windows 10

Using `with_query` with an empty mapping causes the returned URL instance to contain an invalid `query_string` property that raises a `TypeError` when it is being accessed.

Example code:
```
import yarl
a = yarl.URL("http://www.example.com")
print(a.with_query(a.query).query_string)  
# expected output: "" 
# actual output: "TypeError: Argument should be str"
```
The culprit seems to be that 

Looking in `_get_str_query` it appears the easiest fix would be to replace:
```
# ... snip ...
if query is None:
  query = ''
# ... snip ...
```
with the following code:
```
# ... snip ...
if not query:
  query = ''
# ... snip ...
```

The relevant test case - `test_update_query.py:test_with_query_empty_dict` - should be updated like this:
```
def test_with_query_empty_dict():
    url = URL('http://example.com/?a=b')
    new_url = url.with_query({})
    assert new_url.query_string == ''
    assert str(new_url) == 'http://example.com/'
```

A pull request with the fix is on the way.

asvetlov:
Found proper fix for the issue: 35415a1
Thanks for raising up the regression and writing a test.

CCP-Aporia:
Oh yeah, that's the obvious one. A lot more elegant than mine! :-)

asvetlov:
Anyway release 0.16 is scheduled for deployment.
Will be published on PyPI in an hour or something like this.

PR #142, opened 2017-12-07
Fix handling of emtpy mappings in with_query

CCP-Aporia:
Fixes aio-libs/yarl#141

### DIFF
--- a/yarl/__init__.py
+++ b/yarl/__init__.py
@@ -745,7 +745,7 @@ class URL:
             raise ValueError("Either kwargs or single query parameter "
                              "must be present")
 
-        if query is None:
+        if not query:
             query = ''
         elif isinstance(query, Mapping):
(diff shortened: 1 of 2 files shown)


yarl/__init__.py
```diff
@@ -745,7 +745,7 @@ def _get_str_query(self, *args, **kwargs):
             raise ValueError("Either kwargs or single query parameter "
                              "must be present")
 
-        if query is None:
+        if not query:
```
asvetlov: Please no.
Your change allows to pass `url.with_query(0)` or `url.with_query(False)`.
Better to add a check for empty mapping 3 lines below.

CCP-Aporia: Fair point, pushing a change.

asvetlov:
Thanks!

PR #144, opened 2017-12-14
Fix IPv6 zone handling, lowercase in domain name and compressed ip ad…

socketpair:
…dresses.

### DIFF
--- a/yarl/__init__.py
+++ b/yarl/__init__.py
@@ -160,16 +160,16 @@ class URL:
                 if netloc is None:
                     raise ValueError(
                         "Invalid URL: host is required for abolute urls.")
-                try:
-                    netloc.encode('ascii')
-                except UnicodeEncodeError:
-                    netloc = netloc.encode('idna').decode('ascii')
                 else:
+                    (ip, *zone) = netloc.split('%', 1)
                     try:
-                        ip = ip_address(netloc)
+                        ip = ip_address(ip)
                     except ValueError:
-                        pass
+                        netloc = netloc.encode('idna').decode('ascii')
                     else:
+                        netloc = ip.compressed
+                        if zone:
+                            netloc += '%' + zone[0]
                         if ip.version == 6:
                             netloc = '[' + netloc + ']'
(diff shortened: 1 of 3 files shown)


yarl/__init__.py
```diff
@@ -160,16 +160,16 @@ def __init__(self, val='', *, encoded=False, strict=None):
...
-                except UnicodeEncodeError:
-                    netloc = netloc.encode('idna').decode('ascii')
                 else:
+                    (ip, *zone) = netloc.split('%', 1)
                     try:
-                        ip = ip_address(netloc)
+                        ip = ip_address(ip)
                     except ValueError:
-                        pass
+                        netloc = netloc.encode('idna').decode('ascii')
```
socketpair:
```
In [9]: timeit 'http://lasdhglajkdshflkajsdfh.asdf.asdf.asdf.asdf.asdf.rtyty'.encode('ascii')
1000000 loops, best of 3: 281 ns per loop

In [10]: timeit 'http://lasdhglajkdshflkajsdfh.asdf.asdf.asdf.asdf.asdf.rtyty'.encode('idna').decode('ascii')
100000 loops, best of 3: 2.79 µs per loop

```

socketpair: Is it the case ?

asvetlov: Sorry, don't follow the question.

yarl/__init__.py
```diff
@@ -160,16 +160,16 @@ def __init__(self, val='', *, encoded=False, strict=None):
                 if netloc is None:
                     raise ValueError(
                         "Invalid URL: host is required for abolute urls.")
-                try:
-                    netloc.encode('ascii')
-                except UnicodeEncodeError:
-                    netloc = netloc.encode('idna').decode('ascii')
                 else:
+                    (ip, *zone) = netloc.split('%', 1)
```
asvetlov: use `.partition()`

asvetlov:
Merged by 7d8f58b

Issue #148, opened 2017-12-28
Germanic problems -- IDNA does not round-trip


wumpus:
```python
import aiohttp
import asyncio
import sys

hostname = None

async def fetch(session, url):
        async with session.get(url) as response:
            return await response.text()

async def main():
    async with aiohttp.ClientSession() as session:
        html = await fetch(session, hostname)
        print(html)

hostname = sys.argv[1]
loop = asyncio.get_event_loop()
loop.run_until_complete(main())
```
```
$ python ./bug.py http://xn--einla-pqa/robots.txt
...
UnicodeError: ('IDNA does not round-trip', b'xn--einla-pqa', b'einlass')
```

wumpus:
```
UnicodeError: ('IDNA does not round-trip', b'xn--schnellimbi-56a', b'schnellimbiss')
UnicodeError: ('IDNA does not round-trip', b'xn--fubett-cta', b'fussbett')
UnicodeError: ('IDNA does not round-trip', b'xn--fleiig-eta', b'fleissig')
UnicodeError: ('IDNA does not round-trip', b'xn--fupflegegeraet-1fb', b'fusspflegegeraet')
UnicodeError: ('IDNA does not round-trip', b'xn--fupflegegeraet-1fb', b'fusspflegegeraet')
UnicodeError: ('IDNA does not round-trip', b'xn--fubett-cta', b'fussbett')
UnicodeError: ('IDNA does not round-trip', b'xn--metallverschluee-ulb', b'metallverschluesse')
UnicodeError: ('IDNA does not round-trip', b'xn--einla-pqa', b'einlass')
```

asvetlov:
The exception is raised by python internals on `b'xn--einla-pqa'.decode('idna')`.
The codec doesn't support error handling mode other than 'strict'.
Honestly I have no idea how to help you.

Quick googling give resources like https://stackoverflow.com/questions/9806036/idna-does-not-round-trip
It describes the problem but has no solution.

socketpair:
I have create bug report:
https://bugs.python.org/issue32437

asvetlov:
Well, https://pypi.python.org/pypi/idna supports IDNA2008 standard.
We could just replace builtin idna codec with the library.
The change is trivial.
@wumpus would you provide a Pull Request?

wumpus:
I can't figure out how to configure the yarl development environment, sorry.

You're correct that it's a fairly trivial change, 3 lines of code and 4 lines of test.

[...]

PR #149, opened 2017-12-29
update to idna 2008 and test

wumpus:
Discussed in https://github.com/aio-libs/yarl/issues/148

### DIFF
--- a/requirements/ci.txt
+++ b/requirements/ci.txt
@@ -3,4 +3,5 @@ wheel==0.30.0
 pytest-cov>=2.3.1
 pytest-flake8>=0.6
 sphinx==1.6.5
+idna==2.6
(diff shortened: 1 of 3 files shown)


wumpus:
Now that's an interesting failure!

URL('http://[::1]').host
idna.core.InvalidCodepoint: Codepoint U+003A at position 1 of '::1' not allowed

wumpus:
I am a little surprised to find that "make test" doesn't do the doctest, and if I fix the resulting bug, I'll fail patch code coverage because the doctest isn't run in the patch code coverage test!

asvetlov:
Do you mean doctests from `docs/api.rst`?
Testing them is good idea. Please make a separate request if you want.

PR #154, opened 2018-01-11
Optimization

asvetlov:
Use fast path if quoted string doesn't require requoting

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -101,3 +101,4 @@ yarl/_quoting.html
 .idea
 
 .mypy_cache
+.install-deps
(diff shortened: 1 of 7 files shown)


fafhrd91:
this is pretty cool

asvetlov:
In conjunction with https://github.com/aio-libs/yarl/pull/157 and https://github.com/aio-libs/yarl/pull/155 the time difference between Python and Cython is about 60 times :)

fafhrd91:
ported Quoter code to [rust](https://github.com/actix/actix-web/blob/master/src/uri.rs#L67)

want to port yarl to rust? :)

PR #171, opened 2018-02-18
Forbid inheritance, replace __init__ with __new__

asvetlov:
yarl.URL was always considered as **final** class.


yarl/__init__.py
```diff
@@ -225,6 +225,10 @@ def build(cls, *, scheme='', user='', password='', host='', port=None,
         else:
             return url
 
+    def __init_subclass__(cls):
+        raise TypeError("Inheritance a class {} from URL "
```
hellysmile: Seems here missed formattin `'{}'.format(cls)`

asvetlov: My fault.
Fixed.

PR #214, opened 2018-07-11
Replace \x usage for Unicode character escaping with \u

behnam:
Using `\xYZ' in Unicode string literals can be a source of confusion, as
`\xYZ` escpaing in commonly used for byte escaping and not character
escaping.

Using `\u00YZ` on the other hand, is more clear and won't be misleading
any reader about the intention.


asvetlov:
thanks

Issue #147, opened 2017-12-25
There should be a way to add a empty fragment with URL.with_fragment()


SzieberthAdam:
[According to the HTML5 spec](https://www.w3.org/TR/html5/single-page.html#an-indicated-part-of-the-document) the empty fragment `#` and the special fragment `#top` will link to the top of the page.

I would like to be able to create an URL with an empty fragment using the `URL.with_fragment()` method.

Currently both `URL.with_fragment(None)` and `URL.with_fragment("")` _removes_ the fragment from the URL.

I suggest that with an empty string parameter it should create an empty fragment `#`.

Thanks for aiohttp and this great library!

asvetlov:
I'm ok to `URL.with_fragment("")`.
Would you make a Pull Request?
The task looks easy.

SzieberthAdam:
Sure, just give me some time. Christmas here.

asvetlov:
Ping

PR #216, opened 2018-07-12
tests: Drop some xfails and add info for the rest

behnam:
Behavior of `test_none_value_for_query()`, marked previously as
`@pytest.mark.xfail`, seems to be expected behavior. Updated as such.

Add FIXME and reference for other `@pytest.mark.xfail` tests.

Also fix test function names and add new test for the one with test body
not doing what the name was saying.


tests/test_url.py
```diff
@@ -942,10 +942,9 @@ def test_empty_value_for_query():
     assert str(url) == 'http://example.com/path?a='
 
 
-@pytest.mark.xfail
 def test_none_value_for_query():
-    url = URL('http://example.com/path').with_query({'a': None})
-    assert str(url) == 'http://example.com/path?a'
+    with pytest.raises(TypeError):
+        URL('http://example.com/path').with_query({'a': None})
```
behnam: I changed this one since it looks like we have explicit check against `None` value. So, this is currently the expected behavior.

But, similar to GH-147 (None vs empty for fragment), I would like to add support for `None` value to query values, as well.

What do you think?

asvetlov: passing `None` for removing query/fragment entirely looks pretty good for me.
But `None` for dropping the only on query param is cumbersome and counter-intuitive.
Explicit manipulation is better I believe:
```
q = url.query
q = MultiDict(q)  # make it mutable
del q['key']
url.with_query(q)
```

tests/test_quoting.py
```diff
@@ -166,43 +166,48 @@ def test_unquoting(num, unquoter):
...
 
 
 @pytest.mark.xfail
-def test_unquoting_mixed_case(unquoter):
-    # Test unquoting on mixed-case hex digits in the percent-escapes
-    given = '%Ab%eA'
-    expect = '\u00ab\u00ea'
-    result = unquoter()(given)
-    assert expect == result
+# FIXME: After conversion to bytes, should not cause UTF-8 decode fail.
```
webknjaz, line 190: @behnam do you remember why you added this specific comment? I don't really know what to make of it…

cc @mjpieters @Dreamsorcerer any ideas what this could mean? This combo of `pytest.raises()` + `xfail` without a clearly stated reason is rather confusing...

Dreamsorcerer: Isn't it just repeating the note at the top of link?

> It is generally a good idea for sequences of [percent-encoded bytes](https://url.spec.whatwg.org/#percent-encoded-byte) to be such that, when [percent-decoded](https://url.spec.whatwg.org/#string-percent-decode) and then passed to [UTF-8 decode without BOM or fail](https://encoding.spec.whatwg.org/#utf-8-decode-without-bom-or-fail), they do not end up as failure.

Dreamsorcerer: Had another look. So, the spec says:
> Using anything but [UTF-8 decode without BOM](https://encoding.spec.whatwg.org/#utf-8-decode-without-bom) when input contains bytes that are not [ASCII bytes](https://infra.spec.whatwg.org/#ascii-byte) might be insecure and is not recommended.

Which I think translates to doing `bytes.decode(errors="replace")`. On this commit though, the code was not using replace:
https://github.com/aio-libs/yarl/blob/0e9b288e10e0b41d01fddb3c5cc2c9d8a2d14678/yarl/quoting.py#L132

So, the code must have produced an exception there, when the expected behaviour is that it would produce an output of `�`. I've not checked the code in the latest commit.

Dreamsorcerer: Although, that exception was actually suppressed and doesn't explain why it was changed to expect a ValueError...

Maybe I'm still misreading the spec, as it refers specifically to the input. So, I think that maybe we should be doing `val.encode(errors="replace")` at the very start of the function, and processing the bytes, rather than the str.

Maybe the comment was related to a similar misreading of the spec..

PR #217, opened 2018-07-13
URL::build(): Accept host-only URLs

behnam:
Accept host-only URLs, which are buildable using the parsing method as
`//host-name'.

Also, add check for setting port without a host, along-side test case
for it.

### DIFF
--- a/tests/test_url_build.py
+++ b/tests/test_url_build.py
@@ -16,15 +16,27 @@ def test_build_simple():
     assert str(u) == 'http://127.0.0.1'
 
 
-def test_build_scheme_and_host():
-    with pytest.raises(ValueError):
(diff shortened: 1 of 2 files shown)


behnam:
This could be considered an API-breaking change, since a call previously raising error now doesn't do so, and some application logic may depend on it. What do you think?

Btw, what's the best way to track API-breaking changes here?

behnam:
#ping Would you like me to make any changes here?

webknjaz:
LGTM, but I'll leave it up to @asvetlov to consider consequences.

Regarding tracking changes it's probably just that the project follows semver. I'd also add entry to `CHANGES.rst`.

asvetlov:
Thanks!

PR #218, opened 2018-07-13
URL: Add explicit_port instance property

behnam:
Having access to `explicit_port` makes it easier for working with the
URL instance as a data model, like when building mutable class on top of
`yarl.URL`.

On the other hand, the existing behavior of `port` property is better
for the applications. Therefore, it's better that we keep it the way it
is.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -211,7 +211,7 @@ There are two kinds of properties: *decoded* and *encoded* (with
       True
 
 
-.. attribute:: URL.port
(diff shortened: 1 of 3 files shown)


behnam:
Thanks, @webknjaz and @asvetlov. I didn't realize the docs are written manually. Updated the API doc for this.

behnam:
Thanks @asvetlov with fixes on the doc! :)

PR #219, opened 2018-07-13
travis.yml: Update Python build versions

behnam:
Set 3.7 to the stable build.

Add 3.8-dev as the next dev target.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,9 @@ language: python
 python:
   - 3.5
   - 3.6
-  - 3.7-dev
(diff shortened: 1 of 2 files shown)


behnam:
This one has to wait for now...

> Recent Python branches require OpenSSL 1.0.2+. As this library is not available for Trusty, 3.7, 3.7-dev, 3.8-dev, and nightly do not work (or use outdated archive).

https://docs.travis-ci.com/user/languages/python/#Development-releases-support

PR #220, opened 2018-07-14
test_url_query: Expand tests

behnam:
Expanding URL-query tests to cover more behaviors around special chars
when parsing, setting, re-setting and getting values.

Bug revealed are marked with `FIXME` to address separately.


behnam:
Looks like we have a conflict here: `mypy` is enforced in CI and I need to use py36-only syntax for type hinting to make it happy, but then it fails on py35.

I think we should limit mypy to py>=36. Let me send a separate PR for that.

tests/test_url_query.py
```diff
@@ -1,37 +1,82 @@
...
-
-
-def test_query_spaces():
-    url = URL('http://example.com?a+b=c+d')
-    assert url.query == MultiDict({'a b': 'c d'})
-
-
-def test_query_empty():
-    url = URL('http://example.com')
-    assert isinstance(url.query, MultiDictProxy)
```
asvetlov: Please keep a check for type returned by `url.query`

asvetlov (dismissed):

tests/test_url_query.py
```diff
@@ -1,37 +1,83 @@
...
+    ),
+]  # type: List[Tuple[URL, MultiDict]]
+
+
+@pytest.mark.parametrize(
+    'original_url, expected_query',
+    URLS_WITH_BASIC_QUERY_VALUES,
+)
+def test_query_basic_parsing(original_url, expected_query):
+    assert isinstance(original_url.query, MultiDictProxy)
```
behnam: Added here.

behnam:
@asvetlov, I believe this is ready for review.

mjpieters (approved):
LGTM, and the original change request has been addressed.

mjpieters:
As I can't push a minor fix to this PR, I'll re-create it and merge that instead.

PR #226, opened 2018-07-25
more informative error message

### LINKED ISSUE #224: Add type of object to error message
Spent some time with such error:
```
TypeError: Invalid variable type: mapping value should be str or int, got 20
```
https://github.com/aio-libs/yarl/blob/master/yarl/__init__.py#L807
Suggest to add type of object to error string, since in my case it was `<class 'numpy.int64'>`

pohmelie:
Fixes #224


webknjaz:
Thank you @pohmelie!

PR #228, opened 2018-07-31
Add extra chars to avoid unquoting

### LINKED ISSUE #223: _PATH_QUOTER unquote %3B to ; in url path. Is it correct behavior?
I have this URL: 
https://www.madewell.com/9%22-high-rise-skinny-jeans-in-isko-stay-blacktrade%3B-G1202.html 
which contains `%3B`(;). 

YARL unquote `%3B` to `;`
https://www.madewell.com/9%22-high-rise-skinny-jeans-in-isko-stay-blacktrade%3B-G1202.html
https://www.madewell.com/9%22-high-rise-skinny-jeans-in-isko-stay-blacktrade;-G1202.html

When I open URL from YARL it redirects me to the homepage. But If I make little changes:
`_PATH_QUOTER = _Quoter(safe='@:', protected='/+;') `- add `;` to protected parameter - it works correct

**My question:** Is it issue with YARL or with a site?
Thanks

amarynets:
In some cases when we have already quoted chars: ``&(%26),(%2C);(%3B)`` into path, yarl unquote it to standard representation: ``&,;` this behavior is wrong because server send redirection message with quoted url, but we again unquote it

Fixes #223

### DIFF
--- a/yarl/__init__.py
+++ b/yarl/__init__.py
@@ -132,7 +132,7 @@ class URL:
     __slots__ = ('_cache', '_val')
 
     _QUOTER = _Quoter()
-    _PATH_QUOTER = _Quoter(safe='@:', protected='/+')
+    _PATH_QUOTER = _Quoter(safe='@:', protected='/+&,;')
     _QUERY_QUOTER = _Quoter(safe='?/:@', protected='=+&;', qs=True)
     _QUERY_PART_QUOTER = _Quoter(safe='?/:@', qs=True)
     _FRAGMENT_QUOTER = _Quoter(safe='?/:@')


amarynets:
It's for fixing #223

webknjaz:
@amarynets any RFC reference confirming what the correct behavior is?

amarynets:
@webknjaz I'm not sure but those 2 RFC is related to this:
https://tools.ietf.org/html/rfc3986#section-3.3 - the last paragraph says that we can use reserved characters in url-path
https://tools.ietf.org/html/rfc3986#section-2.2 - the last paragraph, `applications should percent-encode data octets that correspond to characters in the reserved set`

>

asvetlov:
The change may affect existing code, please let me carefully reread specs to make decision

amarynets:
Sure. I have solved this in another way: `URL(url, encoded=True)`
In aiohttp you can pass URL object or string. The string was already encoded
But the question is still open: if we pass a string URL where reserved characters are already encoded we will get this issue again. Also, we can't use `encoded=True` in the case where string URL isn't encoded

amarynets:
@webknjaz I think that gives this part of encoding to a user has one disadvantage: the user does not want to care about encoding(if it does not use this lib directly). It wants just pass an URL in any format and have a good result(as a `requests` or `urllib` does).
Maybe it should be solved in `aiohttp` lib

webknjaz:
The thing is that yarl holds canonical form of URL internally, so it needs to encode/decode string to get that representation. It cannot magically figure out whether user *wants* to have special chars encoded or not.

amarynets:
Hey, @webknjaz I'm going to close this PR

webknjaz:
I'd wait for @asvetlov's resolution here as he knows yarl's architecture and decisions better

PR #231, opened 2018-08-13
Makefile: Set all to all steps running on CI

behnam:
To be able to cover all the test running on CI, including unit tests,
type hint checks, and doctest, setting the `make all` behavior to
include everything, to be able to catch problems locally as much as
possible.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ install:
 script:
   - make cov
   - make doc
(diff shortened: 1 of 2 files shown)


behnam:
@asvetlov, I believe this is ready for review.

asvetlov:
We don't use travis anymore

PR #243, opened 2018-09-27
Raise ValueError for URL with authority with path without leading slash

rominf:


yarl/__init__.py
```diff
@@ -612,6 +615,12 @@ def name(self):
         """The last part of parts."""
         return self._UNQUOTER(self.raw_name)
 
+    # noinspection PyPep8Naming
```
webknjaz: Don't do this.

yarl/__init__.py
```diff
@@ -612,6 +615,12 @@ def name(self):
...
+    @staticmethod
+    def raise_ValueError_for_URL_with_authority_with_path_without_leading_slash(host, path):
+        if len(host) > 0 and len(path) > 0 and path[0] != '/':
```
webknjaz: `if len(host) > 0 and len(path) > 0 and not path.startswith('/'):`

yarl/__init__.py
```diff
@@ -612,6 +615,12 @@ def name(self):
         """The last part of parts."""
         return self._UNQUOTER(self.raw_name)
 
+    # noinspection PyPep8Naming
+    @staticmethod
+    def raise_ValueError_for_URL_with_authority_with_path_without_leading_slash(host, path):
```
webknjaz: Use `validate_authority_uri_abs_path` and write a sane docstring instead.

rominf:
Is this better?

webknjaz:
Thanks!

PR #250, opened 2018-10-12
Minor fixes in test_quoting.py

gyermolenko:
**What has been done:**
1. Add raw string identifier - to fix "DeprecationWarning: invalid escape sequence \^"
(SHOULD_QUOTE remains the same)
2. `~` was missing from unquoted chars
3. I think `quoter(qs=True)(do_not_quote) == do_not_quote` without assert was unintended.

It could be the wrong place to ask, but... 
**Some questions**
4. `SHOULD_QUOTE` and `do_not_quote` - why not import from `quoting.py`, it has vars with the same meaning.
5. `test_quoting.py` has one todo item: `# TODO: should it encode percent?`
According to `rfc3986` it has to be encoded. 
Maybe the main concern is to do it's encoding/decoding only once? In that case it deserves its own issue.
6. does `assert s1 is s2` from `def test_quote_fastpath_safe():` has to be `is` instead of `==`?
I read this note about [cython limitations ](http://docs.cython.org/en/latest/src/userguide/limitations.html#identity-vs-equality-for-inferred-literals) , but question remains

### DIFF
--- a/tests/test_quoting.py
+++ b/tests/test_quoting.py
@@ -57,9 +57,9 @@ def test_never_quote(quoter):
     do_not_quote = '' .join(["ABCDEFGHIJKLMNOPQRSTUVWXYZ",
                              "abcdefghijklmnopqrstuvwxyz",
                              "0123456789",
-                             "_.-"])
+                             "_.-~"])
     assert quoter()(do_not_quote) == do_not_quote
-    quoter(qs=True)(do_not_quote) == do_not_quote
+    assert quoter(qs=True)(do_not_quote) == do_not_quote
 
 
 def test_safe(quoter):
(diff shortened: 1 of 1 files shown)


tests/test_quoting.py
```diff
@@ -57,9 +57,9 @@ def test_never_quote(quoter):
     do_not_quote = '' .join(["ABCDEFGHIJKLMNOPQRSTUVWXYZ",
                              "abcdefghijklmnopqrstuvwxyz",
                              "0123456789",
-                             "_.-"])
+                             "_.-~"])
```
webknjaz: Is that in RFC?

asvetlov: The change is correct.

unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"

asvetlov:
Thanks

tests/test_quoting.py
```diff
@@ -57,9 +57,9 @@ def test_never_quote(quoter):
     do_not_quote = '' .join(["ABCDEFGHIJKLMNOPQRSTUVWXYZ",
```
webknjaz: Nitpick: it's not necessary to do join, Python syntax automatically does that with a bunch of strings in brackets.
```python
do_not_quote = (
    "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
    "abcdefghijklmnopqrstuvwxyz"
    "0123456789"
    "_.-~"
)
```

asvetlov: Feel free to make this change in a separate commit

PR #262, opened 2018-12-07
Support empty password in URL

himikof:
Unlike `urllib.parse.urlsplit`, `yarl.URL` currently does not distinguish `http://user@localhost` and `http://user:@localhost`, returning `None` as `url.password` in both cases. I think that these cases are different and should be handled similar to `urllib.parse.urlsplit` function.

### DIFF
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -134,7 +134,14 @@ def test_password_non_ascii():
 
 def test_password_without_user():
     url = URL("http://:password@example.com")
+    assert url.user is None
     assert "password" == url.password
+
+
+def test_user_empty_password():
+    url = URL("http://user:@example.com")
+    assert "user" == url.user
+    assert "" == url.password
 
 
 def test_raw_host():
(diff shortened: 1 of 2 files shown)


asvetlov:
Do you have a real use case?

asvetlov:
I'm still curious about the use case but can accept the PR even without it :)

webknjaz:
@asvetlov the use case is pretty common: some services (even GitHub) use a token as a login in API URLs. Same goes for URL-formatted DSNs like ones SQLAlchemy accepts.

tests/test_url.py
```diff
@@ -134,9 +134,16 @@ def test_password_non_ascii():
 
 def test_password_without_user():
     url = URL("http://:password@example.com")
+    assert url.user is None
     assert "password" == url.password
 
 
+def test_user_empty_password():
+    url = URL("http://user:@example.com")
```
webknjaz: I'd also test `http://user@example.com` w/o a colon.

Issue #275, opened 2018-12-25
URL equality/passing None to URL.build


PromyLOPh:
The equality operator (==, \_\_eq\_\_) does not work properly when passing `None` to URL.build. Example:

```python
a, b = URL.build(scheme='http', host='0'), URL.build(scheme='http', host='0', fragment=None)
assert a == b # this is False
assert repr (a) == repr (b) # this is True, because "URL('http://0')" == "URL('http://0')"
```

Passing an empty string works as expected. This is very confusing, because repr produces exactly the same representation.

asvetlov:
That's interesting.
`URL.build` declares `fragment` as non-optional `str` parameter with default to empty string.
Why do you pass `None`? `None` for the argument is a grey zone: sometimes it works but there is no any commitment for this value type.

webknjaz:
I think `None` should correspond to missing fragment with no hash `#` symbol while empty string would be trailing hash in URL. `http://some.com/url` vs. `http://some.com/url#`

asvetlov:
`urllib.parse.urlparse` doesn't distinguish `http://some.com/url` and `http://some.com/url#`.
It returns an empty string fragment for both cases.
We can return to the question if will implement a custom HTTP parser but now there is nothing to do.

PromyLOPh:
I’m using hypothesis to generate test data. More specifically a dict which is then passed to yarl using URL.build(**dict). And I was simply assuming None would be mapped to the default value, which is apparently wrong (judging from https://yarl.readthedocs.io/en/latest/api.html#yarl.URL.fragment). This
is fine for me, however the behavior of == and repr() in this case are very confusing and thus None should either be rejected explicitly (TypeError) or silently converted to '' to prevent this situation.

asvetlov:
TypeError sounds good to me.
Would you prepare a pull request?

asvetlov:
Fixed by #276

PR #276, opened 2018-12-26
Disallow passing NoneType to URL.build

PromyLOPh:
Unsupported and results in strange behavior of equality operator. See
issue #275.

### DIFF
--- a/yarl/__init__.py
+++ b/yarl/__init__.py
@@ -204,6 +204,8 @@ class URL:
             raise ValueError('Can\'t build URL with "port" but without "host".')
         if query and query_string:
             raise ValueError('Only one of "query" or "query_string" should be passed')
+        if path is None or query_string is None or fragment is None:
+            raise TypeError('NoneType not accepted. Use "".')
 
         if not user and not password and not host and not port:
             netloc = ""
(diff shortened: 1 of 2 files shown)


yarl/__init__.py
```diff
@@ -204,6 +204,8 @@ def build(
             raise ValueError('Can\'t build URL with "port" but without "host".')
         if query and query_string:
             raise ValueError('Only one of "query" or "query_string" should be passed')
+        if path is None or query_string is None or fragment is None:
```
webknjaz: How about
```suggestion
        if None in (path, query_string, fragment):
```

asvetlov: I think better to have a separate check for every argument.
It can help with constructing more informative error text.
`'NoneType not accepted. Use "".'` looks too cryptic, a parameter name for the failed check can improve a user experience a lot.

yarl/__init__.py
```diff
@@ -204,6 +204,8 @@ def build(
...
+            raise TypeError('NoneType not accepted. Use "".')
```
webknjaz:
```suggestion
            raise TypeError('NoneType is illegal for "path", "query_string" and "fragment args, use string values instead.')
```

yarl/__init__.py
webknjaz: true

asvetlov:
Perfect! Thanks!

PR #286, opened 2019-02-06
Allow scheme replacement for relative urls

mikenemkov:


kxepal:
What's the case for that? Relative URL are relative because of that. With schema they becomes absolute and not always valid due to schema requirements. For instance, value in test `"http:///path/to"` is not valid HTTP URL, but we're make it feels like ok.

kxepal:
What's the case for that? Relative URL are relative because of that. With schema they becomes absolute and not always valid due to schema requirements. For instance, value in test `"http:///path/to"` is not valid HTTP URL, but we're make it feels like ok.

asvetlov:
Currently on my job we use yarl with non-standard schemas like `storage:`, `user:`, `job:` etc.
I've found that yarl restrictions are too tight somewhere (while I personally was the author of these limitations).
Sometimes we reconstruct `URL` from a string as a workaround.

asvetlov:
I'll merge it with changelog update.

kxepal:
> Currently on my job we use yarl with non-standard schemas like `storage:`, `user:`, `job:` etc.

This case is good, but will it produce valid results for them? Current test shows that it sets empty authority, but at the same time it's optional part of URL which shouldn't be present for `maitto`, `news`, `tel` and more else schemas.

PR #309, opened 2019-05-24
Allow user=None in URL.build and sync .pyi with source.

blazewicz:
## What do these changes do?

Fix type-check warnings when passing `Optional` arguments to URL.build.

Changes related only to static type checking.

## Checklist

  * The format is `<Name> <Surname>`.
  * Please keep alphabetical order, the file is sorted by names. 

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,7 @@ CHANGES
 ------------------
 
 * Distinguish an empty password in URL from a password not provided at all (#262)
+* Fixed annotations for optional parameters of ``URL.build`` (#309)
 
 1.3.0 (2018-12-11)
 ------------------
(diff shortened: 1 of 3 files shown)


asvetlov:
What is the difference between `user=None` and currently supported `user=""`.
Why empty string doesn't work for you?

blazewicz:
There is simply no need to use empty string as default value. Value of this parameter is checked and replaced in `URL._make_netloc` anyway. 

In my case I use `os.getenv()` to get paramters from environment and if variable is not set it returns None - it is very common to use `None` for unassigned values, for example `.get()` method of dict-like objects.

Use of `None` is generaly more lightweight and faster as it uses reference to built-in object instead of creating new object.

asvetlov:
You may be surprised but the empty string has the same caching strategy as `None`, so performance is basically the same.

Nonetheless, agree with the PR.

blazewicz:
> You may be surprised but the empty string has the same caching strategy as None, so performance is basically the same.

Probably true, I won't argue on that.

Should I squash commits?

asvetlov:
No, the project uses GitHub "Squash and merge" button.

asvetlov:
thanks!

PR #319, opened 2019-07-26
Cache for idna decode

sonic182:
Cache for idna decoding

### DIFF
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -2,6 +2,7 @@ import sys
 import pytest
 from urllib.parse import SplitResult
 
+import yarl
 from yarl import URL
 
 
@@ -1171,9 +1172,19 @@ def test_requoting():
     assert u.raw_query_string == "next=http://example.com/"
     assert str(u) == "http://127.0.0.1/?next=http://example.com/"
 
-def test_idna_cache():
+def test_idna_decode_cache():
     u = URL("http://127.0.0.1/")
     host = u.raw_host
-    assert u._get_host_decoded(host) == '127.0.0.1'
-    assert u._get_host_decoded(host) == '127.0.0.1'
-    assert u._get_host_decoded.cache_info().hits == 1
+    yarl._decode_host.cache_clear()
+    assert yarl._decode_host(host) == '127.0.0.1'
+    assert yarl._decode_host(host) == '127.0.0.1'
+    assert yarl._decode_host.cache_info().hits == 1
+
+
+def test_idna_encode_cache():
+    u = URL("http://127.0.0.1/")
+    host = u.raw_host
+    yarl._encode_host.cache_clear()
+    assert yarl._encode_host(host) == '127.0.0.1'
+    assert yarl._encode_host(host) == '127.0.0.1'
+    assert yarl._encode_host.cache_info().hits == 1
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -1170,3 +1170,10 @@ def test_requoting():
     u = URL("http://127.0.0.1/?next=http%3A//example.com/")
     assert u.raw_query_string == "next=http://example.com/"
     assert str(u) == "http://127.0.0.1/?next=http://example.com/"
+
+def test_idna_cache():
+    u = URL("http://127.0.0.1/")
+    host = u.raw_host
+    assert u._get_host_decoded(host) == '127.0.0.1'
+    assert u._get_host_decoded(host) == '127.0.0.1'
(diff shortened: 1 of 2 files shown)


yarl/__init__.py
```diff
@@ -446,7 +447,10 @@ def host(self):
             # fe80::2%Проверка
             # presence of '%' sign means only IPv6 address, so idna is useless.
             return raw
+        return self._get_host_decoded(raw)
 
+    @lru_cache()
```
asvetlov: The default size for `@lru_cache()` is 128: https://docs.python.org/3/library/functools.html#functools.lru_cache
`re` module has 512 default for caching.

I have no strong opinion for `yarl` cache size but would like to discuss it at least.

yarl/__init__.py
```diff
@@ -446,7 +447,10 @@ def host(self):
...
+    def _get_host_decoded(self, raw):
```
asvetlov: `lru_cache` uses `(self, raw)` as a cache key.
It means that the cache is unique for every URL instance because `self` is different.
Please make `_get_host_decoded` not a URL method but a first-class function in the global namespace to make caching really work.

asvetlov:
Thanks for your contribution.
Please address my notes.

yarl/__init__.py
sonic182: We can use 512 as `re` module. The perfect size is application dependant, we could try to make this configurable by doing an cache decorator and getting the cache size by a variable configuration.

What do you think?

socketpair:
I'm interested in benchmarks.

yarl/__init__.py
asvetlov: Let's not overcomplicate from the start: use 512 non-configurable constant.
We can add customization later on people's demand.

asvetlov:
@socketpair after looking on profiling charts (e.g. https://files.gitter.im/aio-libs/Lobby/Op5q/thumb/pythonupdated2.png and https://files.gitter.im/aio-libs/Lobby/KLeT/thumb/output.png) I'm pretty convinced that idna encoding provides significant performance degradation.

Benchmark numbers are always interesting though, sure :)

tests/test_url.py
```diff
@@ -1170,3 +1171,20 @@ def test_requoting():
     u = URL("http://127.0.0.1/?next=http%3A//example.com/")
     assert u.raw_query_string == "next=http://example.com/"
     assert str(u) == "http://127.0.0.1/?next=http://example.com/"
+
+def test_idna_decode_cache():
```
kxepal: How IDNA correlates with IP `127.0.0.1`?

sonic182: This test is only for the cache stuff, idna tests are up

kxepal: Yeap. I just stumbled upon test name. May be remove idna from it name to prevent confusion? Cache feature applies for all kind of hostname.

yarl/__init__.py
```diff
@@ -661,10 +677,7 @@ def _encode_host(cls, host):
             ip, sep, zone = host.partition("%")
             ip = ip_address(ip)
         except ValueError:
-            try:
-                host = idna.encode(host, uts46=True).decode("ascii")
-            except UnicodeError:
-                host = host.encode("idna").decode("ascii")
+            host = _encode_host(host)
```
mjpieters: Why not just cache the *whole* `URL._encode_host()` classmethod body here?

I understand that there's a nice symmetry with `_decode_host`, but it is not as if these two new functions are being used in multiple locations. The `URL.host` implementation is pretty thin beyond the `_decode_host()` call and is itself a cached property. In contrast, `URL._encode_host()` adds more string juggling that would benefit from being cached too. The downside would be that the cache would be used for pure ip addresses as well as hostnames.

tests/test_url.py
mjpieters: I'd rename this to `test_decode_host_cache`, and `test_encode_host_cache` below. Easier to correlate with the actual function names they cover.

sonic182: The main reason of caching is for usage of idna, @asvetlov what do you think about renaming this test something like `test_hostname_encoding_cache`?

sonic182: Forget my last comment, I haven't had read the @mjpieters comment, I'll rename like it

asvetlov:
Sorry, the slow-down reason was incorrectly detected.
https://github.com/aio-libs/yarl/pull/387 and https://github.com/aio-libs/aiohttp/pull/4402 fixes the problem without the need for caching.

PR #326, opened 2019-08-21
add 'from exc' to exception

samuelcolvin:
# What do these changes do?

Fix the traceback on this error.

I've checked and this seems to be the only place with this problem.

No

## Checklist

**I think this template is slightly wrong, there doesn't seem to be a `CHANGES/` directory.**

### DIFF
--- a/yarl/__init__.py
+++ b/yarl/__init__.py
@@ -157,12 +157,13 @@ class URL:
             else:
                 host = val.hostname
                 if host is None:
-                    raise ValueError("Invalid URL: host is required for abolute urls.")
+                    raise ValueError("Invalid URL: host is required for absolute urls")
 
                 try:
                     port = val.port
                 except ValueError as e:
-                    raise ValueError("Invalid URL: port can't be converted to integer") from e
+                    msg = "Invalid URL: port can't be converted to integer"
+                    raise ValueError(msg) from e
 
                 netloc = cls._make_netloc(
                     val.username, val.password, host, port, encode=True
--- a/yarl/__init__.py
+++ b/yarl/__init__.py
@@ -161,8 +161,8 @@ class URL:
 
                 try:
                     port = val.port
-                except ValueError:
-                    raise ValueError("Invalid URL: port can't be converted to integer")
+                except ValueError as e:
+                    raise ValueError("Invalid URL: port can't be converted to integer") from e
 
                 netloc = cls._make_netloc(
                     val.username, val.password, host, port, encode=True
--- a/yarl/__init__.py
+++ b/yarl/__init__.py
@@ -162,8 +162,9 @@ class URL:
                 try:
                     port = val.port
                 except ValueError as e:
-                    msg = "Invalid URL: port can't be converted to integer"
-                    raise ValueError(msg) from e
+                    raise ValueError(
+                        "Invalid URL: port can't be converted to integer"
+                    ) from e
 
                 netloc = cls._make_netloc(
                     val.username, val.password, host, port, encode=True


asvetlov:
thanks!

yarl/__init__.py
```diff
@@ -157,12 +157,13 @@ def __new__(cls, val="", *, encoded=False, strict=None):
...
                 if host is None:
-                    raise ValueError("Invalid URL: host is required for abolute urls.")
+                    raise ValueError("Invalid URL: host is required for absolute urls")
 
                 try:
                     port = val.port
-                except ValueError:
-                    raise ValueError("Invalid URL: port can't be converted to integer")
+                except ValueError as e:
+                    msg = "Invalid URL: port can't be converted to integer"
```
webknjaz: I don't think that an extra var is really needed here. You could just multiline the statement

yarl/__init__.py
```diff
@@ -157,12 +157,13 @@ def __new__(cls, val="", *, encoded=False, strict=None):
...
+                    raise ValueError(msg) from e
```
webknjaz:
```suggestion
                    raise ValueError(
                        "Invalid URL: port can't be converted to integer"
                    ) from e
```

webknjaz: On the second thought: doesn't Python 3 already chain exception causes already?

samuelcolvin: no

webknjaz: Why do you think so?

webknjaz: https://docs.python.org/3/library/exceptions.html#built-in-exceptions

> When raising (or re-raising) an exception in an except or finally clause `__context__` is automatically set to the last exception caught; if the new exception is not handled the traceback that is eventually displayed will include the originating exception(s) and the final exception.

samuelcolvin: I don't "think" anything, this a basic fundamental of python 3 programming.

```py
try:
    raise ValueError('foo')
except ValueError:
    raise TypeError('bar')

"""
  File "test.py", line 4, in <module>
    raise ValueError('foo')
ValueError: foo

During handling of the above exception, another exception occurred:

  File "test.py", line 6, in <module>
    raise TypeError('bar')
TypeError: bar
"""

try:
    raise ValueError('foo')
except ValueError as e:
    raise TypeError('bar') from e

"""
  File "test.py", line 22, in <module>
    raise ValueError('foo')
ValueError: foo

The above exception was the direct cause of the following exception:

  File "test.py", line 24, in <module>
    raise TypeError('bar') from e
TypeError: bar
"""
```

They look similar, but `During handling of the above exception, another exception occurred:` semantically means "we tried to handle the error, but while handling it another error happened", this is not what happened here.

webknjaz: It's just implicit vs. explicit cause difference IMO.

samuelcolvin: no. `... from e` is demonstrably correct, canonical python in this situation.

webknjaz: Both are correct: https://www.python.org/dev/peps/pep-3134

webknjaz: Explicit reason setting will just override the implicit one.

samuelcolvin: no, see my example above - the output is different.

Shall we have an argument about the spelling of "absolute" too? It would seem equally arcane and pointless to me.

I don't mind what you do here, but attempts to find problems and disagree for the sake of it will not encourage further contributions to this package from me or others.

webknjaz: I think it may be a personal preference but I don't really understand why explicit `from` is considered better. Other than that I don't really mind merging this.

samuelcolvin: Consider

```py
try:
    do_something()
except ValueError as e:
    record_error_in_do_something(e)  # <-- this raises an exception
```
Here, while trying to handle the error another unexpected error occurred, hence the traceback will say `During handling of the above exception, another exception occurred`.

Now consider,

```py
try:
    do_something()
except ValueError as e:
    raise ValueError(
        'an error occurred while doing something, try doing something else'
    ) from e
```

This is different, here we're just adding more colour to the original error. This is **not** an error that occurs while handling the exception, it's a new more informative exception. Hence the traceback says `The above exception was the direct cause of the following exception`. 

This is definitely what's happening in this case.

If you still need convincing please watch [this](https://youtu.be/jWKnA2ZdzHk) from 5:56:44.

asvetlov: Explicit `raise .. from ...` is better from explicit vs implicit perspective.
Another option is providing invalid port repr in exception text, it may help to pin the problem quickly.

I'm not very concerned about, the fix provided by @samuelcolvin is correct and good enough, it improves user experience.

asvetlov:
thanks!

PR #329, opened 2019-08-31
Adapt Cython building to pip==19

asvetlov:


.travis.yml
```diff
@@ -5,25 +5,28 @@ language: python
...
-  - python: 3.7
-    <<: *_xenial
-  - python: 3.8-dev
-    <<: *_xenial
   - stage: Deploy
     name: Publishing a source dist to PyPI
     python: 3.6
     if: tag IS present
-    script: skip
+    before_install: skip
```
webknjaz:
```suggestion
    before_install: []
```

.travis.yml
```diff
@@ -5,25 +5,28 @@ language: python
...
   allow_failures:
   - python: 3.8-dev
   include:
-  - python: 3.7
-    <<: *_xenial
-  - python: 3.8-dev
-    <<: *_xenial
   - stage: Deploy
     name: Publishing a source dist to PyPI
     python: 3.6
```
webknjaz: maybe update this too?
```suggestion
    python: 3.7
```

.travis.yml
```diff
@@ -38,7 +41,8 @@ env:
     - secure: "EvZbFFwD/XggLEy7H6yY/6OQrFbFbsRjRPrT1Hb8/TDGE8VlwC7us7xFSVIkLhj9K//VvEQucviP/bGg2EIziRng8oSz4x+X6CM0hra3PdmogZzMtBUFoY7olY5c/opo967WO8QZ+2yIArmvB1RbiQ6eh82l13qaI+R8wfIjxT+JpAjMHUGnRmqx1sUuM7ZQYQKlpt0Q9sFaPh8U2FEgspi9auvHp7WOCaJR8uGrHImfvf/c7lzxWjcImqaT357XJnxwPbV418J3p/cOpgHvjJD2jTA8yY0qo62rsemqYyhzCwaTU+Q0Sz5ykUW80a49tW3iGRVo1K0grzL9TcV3HAVNb419YBgKm2mko4hG/JpDyphydVRLvJLv+jv6kb8Iyx5N3h5m/n782EJIoitNxObnuqM9L+RqO5XcqCNBqfD38NSShvp6WKPUa8JVMU+d+yJ5rfFLcJxIWRxSGvoSi7uVuQ9ceOJRHhuV3O6p67KhDfd7/79xtxERxXU/9gwIliSURXU4r/uU7agr/jlKAP7P/6btQrfuxiEhFNr6Yc1YmxnF3ul/uzQyOZIa6nN1mzGrCtO5bLXMo9b66mPOaW4OXjIGELWD99Q1xIolj92w7cKRp5iKm5fnBuvbSnsQ1xq3GSNfQWp6FzxHhvxVL4C42tfth61kpHuxwbY7U9o="
```
webknjaz: @asvetlov do you remember what's in this global var? We use the deployment provider directly... Do we still need this?

.travis.yml
asvetlov: I'm curious why?
`skip` works too, isn't it?

.travis.yml
```diff
@@ -5,25 +5,28 @@ language: python
 python:
   - 3.5
   - 3.6
+  - 3.7
+  - 3.8-dev
+
+dist: xenial
+sudo: required
```
webknjaz: This is irrelevant. It's a default now since Travis CI doesn't support container-based runners anymore.
```suggestion
```

.travis.yml
asvetlov: I don't. Let's check deployment (and fix if necessary) in another PR.

.travis.yml
```diff
@@ -5,25 +5,28 @@ language: python
...
   - python: 3.8-dev
   include:
-  - python: 3.7
-    <<: *_xenial
-  - python: 3.8-dev
-    <<: *_xenial
   - stage: Deploy
     name: Publishing a source dist to PyPI
-    python: 3.6
+     python: 3.7
```
webknjaz:
```suggestion
    python: 3.7
```

.travis.yml
webknjaz: Yes, it's just cleaner this way: the empty sequence of commands `[]` means run nothing while `skip` is a hack for `install`/`script` steps which require a command there. It's an internal no-op command in Travis CI.

.travis.yml
webknjaz: Looks like it was an attempt to figure out adding creds: https://github.com/aio-libs/yarl/commit/f825a5d7068e52346615e87b9f5c1753b38d73d3

yarl/quoting.py
```diff
@@ -1,7 +1,16 @@
+import os
 import re
+import sys
 from string import ascii_letters, ascii_lowercase, digits
 from typing import Optional, TYPE_CHECKING, cast
 
+
+NO_EXTENSIONS = bool(os.environ.get("YARL_NO_EXTENSIONS"))  # type: bool
```
webknjaz: it's shorter:
```suggestion
NO_EXTENSIONS = bool(os.getenv("YARL_NO_EXTENSIONS"))  # type: bool
```

yarl/quoting.py
```diff
@@ -1,7 +1,16 @@
...
+
+if sys.implementation.name != "cpython":
```
webknjaz: I prefer a higher-level API for this:
```suggestion
if platform.python_implementation() != "CPython":
```

yarl/quoting.py
```diff
@@ -1,7 +1,16 @@
+import os
 import re
+import sys
```
webknjaz:
```suggestion
import platform
```

yarl/quoting.py
webknjaz: How about having `EXTENSIONS` instead of `NO_EXTENSIONS`? Double negation like `not NO_EXTENSIONS` looks confusing.

yarl/__init__.py
```diff
@@ -207,8 +207,10 @@ def build(
         if query and query_string:
             raise ValueError('Only one of "query" or "query_string" should be passed')
         if path is None or query_string is None or fragment is None:
-            raise TypeError('NoneType is illegal for "path", "query_string" and '
-                            '"fragment" args, use string values instead.')
+            raise TypeError(
+                'NoneType is illegal for "path", "query_string" and '
+                '"fragment" args, use string values instead.'
```
webknjaz: I usually prefer to have a trailing comma at the end of multiline args list. It makes the style more consistent should the need to add/remove other args arise.
```suggestion
                '"fragment" args, use string values instead.',
```

yarl/quoting.py
asvetlov: 1. The suggestion requires additional import
2. More importantly, I would say thay `sys.implementation` is more high-level API than low-level `platform` helper module.

webknjaz: Weird, I was reading somewhere that `platform` handles corner cases better...

yarl/quoting.py
asvetlov: Do you really care? I prefer dict-like `os.environ` over more narrow `os.getenv()` / `os.setenv()`.

webknjaz: I like it more but I don't have strong opinions regarding this. Feel free to ignore this.

asvetlov: `NO_EXTENSIONS` corresponds to env var, `EXTENSIONS` is for *not env var* which is an extra step in the logic understanding.

yarl/__init__.py
asvetlov: 1. This is not multi-args list but the single argument.
2. I don't want to discuss code styling anymore. The code is formatted by `black`, convince them.
3. Trailing comma is a syntax error in Python 3.5, BTW

asvetlov:
@webknjaz I very appreciate your contribution.
Please stop haggling, get the fact that not every code line is written in your preferred style.
If the code works good and it is a well readable -- that's enough usually.
Replacing a perfect line with awesome one doesn't add value if both lines are basically equal.

PR #353, opened 2019-10-01
Add linters on azure

asvetlov:


setup.cfg
```diff
@@ -1,11 +1,16 @@
...
 license_file = LICENSE
 
 
+[tool:pytest]
+addopts = --cov=yarl -v
+filterwarnings=error
+norecursedirs = dist docs build .tox .eggs venv virtualenv .git
+minversion = 3.8.2
+testpaths = tests/
+junit_suite_name = yarl_test_suite
```
webknjaz: @tonybaloney released https://pypi.org/project/pytest-nunit/ in recent months. AFAIU this format supports richer reporting. @asvetlov do you want to give it a try?

setup.cfg
```diff
@@ -1,11 +1,16 @@
...
-
 [metadata]
 license_file = LICENSE
 
 
+[tool:pytest]
+addopts = --cov=yarl -v
+filterwarnings=error
+norecursedirs = dist docs build .tox .eggs venv virtualenv .git
+minversion = 3.8.2
```
webknjaz: how about 4.0/5.0?

setup.cfg
asvetlov: Thanks for the link.
Created https://github.com/aio-libs/yarl/issues/355

Now I want to make everything working, left everything not required to further PRs

setup.cfg
asvetlov: This is a copy-paste from other project.
Please feel free to upgrade to newer versions.

PR #356, opened 2019-10-02
Publish dists on PyPI

asvetlov:


.azure-pipelines/deploy.yml
```diff
@@ -0,0 +1,52 @@
...
+        inputs:
+          versionSpec: '3.7'
+        displayName: "Use Python 3.7"
+
+      - script: |
+          pip install -U setuptools twine wheel
+        displayName: "Install tools"
+
+      - script: |
+          python setup.py sdist
```
webknjaz: Only source dists?

asvetlov: The PR is still in progress.
I want to have a job for publishing tarball and another one for manylinux.

PR #376, opened 2019-11-14
Include name in exception - starting from slash is forbidden

clickthisnick:
## What do these changes do?
Saw this in our error logs today:
```bash
raise ValueError("Appending path " "starting from slash is forbidden")
```

Was confusing because I thought the path was an empty string. I changed error message to include the name since its not exactly clear what caused it.

Our urls are like baseurl / a / b / c / d

The error message will be changed and include more relevant information.

## Checklist

- [X] I think the code is well written - Yes
- [X] Unit tests for the changes exist - There is already test
- [X] Documentation reflects the changes - I didn't see doc on this error specifically and not sure how changelog/release notes get formed.
- [x] Add a new news fragment into the `CHANGES` folder - will do


clickthisnick:
Will black format later tonight, on computer with just web browser now

yarl/__init__.py
```diff
@@ -292,7 +292,9 @@ def __gt__(self, other):
     def __truediv__(self, name):
         name = self._PATH_QUOTER(name)
         if name.startswith("/"):
-            raise ValueError("Appending path " "starting from slash is forbidden")
+            raise ValueError(
+                "Appending path \"{}\" starting from slash is forbidden".format(name)
```
asvetlov:
```suggestion
                "Appending path {!r} starting from slash is forbidden".format(name)
```

asvetlov:
Thanks!

PR #397, opened 2019-12-12
Add YARL_NO_EXTENSIONS environment variable in the docs

### LINKED ISSUE #394: Add YARL_NO_EXTENSIONS to README or somewhere in the documentation.
Some popular distros, like Alpine Linux, don't have gcc installed by default. I understand the performance reasoning behind having optimized version as default, but there are valid reasons for not having gcc installed, like image size, so mentioning this environment variable somewhere in documentation as an option is a good idea.

NazarTopolsky:
Fixes #394


README.rst
```diff
@@ -87,6 +87,19 @@ Installation
 
 The library is Python 3 only!
 
+PyPI contains binary wheels for Linux, Windows and MacOS.  If you want to install
+``yarl`` on another operation system (or *Alpine Linux* inside a Docker) the
+Tarball will be used to compile the library from sources.  It requires C compiler and
```
webknjaz: I don't think that tarball should be capitalized

webknjaz: Also: "a C compiler"

README.rst
```diff
@@ -87,6 +87,19 @@ Installation
...
+Python headers installed.
+
+To skip the compilation please use `YARL_NO_EXTENSIONS` environment variable,
```
webknjaz:
```suggestion
To skip the compilation, you must explicitly opt-in by setting `YARL_NO_EXTENSIONS` environment variable to a non-empty value,
```

README.rst
```diff
@@ -87,6 +87,19 @@ Installation
...
+e.g.:
+
+.. code-block:: bash
+
+   $ YARL_NO_EXTENSIONS=1 pip install yarl
+
+Please note, Pure Python (uncompiled) version is much slower.
```
webknjaz: Please also mention that it's always pure-python under PyPy interpreter series.

asvetlov: I'm not sure that PyPy should be explicitly noted.
The performance under PyPy is fine; PyPy executes C Extensions slower than native Python code.

webknjaz: Just to inform users that they don't have to set `YARL_NO_EXTENSIONS` as it won't change anything.

NazarTopolsky:
Something I haven't considered before is multidict dependency. On platforms without a compiler, installation will fail even with `YARL_NO_EXTENSIONS` because multidict also requires a compiler. I think it makes sense to mention `MULTIDICT_NO_EXTENSIONS` as well.

Maybe we can expand [dependencies](https://yarl.readthedocs.io/en/latest/#dependencies) section with something like this:

>Note that, like yarl, multidict also relies on C extensions for performance boost. If your platform prevents you from using a binary wheel and doesn't have a compiler, you will also need to set `MULTIDICT_NO_EXTENSIONS` environment variable as well. Consult multidict documentation for details.

Should I add this?

README.rst
NazarTopolsky: "Please note Pure Python (uncompiled) version is much slower. However, PyPy always uses pure python implementation, and, as such, is unaffected by this variable."

Would something like this work?

webknjaz: s/is/it is/

webknjaz:
s/and doesn't have a compiler/or doesn't have a compiler/

webknjaz:
But I think that multidict implementation details are of no concern for us in YARL docs. The relevant info should be mentioned in the multidict's docs.

NazarTopolsky:
I see the point. Yarl installation will blow up if there's no C compiler in the system even with YARL_NO_EXTENSIONS, but the solution can be found on Google.

README.rst
```diff
@@ -87,6 +87,20 @@ Installation
 
 The library is Python 3 only!
 
+PyPI contains binary wheels for Linux, Windows and MacOS.  If you want to install
+``yarl`` on another operation system (or *Alpine Linux* inside a Docker) the
```
webknjaz: The correct word is "operating". Also, docker is irrelevant here: the containerization technique doesn't govern the presence of a compiler.
```suggestion
``yarl`` on another operating system (like *Alpine Linux*, which is not
manylinux-compliant because of the missing glibc and therefore, cannot be
used with our wheels) the
```

README.rst
```diff
@@ -87,6 +87,20 @@ Installation
...
+tarball will be used to compile the library from sources.  It requires a C compiler and
```
webknjaz:
```suggestion
tarball will be used to compile the library from the source code. It requires a C compiler and
```

README.rst
```diff
@@ -87,6 +87,20 @@ Installation
...
+Python headers installed.
+
+To skip the compilation you must explicitly opt-in by setting use `YARL_NO_EXTENSIONS`
```
webknjaz:
```suggestion
To skip the compilation, you must explicitly opt-in by setting the `YARL_NO_EXTENSIONS`
```

README.rst
```diff
@@ -87,6 +87,20 @@ Installation
...
+environment variable to a non-empty value, e.g.:
+
+.. code-block:: bash
+
+   $ YARL_NO_EXTENSIONS=1 pip install yarl
+
+Please note that Pure Python (uncompiled) version is much slower. However, PyPy always
```
webknjaz: "pure-Python" is the proper spelling:
```suggestion
Please note that the pure-Python (uncompiled) version is much slower. However, PyPy always
```

README.rst
```diff
@@ -87,6 +87,20 @@ Installation
...
+uses pure python implementation, and, as such, it is unaffected by this variable.
```
webknjaz:
```suggestion
uses a pure-Python implementation, and, as such, it is unaffected by this variable.
```

asvetlov:
@webknjaz would you assign the PR to yourself and drive to down to finish?
I'm pretty sure that you are the best person for polishing the wording.

CHANGES.rst
```diff
@@ -14,6 +14,24 @@ Changelog
 
 .. towncrier release notes start
 
+1.5.0a0 (2020-01-07)
```
webknjaz: You shouldn't change this file.

webknjaz:
@NazarTopolsky your commits have the email that is not bound to your GitHub user. I think, your computer is misconfigured. You can fix this (for future commits) using `git config --global user.email 'your@email.here'` or by confirming the address you currently use in your profile.

CHANGES.rst
NazarTopolsky: Sorry, missed this when committing. I should simply revert this change, right?

webknjaz: Yes, just make it go away :)

webknjaz:
Thanks!

PR #401, opened 2019-12-20
Support for boolean parameter parsing

tigerism:
## What do these changes do?

This pull request allows for parameter parsing of booleans in URL.with_query.

TypeError will no longer be thrown for booleans in URL.with_query.

## Related issue number

None.

## Checklist

### DIFF
--- a/yarl/__init__.py
+++ b/yarl/__init__.py
@@ -909,8 +909,8 @@ class URL:
     def with_query(self, *args, **kwargs):
         """Return a new URL with query part replaced.
 
-        Accepts any Mapping (e.g. dict, multidict.MultiDict instances)
-        or str, autoencode the argument if needed.
+        Accepts any Mapping (e.g. dict, multidict.MultiDict instances),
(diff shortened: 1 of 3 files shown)


yarl/__init__.py
```diff
@@ -858,6 +858,8 @@ def _query_var(v):
             return v
         if type(v) is int:  # no subclasses like bool
             return str(v)
+        if isinstance(v, bool):
+            return "true" if v else "false"
```
asvetlov: Thank you for the PR.

There is no the single representation for `bool` in the query string.
Some systems prefer `true`/`false`, others like `yes`/`no`, `on`/`off`, `Y`/`N`, `1`/`0`, etc.
That's why `yarl` doesn't make this decision and don't support `bool` explicitly. Please convert booleans to your preferred form explicitly before calling `yarl` API.

asvetlov:
#405 add a note to docs; the note describes why bools are not supported

PR #405, opened 2019-12-20
Clarify bools support

asvetlov:
Describe why booleans are not supported

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -107,7 +107,7 @@ There is no standard for boolean representation of boolean values.
 Some systems prefer ``true``/``false``, others like ``yes``/``no``, ``on``/``off``,
 ``Y``/``N``, ``1``/``0``, etc.
 
-``yarl`` cannot make an unambiguous decision how to serialize a :class:`bool` values.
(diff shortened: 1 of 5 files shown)


README.rst
```diff
@@ -99,6 +99,19 @@ API documentation
 
 The documentation is located at https://yarl.readthedocs.org
 
+Why boolean is not supported by URL query API?
```
webknjaz:
```suggestion
Why isn't boolean supported by the URL query API?
```

README.rst
```diff
@@ -99,6 +99,19 @@ API documentation
...
+----------------------------------------------
+
+There is no standard for boolean representation of boolean values.
+
+Some systems prefer ``true``/``false``, others like ``yes``/``no``, ``on``/``off``,
+``Y``/``N``, ``1``/``0``, etc.
+
+``yarl`` cannot make an unambiguous decision how to serialize a ``bool`` values.
```
webknjaz:
```suggestion
``yarl`` cannot make an unambiguous decision on how to serialize ``bool`` values because it is specific to how the end-user's application is built and would be different for different apps.
```

asvetlov:
Updated

PR #409, opened 2019-12-28
Fix URL parsing with Python 3.8.1 and 2.7.15

onovy:
## What do these changes do?

URL parsing has been changed in newer Python releases, see: https://bugs.python.org/issue27657

No, just test fix.

## Related issue number

None.

## Checklist

### DIFF
--- a/CHANGES/409.bugfix
+++ b/CHANGES/409.bugfix
@@ -1,0 +1,1 @@
+Fix tests with newer Python (3.7.6, 3.8.1 and 3.9.0+).
--- a/tests/test_url_parsing.py
+++ b/tests/test_url_parsing.py
@@ -40,9 +40,15 @@ class TestScheme:
 
     def test_no_scheme1(self):
         u = URL("google.com:80")
-        assert u.scheme == ""
-        assert u.host is None
-        assert u.path == "google.com:80"
+        # See: https://bugs.python.org/issue27657
+        if sys.version_info[:3] == (3, 7, 6) or sys.version_info[:3] == (3, 8, 1) or sys.version_info >= (3, 9, 0):
+            assert u.scheme == "google.com"
+            assert u.host is None
+            assert u.path == "80"
+        else:
+            assert u.scheme == ""
+            assert u.host is None
+            assert u.path == "google.com:80"
         assert u.query_string == ""
         assert u.fragment == ""
 


bigon:
Are you sure this is OK?

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=952211

bigon:
This makes more sense (if we don't care about python2 compatibility):

``` diff
--- a/tests/test_url_parsing.py
+++ b/tests/test_url_parsing.py
@@ -40,9 +40,15 @@ class TestScheme:

     def test_no_scheme1(self):
         u = URL("google.com:80")
-        assert u.scheme == ""
-        assert u.host is None
-        assert u.path == "google.com:80"
+        # See: https://bugs.python.org/issue27657
+        if sys.version_info <= (3, 8, 2):
+            assert u.scheme == "google.com"
+            assert u.host is None
+            assert u.path == "80"
+        else:
+            assert u.scheme == ""
+            assert u.host is None
+            assert u.path == "google.com:80"
         assert u.query_string == ""
         assert u.fragment == ""
```

bigon:
**I deleted my last comment as it was incorrect**

According to https://bugs.python.org/issue27657, they have reverted the fix in 3.8.2 and 3.7.7 as it was causing unexpected behavior changes in a minor release.

So the patch should look more like the following if I understand well:

```diff
--- a/tests/test_url_parsing.py
+++ b/tests/test_url_parsing.py
@@ -40,9 +40,15 @@ class TestScheme:

     def test_no_scheme1(self):
         u = URL("google.com:80")
-        assert u.scheme == ""
-        assert u.host is None
-        assert u.path == "google.com:80"
+        # See: https://bugs.python.org/issue27657
+        if sys.version_info[:3] == (3, 7, 6) or sys.version_info[:3] == (3, 8, 1) or sys.version_info >= (3, 9, 0):
+            assert u.scheme == "google.com"
+            assert u.host is None
+            assert u.path == "80"
+        else:
+            assert u.scheme == ""
+            assert u.host is None
+            assert u.path == "google.com:80"
         assert u.query_string == ""
         assert u.fragment == ""
```

onovy:
> Are you sure this is OK?

yes, I was sure it was right. Change in Python was reverted after this PR.

I checked your patch and looks correct. Updated this PR to reflect that.

Thanks.

bigon:
Thanks

asvetlov:
Merged manually by 3d5bf8f
Thanks for the PR!

PR #443, opened 2020-04-27
Support quoting list values in mappings

Lonami:
## What do these changes do?

These changes resolve aio-libs/aiohttp#4714.

Yes, users can now use lists or tuples when quoting mappings.

## Checklist

### DIFF
--- a/CHANGES/443.feature
+++ b/CHANGES/443.feature
@@ -1,6 +1,5 @@
-These changes fix https://github.com/aio-libs/aiohttp/issues/4714, that is,
-we now support the use of lists and tuples when quoting mappings such as
-Python's builtin dict. As an example from the tests:
+Allow use of sequences such as :class:`list` and :class:`tuple` in the values
+of a mapping such as :class:`dict` to represent that a key has many values:
 
     url = URL("http://example.com")
     assert url.with_query({"a": [1, 2]}) == URL("http://example.com/?a=1&a=2")
--- a/CHANGES/443.feature
+++ b/CHANGES/443.feature
@@ -1,0 +1,6 @@
+These changes fix https://github.com/aio-libs/aiohttp/issues/4714, that is,
+we now support the use of lists and tuples when quoting mappings such as
+Python's builtin dict. As an example from the tests:
+
+    url = URL("http://example.com")
+    assert url.with_query({"a": [1, 2]}) == URL("http://example.com/?a=1&a=2")
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -547,6 +547,9 @@ section generates a new *URL* instance.
 
       The library accepts :class:`str` and :class:`int` as query argument values.
 
+      If a mapping such as :class:`dict` is used, the values may also be
+      :class:`list` or :class:`tuple` to represent a key has many values.
+
       Please see :ref:`yarl-bools-support` for the reason why :class:`bool` is not
       supported out-of-the-box.
 
@@ -556,6 +559,8 @@ section generates a new *URL* instance.
       URL('http://example.com/path?c=d')
       >>> URL('http://example.com/path?a=b').with_query({'c': 'd'})
       URL('http://example.com/path?c=d')
+      >>> URL('http://example.com/path?a=b').with_query({'c': [1, 2]})
+      URL('http://example.com/path?c=1&c=2')
       >>> URL('http://example.com/path?a=b').with_query({'кл': 'зн'})
       URL('http://example.com/path?%D0%BA%D0%BB=%D0%B7%D0%BD')
       >>> URL('http://example.com/path?a=b').with_query(None)
@@ -591,6 +596,9 @@ section generates a new *URL* instance.
 
       The library accepts :class:`str` and :class:`int` as query argument values.
 
+      If a mapping such as :class:`dict` is used, the values may also be
+      :class:`list` or :class:`tuple` to represent a key has many values.
+
       Please see :ref:`yarl-bools-support` for the reason why :class:`bool` is not
       supported out-of-the-box.
 
@@ -600,6 +608,8 @@ section generates a new *URL* instance.
       URL('http://example.com/path?a=b&c=d')
       >>> URL('http://example.com/path?a=b').update_query({'c': 'd'})
       URL('http://example.com/path?a=b&c=d')
+      >>> URL('http://example.com/path?a=b').update_query({'c': [1, 2]})
+      URL('http://example.com/path?a=b&c=1&c=2')
       >>> URL('http://example.com/path?a=b').update_query({'кл': 'зн'})
       URL('http://example.com/path?a=b&%D0%BA%D0%BB=%D0%B7%D0%BD')
       >>> URL('http://example.com/path?a=b&b=1').update_query(b='2')
--- a/tests/test_update_query.py
+++ b/tests/test_update_query.py
@@ -122,6 +122,7 @@ def test_with_query_list_int():
     [
         pytest.param({"a": [1, 2]}, "?a=1&a=2", id="list"),
         pytest.param({"a": (1, 2)}, "?a=1&a=2", id="tuple"),
+        pytest.param({"a[]": [1, 2]}, "?a[]=1&a[]=2", id="key with braces"),
         pytest.param({"a": ["1", 2]}, "?a=1&a=2", id="mixed types"),
         pytest.param({"a": 1, "b": [2, 3]}, "?a=1&b=2&b=3", id="single then list"),
         pytest.param({"a": [1, 2], "b": 3}, "?a=1&a=2&b=3", id="list then single"),
--- a/tests/test_update_query.py
+++ b/tests/test_update_query.py
@@ -134,12 +134,14 @@ def test_with_query_sequence_mixed_types():
 
 def test_with_query_sequence_single_then_list():
     url = URL("http://example.com")
-    assert url.with_query({"a": 1, "b": [2, 3]}) == URL("http://example.com/?a=1&b=2&b=3")
+    expected = URL("http://example.com/?a=1&b=2&b=3")
+    assert url.with_query({"a": 1, "b": [2, 3]}) == expected
 
 
 def test_with_query_sequence_list_then_single():
     url = URL("http://example.com")
-    assert url.with_query({"a": [1, 2], "b": 3}) == URL("http://example.com/?a=1&a=2&b=3")
+    expected = URL("http://example.com/?a=1&a=2&b=3")
+    assert url.with_query({"a": [1, 2], "b": 3}) == expected
 
 
 def test_with_query_sequence_nested_sequence():
--- a/tests/test_update_query.py
+++ b/tests/test_update_query.py
@@ -120,25 +120,26 @@ def test_with_query_list_int():
 @pytest.mark.parametrize(
     ("query", "expected"),
     [
-        pytest.param({"a": []}, "?", id="empty list"),
-        pytest.param({"a": ()}, "?", id="empty tuple"),
-        pytest.param({"a": [1]}, "?a=1", id="single list"),
-        pytest.param({"a": (1,)}, "?a=1", id="single tuple"),
-        pytest.param({"a": [1, 2]}, "?a=1&a=2", id="list"),
-        pytest.param({"a": (1, 2)}, "?a=1&a=2", id="tuple"),
-        pytest.param({"a[]": [1, 2]}, "?a[]=1&a[]=2", id="key with braces"),
-        pytest.param({"&": [1, 2]}, "?%26=1&%26=2", id="quote key"),
-        pytest.param({"a": ["1", 2]}, "?a=1&a=2", id="mixed types"),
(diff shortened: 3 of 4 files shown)


Lonami:
I haven't checked performance but if it's an issue it can be improved by re-implementing `_query_var` inside the new `_query_seq_pairs`. And if performance is an absolute concern, we may use the trick where we avoid global look-ups to things like `isinstance` or the `tuple/list` types.

Lonami:
With regards to:
> Make sure to use full sentences with correct case and punctuation, for example: `Fix issue with non-ascii contents in doctest text files.`

Is that necessary? Taking a quick look at the commit log, it seems to be used inconsistently anyway (and many commits don't follow that convention).

Lonami:
Oh, I hadn't seen the `CHANGES/.TEMPLATE.rst` file because it starts with a dot so it was hidden. Was I supposed to use that in the changes? If yes, it could be a good idea to rename it to something else and add comments indicating "WRITE HERE" or something.

webknjaz:
> Oh, I hadn't seen the `CHANGES/.TEMPLATE.rst` file because it starts with a dot so it was hidden. Was I supposed to use that in the changes?

No, that template is used by towncrier to generate `CHANGES.rst` (which we run before making releases). You should usually add a short change note file and that's it. Here's an example https://github.com/aio-libs/yarl/blob/01e529690fbb6d122e4a4a6575e1aa5edba4ce47/CHANGES/389.feature, check how it looks in https://github.com/aio-libs/yarl/blob/master/CHANGES.rst#features.

webknjaz:
> Is that necessary? Taking a quick look at the commit log, it seems to be used inconsistently anyway (and many commits don't follow that convention).

It's not about the commit messages but about the changes to docs.

tests/test_update_query.py
```diff
@@ -117,6 +117,45 @@ def test_with_query_list_int():
...
+def test_with_query_sequence_single_then_list():
+    url = URL("http://example.com")
+    expected = URL("http://example.com/?a=1&b=2&b=3")
+    assert url.with_query({"a": 1, "b": [2, 3]}) == expected
+
+
+def test_with_query_sequence_list_then_single():
+    url = URL("http://example.com")
+    expected = URL("http://example.com/?a=1&a=2&b=3")
+    assert url.with_query({"a": [1, 2], "b": 3}) == expected
```
webknjaz, line 144: these tests are essentially one sigle test and so it should use `@pytest.mark.parametrize`. You can use `pytest.param(id=)` if you want to improve the test report.

tests/test_update_query.py
```diff
@@ -117,6 +117,45 @@ def test_with_query_list_int():
...
+def test_with_query_sequence_nested_sequence():
+    url = URL("http://example.com")
+    with pytest.raises(TypeError):
+        url.with_query({"a": [[1]]})
+
+
+def test_with_query_sequence_using_pairs():
+    url = URL("http://example.com")
+    with pytest.raises(TypeError):
+        url.with_query([("a", [1, 2])])
```
webknjaz, line 156: parametrize plz

docs/api.rst
```diff
@@ -600,6 +608,8 @@ section generates a new *URL* instance.
       URL('http://example.com/path?a=b&c=d')
       >>> URL('http://example.com/path?a=b').update_query({'c': 'd'})
       URL('http://example.com/path?a=b&c=d')
+      >>> URL('http://example.com/path?a=b').update_query({'c': [1, 2]})
+      URL('http://example.com/path?a=b&c=1&c=2')
```
webknjaz, line 612: This may be seen as controversial because, for example, web-browsers use `varname[]` format for arrays:
```diff
      URL('http://example.com/path?a=b&c=1&c=2')
      URL('http://example.com/path?a=b&c[]=1&c[]=2')
```

Lonami: Does that not depend on the REST API? Users can use `'c[]'` if the API expects arrays to use `'c[]'` as the key, although I haven't seen that.

Once again, `'c'` and not `'c[]'` is consistent with both `urllib` and `requests`, and sites like Google translate don't use `dt[]` but `dt`.

tests/test_update_query.py
Lonami: I'm not that familiar with `pytest`, and none of the previous test used that (although it seems they could), so that's why I didn't. But I will make the changes.

tests/test_update_query.py
Lonami: I'm not sure on this one, in my opinion they're quite different in nature so they could deserve different functions, but I will also parametrize it if you want.

docs/api.rst
webknjaz: I think this depends on how the backend software handles this. PHP used to handle it gracefully (the form sends `c[]` but then in HTTP handler it's already `c` with values).

You should at least add a good tests coverage including such corner cases.

CHANGES/443.feature
```diff
@@ -0,0 +1,6 @@
+These changes fix https://github.com/aio-libs/aiohttp/issues/4714, that is,
```
webknjaz, line 1: This is going to be a part of the changelog under the version.
It'd be weird to see something like:

v1.0.0
======
features
--------
* This changes fix xxx

webknjaz: Because it's not a bug, there shouldn't be a mention of fixing things.
It may say something like "add support for XXX. resolves issue YYY"

docs/api.rst
Lonami: Could you suggest a way to test this? We already have tests that cover the fact it's `c` and not `c[]`. I will try to write some test page and server to see what browsers send.

tests/test_update_query.py
```diff
@@ -117,6 +117,34 @@ def test_with_query_list_int():
...
 
 
+@pytest.mark.parametrize(
+    "query,expected",
+    [
+        pytest.param({"a": [1, 2]}, "?a=1&a=2", id="list"),
+        pytest.param({"a": (1, 2)}, "?a=1&a=2", id="tuple"),
+        pytest.param({"a": ["1", 2]}, "?a=1&a=2", id="mixed types"),
+        pytest.param({"a": 1, "b": [2, 3]}, "?a=1&b=2&b=3", id="single then list",),
+        pytest.param({"a": [1, 2], "b": 3}, "?a=1&a=2&b=3", id="list then single",),
```
webknjaz, line 127: could also include potentionally buggy cases
```suggestion
        pytest.param({"a": [1, 2], "b": 3}, "?a=1&a=2&b=3", id="list then single"),
        pytest.param({"a": "1&a=2", "b": 3}, "?a=1%26a=2&b=3", id="ampersand value"),
        pytest.param({"a": ["1&a=2", 3], "b": 3}, "?a=1%26a=2&a=3&b=3", id="ampersand value mixed with a list"),
```

tests/test_update_query.py
```diff
@@ -117,6 +117,34 @@ def test_with_query_list_int():
...
+@pytest.mark.parametrize(
+    "query",
+    [
+        pytest.param({"a": [[1]]}, id="nested"),
+        pytest.param([("a", [1, 2])], id="tuple list"),
+    ],
+)
+def test_with_query_sequence_invalid_use(query):
+    url = URL("http://example.com")
+    with pytest.raises(TypeError):
```
webknjaz, line 144: the best practice is to always use `match=` arg
```suggestion
    with pytest.raises(TypeError):
```

CHANGES/443.feature
Lonami: I used the fix terminology because it "fixes" an enhancement issue on `aiohttp`, but I've reworded it as "Allow" (I've searched the changelog and there are two other uses of this wording before).

docs/api.rst
webknjaz: Just add a test with `{'c[]': [1, 2]}` making sure that `[]` is preserved non-encoded.

tests/test_update_query.py
Lonami: I think there's no need to test for `"1&a=2"`, because this test is about the use of sequences, so the latter should be enough. We already cover a similar case in `test_with_query_multidict_with_unsafe` and `test_with_query_params`.

CHANGES/443.feature
webknjaz: Right, things aren't very consistent here, unfortunately. But it shouldn't prevent us from trying to do better :)

tests/test_update_query.py
webknjaz: I'd still have it here. Because updates in one part of code can sometimes break others. Better safe than sorry.

tests/test_update_query.py
Lonami: None of the other existing tests do this, so I tried to stay in line with them (and also was unaware of such thing). Made a commit with your suggestion.

tests/test_update_query.py
Lonami: I will update the `test_with_query_params` and `test_with_query_params2` to use `parametrize` too and also add your new case. Does that sound good?

Lonami:
I've also added test for edge cases where the list/tuple are empty or have a single item which we didn't have before. I think it should be ready for review again.

Lonami:
Bumping this now that we're on the weekend. Is there any reason not to merge this other than lack of time?

tests/test_update_query.py
```diff
@@ -117,6 +117,43 @@ def test_with_query_list_int():
     assert str(url.with_query([("a", 1)])) == "http://example.com/?a=1"
 
 
+@pytest.mark.parametrize(
+    "query,expected",
```
webknjaz, line 121:
```suggestion
    ("query", "expected"),
```

tests/test_update_query.py
```diff
@@ -117,6 +117,43 @@ def test_with_query_list_int():
...
+        pytest.param({"&": ["=", 2]}, "?%26=%3D&%26=2", id="quote key and value"),
+        pytest.param({"a": 1, "b": [2, 3]}, "?a=1&b=2&b=3", id="single then list"),
+        pytest.param({"a": [1, 2], "b": 3}, "?a=1&a=2&b=3", id="list then single"),
+        pytest.param({"a": ["1&a=2", 3]}, "?a=1%26a%3D2&a=3", id="ampersand then int"),
+        pytest.param({"a": [1, "2&a=3"]}, "?a=1&a=2%26a%3D3", id="int then ampersand"),
+    ],
+)
+def test_with_query_sequence(query, expected):
+    url = URL("http://example.com")
+    assert url.with_query(query) == URL("http://example.com/" + expected)
```
webknjaz, line 141: Maybe it'll be better to compare strings? Currently you're testing that the main initializer does the same thing as `URL.with_query()` method.
```suggestion
    assert str(url.with_query(query)) == "http://example.com/{expected}".format_map(locals())
```

tests/test_update_query.py
```diff
@@ -196,16 +233,21 @@ def test_with_query_memoryview():
...
 
-def test_with_query_params():
-    url = URL("http://example.com/get")
-    url2 = url.with_query([("key", "1;2;3")])
-    assert str(url2) == "http://example.com/get?key=1%3B2%3B3"
-
-
-def test_with_query_params2():
+@pytest.mark.parametrize(
+    "query,expected",
```
webknjaz, line 237:
```suggestion
    ("query", "expected"),
```

webknjaz:
@Lonami I'd like tests and mentions in all docs for `a[]=1&a[]=2`

Lonami:
Yeah don't mind the extra space there it's fixed in the next commit which I'll push once I polish.

Lonami:
> mentions in all docs for a[]=1&a[]=2

Changing the test to compare `str()` actually did reveal that this escapes `[]` in the key when we shouldn't, although this is a bug from before. That is, this test fails in master:

```python
def test_stuff():
    url = URL("http://example.com/path")
    u2 = url.with_query([("a[]", "3"), ("a[]", "4")])

    assert str(u2) == "http://example.com/path?a[]=3&a[]=4"
```

With:
```
E       AssertionError: assert 'http://examp...D=3&a%5B%5D=4' == 'http://examp...h?a[]=3&a[]=4'
E         - http://example.com/path?a[]=3&a[]=4
E         ?                          ^^    ^^
E         + http://example.com/path?a%5B%5D=3&a%5B%5D=4
E         ?                          ^^^^^^    ^^^^^^

tests/test_update_query.py:228: AssertionError
```

So I suppose it's fine to fix this because people relying on it being escaped was buggy (not intended) behaviour.

Lonami:
Opened #445 to address that.

webknjaz:
As per https://github.com/aio-libs/yarl/pull/445#pullrequestreview-409024181, use encoded values for the square brackets in tests. It is still useful to have tests with non-letter values.

Lonami:
I have "ported" the test using a list of tuples from https://github.com/aio-libs/yarl/pull/445/files#diff-ae226bc0f15cb48ba935d31884c1034dR226 over the new `parametrize` from here, tested also a single key, and fixed the test above to use the quoted braces.

Lonami:
Hooray 🎉, thanks!

PR #445, opened 2020-05-03
Fix [] was being escaped in query key position

Lonami:
## What do these changes do?

These changes fix a bug where `[]` was being escaped when used in the key of queries, namely:

```python
def test_with_query_no_escape_braces():
    url = URL("http://example.com/get")
    url2 = url.with_query([("a[]", "3"), ("a[]", "4")])
    assert str(url2) == "http://example.com/get?a[]=3&a[]=4"
```

Yes, if users were relying on unintended, buggy behaviour (`[]` being escaped) this change may cause issues in their code.

## Related issue number

None as far as I know.

## Checklist

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -575,7 +575,7 @@ section generates a new *URL* instance.
       URL('http://example.com/path?b=2')
       >>> URL('http://example.com/path?a[]=1&b[]=2')\
       ...     .with_query([("a[]", "b"), ("a[]", "c")])
-      URL('http://example.com/?a[]=b&a[]=c')
+      URL('http://example.com/path?a[]=b&a[]=c')
 
    .. versionchanged:: 1.4.3
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -439,6 +439,10 @@ section generates a new *URL* instance.
       >>> URL.build(scheme="http", host="example.com", query_string="a=b")
       URL('http://example.com/?a=b')
 
+      >>> URL.build(scheme="http", host="example.com",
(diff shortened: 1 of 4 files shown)


Lonami:
No idea if the version will be 1.4.3, I used that because the last released one according to https://github.com/aio-libs/yarl/blob/master/CHANGES.rst (as of https://github.com/aio-libs/yarl/commit/2311d0202afc442f45d93005bffe55d5f1f3be15) is 1.4.2.

Lonami:
Another bump for the weekend? (Forgot to do it yesterday.)

docs/api.rst
```diff
@@ -610,11 +627,19 @@ section generates a new *URL* instance.
...
+      URL('http://example.com/path?a=b&c[]=d&c[]=e')
 
    .. versionchanged:: 1.0
 
       All multiple key/value pairs are applied to the multi-dictionary.
 
+   .. versionchanged:: 1.4.3
+
+      The ``'[]'`` characters are not escaped in the keys of the query
+      because some endpoints expect those to be present when specifying arrays.
```
webknjaz, line 641: Do you mind also referencing an RFC document that specifies this?

Lonami: I do not have any RFC on this, I only wrote this based on what you said in https://github.com/aio-libs/yarl/pull/443#discussion_r415751144.

> web-browsers use `varname[]` format for arrays

Lonami:
Okay, thanks for the investigation. So there's nothing for me to fix in #443 with regards to this and `[]` should be escaped.

PR #472, opened 2020-07-04
fix requoting query components

### LINKED ISSUE #426: `with_query` autoencode unexpected result
I've ran into this when some external service sent an escaped character:

```
In [16]: str(URL("/").with_query(foo="bar&baz%26qux"))
Out[16]: '/?foo=bar%26baz%26qux'
```

The distinction between inputs `&` and `%26` is lost:

```
In [19]: URL(x).query
Out[19]: <MultiDictProxy('foo': 'bar&baz&qux')>
```

This makes the library hard to use for arbitrary input.

The specific case where `yarl` broke involved 2 redirects (i.e. nested) where each had own query parameters.

xppt:
## What do these changes do?

These changes fix a bug where query component, passed in a form of mapping or sequence, is unquoted in unexpected way.
```python
def test_with_query_complex_url():
    target_url = "http://example.com/?game=bulls+%26+cows"
    url = URL("/redir").with_query({"t": target_url})
    assert url.query["t"] == target_url
```

User will now be able to use any string as query component in `URL.with_query`, `URL.update_query` and `URL.build`.
Which will be properly encoded.

## Related issue number

Fixes #426 

## Checklist

### DIFF
--- a/CHANGES/426.bugfix
+++ b/CHANGES/426.bugfix
@@ -1,0 +1,1 @@
+Fix a bug where query component, passed in a form of mapping or sequence, is unquoted in unexpected way.
--- a/tests/test_quoting.py
+++ b/tests/test_quoting.py
@@ -93,6 +93,8 @@ def test_default_quoting_percent(quoter):
     assert "%25" == result
     result = quoter(qs=True)("%25")
     assert "%25" == result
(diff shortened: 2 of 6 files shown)


asvetlov:
Thanks for the PR.
Please let me sleep on it before the applying.

PR #475, opened 2020-07-19
Hide Query and QueryVariable type aliases

### LINKED ISSUE #431: Hide Query, QueryType typevars in .pyi
PyCharm 2019.3 does lookup for module attributes from `yarl/__init__.pyi` and as `Query` and `QueryVariable` are declared there, it assumes they exist in the `yarl/__init__.py`, but those two type variables are not declared there. So doing: `from yarl import Query, QueryVariable` is totally legal for PyCharm, but will throw `ImportError` in runtime.

As `Query` and `QueryVariable` sound very much like they're part of API, I think, it will be a rational  move to solve this issue.

Naive solutions: 
1. Declare `__all__` in `.pyi`, which didn't work out for me
2. Add `_` prefix to `Query` and `QueryVariable` making them "protected members" of interface file.
3. Declare same types in `__init__.py`. I don't think this is a good idea.

asvetlov:
Fixes #431


uwinx:
@asvetlov my $0.01 for the future. What do you think about removing all the code form `__init__.py`? which may and probably will come as a breaking change.

We can have `yarl/_url.py` with all the stuff from current `__init__.py` and have single line with import like `from ._url import URL` in new `__init__.py`. IMO, Otherwise yarl has pretty bad public API.

asvetlov:
Do you mean the number of internal names exposed by `__init__.py`?

uwinx:
yes.

asvetlov:
These internal names are not exposed by `__all__`, so the situation is not dramatically bad.
I'm +0 for the change (need to finish all pending PRs first)

PR #483, opened 2020-07-25
Hide the library implementation

asvetlov:
Follow-up for https://github.com/aio-libs/yarl/pull/475#issuecomment-660618807 discussion


webknjaz:
(re-opened to trigger the changelog checker app)

yarl/_quoting_c.c
```diff
@@ -0,0 +1,12380 @@
+/* Generated by Cython 0.29.21 */
```
webknjaz, line 1: Oops

PR #484, opened 2020-07-27
Require typing_extensions for py<3.8 only

mgorny:
## What do these changes do?

They make it possible to run on py3.8+ without external `typing_extensions`. This also fixes support for py3.9 where typing_extensions don't work (yet).

Nope.

## Related issue number

Nope.

## Checklist


webknjaz:
@mgorny thanks!

PR #489, opened 2020-07-31
Fix MANIFEST.in to reflect changed C file names

### LINKED ISSUE #485: File "yarl/_quoting.pyx", line 192, in yarl._quoting._Quoter.__init__
Facing this issue with any version of aiohttp starting from 3.4.0 and with recently released yarl 1.5.0.
Using previous yarl==1.4.2 solves this issue (error is gone).
Thought it was old bug bubbling out: https://github.com/aio-libs/aiohttp/issues/1364, but it seems that this issue is new.

asvetlov:
Fixes #485


asvetlov:
Done, thanks!

CHANGES/485.bugfix
```diff
@@ -0,0 +1 @@
+Fix MANIFEST.in to reflect changed C file names.
```
webknjaz, line 1: I think a better user-facing note would be
```suggestion
Fix including relocated internal ``yarl._quoting_c`` C-extension into published PyPI dists.
```

PR #575, opened 2021-04-16
Fix test with Python 3.9.2+

### LINKED ISSUE #563: Fail test_semicolon_as_separator  with Python 3.9.2 and Python 3.8.8
🐞 **Describe the bug**
Because of new behavior of parse_qsl, test_semicolon_as_separator began to fall - https://bugs.python.org/issue42967

📋 **Logs/tracebacks**
```python-traceback (paste your traceback in the next line)
    def test_semicolon_as_separator():
        u = URL("http://127.0.0.1/?a=1;b=2")
>       assert len(u.query) == 2
E       assert 1 == 2
E         +1
E         -2
```

📋 **Your version of the Python**

```console
$ python --version
Python 3.9.2
```

shadchin:
## What do these changes do?

Fixes #563 

Yes

## Related issue number

https://bugs.python.org/issue42967

## Checklist


yarl/_url.py
```diff
@@ -551,7 +552,21 @@ def query(self):
...
+            or (3, 7, 10) <= sys.version_info < (3, 8)
+            or (3, 8, 8) <= sys.version_info < (3, 9)
+            or sys.version_info >= (3, 9, 2)
+        ):
+            ret = MultiDict(
+                parse_qsl(
+                    self.raw_query_string,
+                    keep_blank_values=True,
+                    separator=self._separator,
+                )
```
webknjaz, line 566: This should address the LGTM alert
```suggestion
                )  # lgtm [py/call/wrong-named-argument]
```

yarl/_url.py
```diff
@@ -551,7 +552,21 @@ def query(self):
         Empty value if URL has no query part.
 
         """
-        ret = MultiDict(parse_qsl(self.raw_query_string, keep_blank_values=True))
+        if (
+            (3, 6, 13) <= sys.version_info < (3, 7)
+            or (3, 7, 10) <= sys.version_info < (3, 8)
+            or (3, 8, 8) <= sys.version_info < (3, 9)
+            or sys.version_info >= (3, 9, 2)
```
webknjaz, line 559: This module is not the right place for such conditionals. We should put `parse_qsl()` compat code into a `_compat` module and then import it here + use w/o conditionals.

yarl/_url.py
```diff
@@ -142,7 +142,7 @@ class URL:
     _PATH_UNQUOTER = _Unquoter(unsafe="+")
     _QS_UNQUOTER = _Unquoter(qs=True)
 
-    def __new__(cls, val="", *, encoded=False, strict=None):
+    def __new__(cls, val="", *, encoded=False, strict=None, separator="&"):
```
webknjaz, line 145: If we start providing this public API to the end-users, we must ensure that users calling it with `separator=";"` on old and new Pythons will get the same results regardless of whether the underlying function in stdlib supports this argument or not.

This means, that we need a compatibility shim in the `_compat` module normalizing the behavior across the board.

tests/test_url_query.py
```diff
@@ -63,7 +63,7 @@ def test_ampersand_as_value():
 
 
 def test_semicolon_as_separator():
-    u = URL("http://127.0.0.1/?a=1;b=2")
+    u = URL("http://127.0.0.1/?a=1;b=2", separator=";")
```
webknjaz, line 66: This will fail under some environments unless the behavior isn't made predictable. Also, we probably need tests with combinations of `;` and `&`.

CHANGES/563.bugfix
```diff
@@ -0,0 +1 @@
+Fix issue with semicolon as a separator in query parameters.
```
webknjaz, line 1: As it stands now, this PR also adds a feature of passing a separator arg to the constructor. So this probably shouldn't be a `.bugfix` file.

webknjaz: Also, this note is supposed to be user-oriented. When somebody reads the changelog, it should give them some idea of what this is about. It currently does not name the issue, nor does it mention the way it's addressed and what the impact is for the users.

webknjaz (changes requested):
Do we really need to change the API of `yarl.URL()`? This PR needs more work introducing better compatibility shims. Otherwise, it may be easier to just always pass an ampersand w/o changing the API.

asvetlov:
Thanks for the PR.
I rather prefer to drop the test and give up the semicolon separator support.
It is a very rarely used mode, I never saw it in the real-life since about 1999.

PR #582, opened 2021-05-11
added bytes conversion

ChristianMertes:
## What do these changes do?

Added `__bytes__()` magic method so that `bytes(url)` will work and use optimal ASCII encoding, as URL string representations are always encoded already.

While I was at it, I fixed some typos I saw and I had to add a mypy ignore for the tests to work for me. There is one test that fails for me (`test_semicolon_as_separator`) because I was using a Python > 3.9.2 and `parse_qsl()` [changed its behavior](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.parse_qsl).

Existing functionality shouldn't be affected.

## Related issue number

n/a

## Checklist


ChristianMertes:
It is interesting that said test also fails for the other Python versions. I am not convinced this has anything to do with the proposed changes, however....

yarl/_url.py
```diff
@@ -272,6 +272,9 @@ def __str__(self):
     def __repr__(self):
         return "{}('{}')".format(self.__class__.__name__, str(self))
 
+    def __bytes__(self):
+        return str(self).encode("ascii")
```
MyrikLD, line 276: Why not UTF-8?

ChristianMertes: It's been a while but IIRC the string representation already didn't use any non-ascii characters as everything is URL-encoded, right? Or can there be Unicode in the domain name?

asvetlov: No, unicode domains are idna encoded, e.g. `xn--h1aagokeh.xn--p1ai`

setup.cfg
```diff
@@ -30,3 +30,5 @@ ignore_missing_imports = true
 [mypy-pytest]
 ignore_missing_imports = true
 
+[mypy-multidict]
+ignore_missing_imports = true
```
asvetlov, line 34:
```suggestion
```

multidict *provides* type annotations

asvetlov:
Suggested typos are applied to master in a separate commit, thanks

asvetlov (approved):

Issue #12, opened 2016-09-17
/-joining of urls can produce multiple slashes


mgedmin:
```python
>>> from yarl import URL
>>> URL('http://example.com/api/') / 'sub/path'
URL('http://example.com/api//sub/path')
```

Should that double slash be there?

Other corner cases:

```python
>>> URL('http://example.com/api') / '/sub/path'
URL('http://example.com/api//sub/path')
>>> URL('http://example.com/api/') / '/sub/path'
URL('http://example.com/api///sub/path')
```

These examples work the way I expect:

```python
>>> URL('http://example.com/') / 'sub/path'
URL('http://example.com/sub/path')
>>> URL('http://example.com') / 'sub/path'
URL('http://example.com/sub/path')
```

These do not:

```python
>>> URL('http://example.com/') / '/sub/path'
URL('http://example.com//sub/path')
>>> URL('http://example.com') / '/sub/path'
URL('http://example.com//sub/path')
```

Bug or explicit design decision?

asvetlov:
Thank for pointing on quirks out!
`URL('http://example.com/api/') / 'sub/path'` shouldn't produce double slash.
But appending a path starting from slash should be forbidden explicitly I think.
For `URL('http://example.com/api') / '/sub/path'` case the right part looks like an absolute path, appending absolute path to absolute URL is wrong from my perspective.

Would you make a PR with fix?

mgedmin:
I'm ambivalent about `URL(...) / '/subpath/starting/with/a/slash'`.  Sometimes it seems to make sense, e.g. when you're considering a WSGI application mounted at some URL prefix, and you're thining about API endpoints in terms of `/path/from/the/app/root`.  I don't personally care strongly about it.  (I don't in fact use yarl anywhere yet. :wink:)

asvetlov:
Let's forbid it now.
If we'll change our mind in future we may relax the rule without any backward incompatibility issue.
But restricting is much more painful process :)

mgedmin:
Do you want me to update my PR, or ... ?

(I'm feeling kind of lazy.)

asvetlov:
If you would to update it or create new one with preventing absolute paths -- please go ahead.
If not -- I'll just merge existing code and will try to find a time for fixing by myself.

mgedmin:
Go ahead and merge then.

Thanks!

PR #619, opened 2021-09-20
Problem with joining urls starting with /

MyrikLD:
## What do these changes do?

```python
>>> from yarl import URL
>>> URL('http://example.com/api/') / '/sub/path'
URL('http://example.com/api/sub/path')
```

The user will be happy to receive valid url instead of exception

## Related issue number

https://github.com/aio-libs/yarl/issues/12

## Checklist


asvetlov:
I'm not sure if the proposed change is correct.
`pathlib` has a different behavior:
```
In [1]: from pathlib import Path

In [2]: Path('/a/b/c/') / '/d/e/f'
Out[2]: PosixPath('/d/e/f')
```
URL join also replaces the absolute path according to RFC: https://datatracker.ietf.org/doc/html/rfc3986#section-5.2.2

mjpieters:
I am closing this PR as the behaviour goes counter to how pathlib and os.path.join() work.

PR #665, opened 2022-01-16
Docs: fix broken internal refs, fix broken intersphinx cross-refs to multidict, minor styling fixes

hoefling:
## What do these changes do?

This PR contains only documentation improvements, no code changes.

1. Broken internal refs to e.g. `URL.human_repr()` are fixed.
2. Broken crossrefs are fixed: dropped duplicate `intersphinx` config, fixed broken URL to `multidict` docs.
3. Markdown styling that hasn't the same meaning in ReST fixed.

No

## Related issue number

Didn't create one since the changes are trivial IMO

## Checklist

- [ ] ~Unit tests for the changes exist~ not needed, but an extra job in the `ci` action could watch over the docs and fail if new errors are introduced


hoefling:
[Docs built from this PR](https://yarl--665.org.readthedocs.build/en/665/)

CHANGES/665.doc
```diff
@@ -0,0 +1,3 @@
+Fixed broken internal references to :meth:`~URL.human_repr()`.
+Fixed broken external references to :doc:`multidict:index` docs.
+Minor styling improvements.
```
webknjaz, line 3: Looks like there are 3 separate notes so they should be in separate files and contain information meaningful to the end-users facing the changelog for the first time. They should tell concrete things, describe the effects of updating from the previous version.
Because the changelog describes the things that happened since the previous release, in general, it should use something like the past tense.
The changelog shouldn't refer to internal implementation details that the end-users aren't expected to be familiar with. Telling them that there were arbitrary unspecified improvements of an arbitrary concept of style (writing style? code style? visual/UX/docs theme style?) is not particularly helpful.

Looks like the first two lines are reasonably useful, I'd put them into two separate files. To avoid filename collision, Towncrier supports files with numeric suffixes after the type, like `665.doc.1` + `665.doc.2` (or `665.doc.1.rst` + `665.doc.2.rst`).

CHANGES/665.doc
```diff
@@ -0,0 +1,3 @@
+Fixed broken internal references to :meth:`~URL.human_repr()`.
```
webknjaz, line 1:
```suggestion
Fixed broken internal references to :meth:`~URL.human_repr`.
```

webknjaz (commented):
Looks like this patch could be coupled with adding strictness to the Sphinx setup. I like having `default_role = 'any'` and [`nitpicky = True`](https://github.com/aio-libs/aiohttp/blob/6aafe3e/docs/conf.py#L387) in `conf.py` for this.
Also, adding [`-W --keep-going -n`](https://github.com/aio-libs/aiohttp/blob/6aafe3e/Makefile#L177) to the `sphinx-build` invocation is as helpful.

asvetlov (approved):
LGTM, thanks!

PR #671, opened 2022-01-23
Fix query type annotation

layday:
## What do these changes do?
Add missing type arg to `query` following https://github.com/aio-libs/multidict/commit/48f29bd3a4339ac1474a21bcc35f39254c74e296.

No.

## Related issue number
—

## Checklist

### DIFF
--- a/CHANGES/671.misc
+++ b/CHANGES/671.misc
@@ -1,0 +1,1 @@
+Fixed `query` type annotation.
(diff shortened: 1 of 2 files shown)


asvetlov:
Please let me take a look at multidict typing first.
Abstract types (`MultiMapping` and `MutableMultiMapping`) should have parametrized keys.
But for concrete containers, key types are `str` and `istr` correspondingly.

asvetlov:
I believe https://github.com/aio-libs/multidict/pull/682 is the correct fix, it doesn't require changing of all dependent libraries, e.g. `yarl`.

Thank you very much for the quick point at the problem!

layday:
Thanks, that looks like a better fix.

PR #700, opened 2022-03-28
Avoid running the workflow twice on PR pushes

mjpieters:
When someone force-pushes to a PR or adds a new commit, the workflow would run
_twice_, as both `push` and `pull_request` events would trigger.

The default for `types` is `[opened, reopened, synchronize]`; the latter is the
type that overlaps with `push`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,8 +10,14 @@ on:
     branches:
       - master
       - ?.?*
+    types: [opened, reopened]
   schedule:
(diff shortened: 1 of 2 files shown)


webknjaz:
> When someone force-pushes to a PR or adds a new commit, the workflow would run
> _twice_, as both `push` and `pull_request` events would trigger.

Important clarification: this only applies to PRs from within the upstream repo, which is why adopted a practice of making forks even where I have a commit bit. This is what we also enforce at Ansible…

PR #704, opened 2022-03-29
Add URL.joinpath(*elements, encoding=False) to build a URL with multiple new path elements

mjpieters:
## What do these changes do?

This adds a new method, `URL.joinpath()`, which creates a new URL object with
the arguments added as path elements:

```python
>>> url = URL("http://example.com/foo")
>>> url.joinpath("bar", "ham", "spam/sub")
URL("http://example.com/foo/bar/ham/spam/sub")
```

The method accepts an `encoded` flag so you can pass in already-encoded path
elements.

There are no changes in behaviour for existing methods, but I did refactor
`.__truediv__` to share code with the implementation for `.joinpath()`.

## Checklist


chey:
How different is this new method compared to using the following?

```python
>>> url = URL("http://example.com/foo")
>>> url / "bar" / "ham" / "spam/sub"
URL('http://example.com/foo/bar/ham/spam/sub')
```

mjpieters:
> How different is this new method compared to using the following?

You can't use that expression with a dynamic list of path elements, not without having to use an explicit a loop. Without this method you can't extend a series of URLs in a list comprehension, say:

>>> import random
>>> from yarl import URL
>>> urls = [URL("http://example.com/foo"), URL("http://example.com/bar")]
>>> names = ("terry_jones", "eric_idle", "john_cleese", "micheal_palin", "graham_chapman", "terry_gilliam")
>>> [url.joinpath("spam", random.choice(names)) for url in urls]
[URL("http://example.com/foo/spam/micheal_palin"), URL("http://example.com/bar/spam/eric_idle")]

This is modelled on [`pathlib.Path.joinpath()`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.joinpath), where `Path()` objects of course also support `/` as a joining operator.

docs/api.rst
```diff
@@ -817,6 +817,30 @@ The path is encoded if needed.
       >>> url
       URL('http://example.com/path/%D1%81%D1%8E%D0%B4%D0%B0')
 
+.. method:: URL.joinpath(*other, encoded=False)
+
+   Construct a new URL by with all ``other`` elements appended to
+   *path*, and cleaned up *query* and *fragment* parts.
+
+   Passing ``encoded=True`` parameter prevents path element auto-encoding, user is
```
webknjaz, line 825: maybe?
```suggestion
   Passing ``encoded=True`` parameter prevents path element auto-encoding, the user is
```

webknjaz: or even?
```suggestion
   Passing ``encoded=True`` parameter prevents path element auto-encoding, the caller is
```

tests/test_url.py
```diff
@@ -767,6 +767,101 @@ def test_div_with_dots():
...
+
+
+def test_joinpath_with_slash():
+    url = URL("http://example.com/path/")
+    assert str(url.joinpath("to")) == "http://example.com/path/to"
+
+
+def test_joinpath_path_starting_from_slash_is_forbidden():
+    url = URL("http://example.com/path/")
+    with pytest.raises(ValueError):
```
webknjaz, line 795: It is strongly recommended to make use of the `match` argument when testing with `pytest.raises()` and `pytest.warns()`. Otherwise, this has a chance of causing false positives when it's unintended.

tests/test_url.py
```diff
@@ -767,6 +767,101 @@ def test_div_with_dots():
     assert url.raw_path == "/path/to"
 
 
+# joinpath
+
+
+def test_joinpath_root():
```
webknjaz, line 773: Most of the tests added here can be collapsed with the use of `@pytest.mark.parametrize`. Could you implement this? For the output readability, it's best to declare human-readable test IDs.

yarl/_url.py
```diff
@@ -701,11 +683,45 @@ def _validate_authority_uri_abs_path(host, path):
                 "Path in a URL with authority should start with a slash ('/') if set"
             )
 
+    def _make_child(self, segments, encoded=False):
+        # add segments to self._val.path, accounting for absolute vs relative paths
```
webknjaz, line 687: Why isn't this a docstring? It explains what the code does. Comments are supposed to explain motivation/unobvious decisions but the behavior description should go into a docstring.

yarl/_url.py
```diff
@@ -701,11 +683,45 @@ def _validate_authority_uri_abs_path(host, path):
...
     def _normalize_path(cls, path):
-        # Drop '.' and '..' from path
+        # Drop '.' and '..' from str path
 
         segments = path.split("/")
+        return "/".join(cls._normalize_path_segments(segments))
+
+    @classmethod
+    def _normalize_path_segments(cls, segments):
+        # Drop '.' and '..' from a sequence of str segments
```
webknjaz, line 723: This should be a docstring too.

yarl/_url.py
```diff
@@ -701,11 +683,45 @@ def _validate_authority_uri_abs_path(host, path):
...
     @classmethod
     def _normalize_path(cls, path):
-        # Drop '.' and '..' from path
+        # Drop '.' and '..' from str path
 
         segments = path.split("/")
+        return "/".join(cls._normalize_path_segments(segments))
+
+    @classmethod
+    def _normalize_path_segments(cls, segments):
```
webknjaz, line 722: This shouldn't be a `@classmethod`. It better be a pure function, or if it really needs to be under the class, use `@staticmethod` since `cls` is unused.

webknjaz:
@mjpieters I've rebased the PR and it still passes but there's still a few things I'd like improved so I won't be merging it right now.

yarl/_url.py
mjpieters: Not sure, I think I meant it to be a docstring..

yarl/_url.py
mjpieters: I was following the general style of the class here; e.g. `_encode_host()` can also be made a staticmethod.

mjpieters:
@webknjaz all comments addressed now.

asvetlov:
The pull request looks good but merge conflict should be resolved first.

mjpieters:
> The pull request looks good but merge conflict should be resolved first.

Done!

asvetlov:
Thanks @mjpieters !

Issue #723, opened 2022-05-10
update_query(None) doesn't clear the query


altvod:
### Describe the bug

`url.update_query(None)` doesn't clear the query

### To Reproduce

In reality it does this:
```
>>> URL('https://mysite.org?what=this').update_query(None)
URL('https://mysite.org/?what=this')
```

### Expected behavior

The [doc says](https://yarl.readthedocs.io/en/latest/api.html#yarl.URL.update_query):
```
URL.update_query(query)
URL.update_query(**kwargs)
Returns a new URL with query part updated.
...
Clear query if None is passed.
```

### Logs/tracebacks

```python-traceback
-
```

### Python Version

```console
$ python --version

Python 3.10.0
```

### multidict Version

```console
$ python -m pip show multidict

Version: 6.0.2
```

### yarl Version

```console
$ python -m pip show yarl

Version: 1.7.2
```

### OS

ubuntu

### Code of Conduct

- [X] I agree to follow the aio-libs Code of Conduct

dereckt:
Hi, I'd like to work on this issue since it seems no progress has been made recently. Let me know if this is a problem.

PR #737, opened 2022-06-11
Fix update_query(None) doesn't clear the query

slymit:
## What do these changes do?

This fix bug when method url.update_query(None) doesn't clear the query.

This changes clear the query if None is passed to method update_query().

## Related issue number

#723 

## Checklist

### DIFF
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -983,8 +983,11 @@ class URL:
         """Return a new URL with query part updated."""
         s = self._get_str_query(*args, **kwargs)
         new_query = MultiDict(parse_qsl(s, keep_blank_values=True))
-        query = MultiDict(self.query)
-        query.update(new_query)
+        if args and args[0] is None:
+            query = new_query
+        else:
+            query = MultiDict(self.query)
+            query.update(new_query)
 
         return URL(self._val._replace(query=self._get_str_query(query)), encoded=True)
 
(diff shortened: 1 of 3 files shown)


yarl/_url.py
```diff
@@ -983,8 +983,11 @@ def update_query(self, *args, **kwargs):
         """Return a new URL with query part updated."""
         s = self._get_str_query(*args, **kwargs)
         new_query = MultiDict(parse_qsl(s, keep_blank_values=True))
-        query = MultiDict(self.query)
-        query.update(new_query)
+        if args and args[0] is None:
+            query = new_query
+        else:
+            query = MultiDict(self.query)
+            query.update(new_query)
```
mjpieters, line 990: Instead of re-testing `args`, check if `s == ""`. `_get_str_query()` will only ever return` ""` if `None` was passed in. We can make use of this to avoid unnecessary round-tripping the query string to `MultiDict` and back:

```suggestion
        s, query = self._get_str_query(*args, **kwargs), None
        if s:
            new_query = MultiDict(parse_qsl(s, keep_blank_values=True))
            query = MultiDict(self.query)
            query.update(new_query)
```

By setting `query = None` at the start we save on parsing an empty string to an empty `MultiDict()` and back again; note that the last line uses `self._get_str_query(query)` again so defaulting `query` to `None` means that `self._val._replace(query="")` is called (where `self._val` is a [`urllib.parse.SplitResult()` named tuple](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.SplitResult)).

mjpieters (changes requested):

asvetlov:
Fixed by #792 I guess

PR #782, opened 2022-10-23
Change to SPDX conform license string

elprimato:
## What do these changes do?

I suggest to change the license string in the package information to an [SPDX parsable license expression](https://spdx.org/licenses/).
This makes it easier for downstream users to get the license information directly from the package metadata.

No changes in functionality.

## Checklist

The checklist items are not applicable for this trivial change.


webknjaz:
To my knowledge, there's no PEP requiring a specific format in that field. It's free-form. Moreover, per specs, it should be used as a fallback if one doesn't enclose the license files (we do).

elprimato:
Yes, currently there is still free-form allowed here.
However, quoting from another [PR of the astropy project](https://github.com/astropy/astropy/pull/13875):

In Python, there is [PEP 639](https://peps.python.org/pep-0639/) which would add a standard 'License-Expression' metadata field. The discussion around this PEP seems to have gone stale in 2021.

This merge request is in lieu of that PEP coming soon. Unless you as project maintainers think the proposed change in this PR has downsides (which you are entitled to!), I would ask that you accept this change so that it's possible for users to attempt to parse your license metadata as an SPDX License Expression.

If PEP 639 is accepted, this package would be ready for the change just by changing the license key to license-expression or whatever key the PEP lands on.

RazerM:
> It's free-form and is intended to be presented to humans on PyPI.

Indeed, it is also true that "Apache 2" is not ambiguous, unlike some projects who may have just had "BSD" and we've contributing changing them to the unambiguous "BSD-3-Clause" etc.

I would argue that "Apache-2.0" satisfies all of the following:

- human readable (already the case)
- unambiguous (already the case)
- Parsable as an SPDX License Expression 🆕

asvetlov (approved):

asvetlov:
thanks

Issue #781, opened 2022-10-20
CPython 3.11 wheels on PyPI


altendky:
### Is your feature request related to a problem?

Just helping bump other projects on 3.11 support as I try to add it to my own.

### Describe the solution you'd like

CPython 3.11 wheels available on PyPI.

### Describe alternatives you've considered

Presently our CI is building the wheels from the sdist on PyPI but this doesn't work well for all users that don't have development tools installed.

### Additional context

Thanks!

### Code of Conduct

- [X] I agree to follow the aio-libs Code of Conduct

webknjaz:
I was planning to do this for yarl+multidict+aiohttp on Oct 24, once it's final.

Senpos:
Hi @webknjaz!
Sorry for the ping, but I just wanted to let you know that the final version of CPython 3.11 has been released, please, consider building wheels. 🙏 Thanks! <3

webknjaz:
Yep, I remember. It's still in my backlog.

AmirHmZz:
Any updates on this ?

alexwilliams0712:
Do we have a timescale at all? What needs doing?

altendky:
Just added https://github.com/aio-libs/yarl/pull/790 to simplify the `pyver:` for 3.11.  Also noted that there isn't a changelog entry for adding 3.11 support.  I'm not sure what PR to credit for it, or maybe multiple, but I found these...

https://github.com/aio-libs/yarl/pull/759
https://github.com/aio-libs/yarl/pull/740
https://github.com/aio-libs/yarl/pull/716
https://github.com/aio-libs/yarl/commit/0c6ec42308d0ec2cab268d236a2f6dc2577fc162

Looks like nothing has changed since v1.8.1 per https://github.com/aio-libs/yarl/compare/v1.8.1..master so I guess any 3.11 wheels could just be added to the v1.8.1 release, if wanted.

Anything I could help with?  Or just need time for the owners to get through the last steps...

webknjaz:
Shipped via v1.8.2

PR #790, opened 2022-11-28
Switch Python 3.11 to stable release in CI

altendky:
## What do these changes do?

Just tidying the now-released CPython 3.11 `pyver` entry.

No.

## Related issue number

Adjacent to https://github.com/aio-libs/yarl/issues/781.


CHANGES/790.misc.rst
```diff
@@ -0,0 +1 @@
+
```
webknjaz, line 1: An empty change note is probably not the best idea. Either fill it out or I'll apply a label for the bot to just ignore it. I'm actually not sure what Towncrier would do with an empty fragment...

altendky: I can certainly remove it, but it is a normal option with towncrier.  That way it's all 'in the system' rather than you having to add external tooling to sometimes ignore the check via a GitHub PR label as it sounds like you have.  I'd link to the twisted contributing docs but...  the link I found is 404.

You don't appear to be configuring custom sections in the changelog.

https://github.com/aio-libs/yarl/blob/68e7d9312ad9eccb695dbc174a4705e9a238c1fa/pyproject.toml
```toml
[tool.towncrier]
package = "yarl"
filename = "CHANGES.rst"
directory = "CHANGES/"
title_format = "{version} ({project_date})"
issue_format = "`#{issue} <https://github.com/aio-libs/yarl/issues/{issue}>`_"
```

So you would be using the defaults where, as I recall, `misc` has `showcontent` set to false.

https://towncrier.readthedocs.io/en/stable/configuration.html?highlight=showcontent#custom-fragment-types
> `showcontent`: Whether if the fragment contents should be included in the news file. If omitted, it defaults to true

So, you get just a list of the issue/pr/file numbers as at https://github.com/aio-libs/yarl/blob/68e7d9312ad9eccb695dbc174a4705e9a238c1fa/CHANGES.rst#misc-1.  Note that if you follow those, their content isn't empty, just ignored.  https://github.com/aio-libs/yarl/pull/647/files#diff-3c671c897bfbc0b283bbb224c8784fda9364a40d0233748d2606c6b78faa0394

altendky: Fun, so GitHub let me edit the 'empty' file that actually had 'two lines' and delete one and commit it and it still seems to be the same file unchanged...  But, you've got a linter complaining about it so I'll go ahead and delete it since that seems to be process here.

altendky:
Side note, what do the `~` etc do anyways?  I didn't see it mentioned at https://github.com/actions/setup-python/tree/b80efd6bc5dcdc82c015d69ecd3e39320d0095e6.  To get pre- or full releases, I use the monstrosity at https://github.com/pytest-dev/pytest-twisted/blob/56991132000791cf89283ac5aa033f3469c63e35/.github/workflows/ci.yml#L154-L167 and we have that in a wrapper action we use at work.

CHANGES/790.misc.rst
altendky: Not to be pounding on this...  but I got the doc link so may as well finalize the references here.

https://docs.twisted.org/en/latest/development/dev-process.html#release-notes-management
> **misc** - Issues which are very minor and not worth summarizing outside of the git changelog. These should be empty (their contents will be ignored)

And the PR fixing the doc link is an example of an empty misc fragment.

https://github.com/twisted/twisted/pull/11777/files

webknjaz: FWIW I've been considering showing that content at some point. Too many people misuse these fragments and put really useful stuff there. And it's hard to catch for humans during reviews.

webknjaz:
@altendky 

`~3.11.0-0` corresponds to `3.11.0-alpha - 3.11.0` — pre-releases up to and including the first stable release of the range. My understanding is that it comes from the nodejs community since all of the official actions are written in JS and they've sorta adopted the corresponding versioning schemes.

The tilda variant has been suggested by the GitHub stuff on the issue tracker of `setup-python` at some point, among the issues requesting things like `3.11` to behave the same. I've argued that it weird to force such scheme upon Pythonistas because it's foreign to us.
I think that there's a PR making `3.11` turn into "stable `3.11` with a fallback to `3.11-dev`". I'm not watching that repo closely at the moment so I can't find the links quickly.

webknjaz (approved):

CHANGES/790.misc.rst
altendky: It's all in your intent in the setup.  Nice and configurable.

To make fragments a bit more reviewable, at least in terms of viewing them rendered in case there's any markup errors, I added a step to dump the draft output to the GitHub workflow summary.

https://github.com/Chia-Network/chia-blockchain/pull/13882/files#diff-b1b82537432942b954fa8846c9e80709d0d919428ed1e1692d7c0bc4bc8271ecR112-R114
```yaml
    - name: Create single version changlog summary
      run: |
        towncrier build --yes --draft >> $GITHUB_STEP_SUMMARY
```
(not sure what the `--yes` is doing there...)

Sample output from this at https://github.com/Chia-Network/chia-blockchain/actions/runs/3445535929#summary-9436412167.

webknjaz: @altendky yea, I actually often make the integration with both Sphinx (via by extension @ https://github.com/sphinx-contrib/sphinxcontrib-towncrier — can error out in nitpicky mode) and GH job summary (requires pandoc for RST->md). See:
- https://github.com/ansible/pylibssh/actions/runs/3533848038#summary-9670210175
- https://github.com/ansible/pylibssh/actions/runs/3533848038/workflow#L469-L516
- https://ansible-pylibssh.rtfd.io/changelog

Issue #746, opened 2022-07-02
Improper validation of the port in `URL.with_port`


decorator-factory:
### Describe the bug

`URL.with_port` doesn't fully validate its argument. It is possible to set a boolean port, a negative port, or a port greater than 65535. Additionally, if the provided port is `0`, it is treated as `None` (i.e. removes the port).

Notably, a `bool` port or a port out of range will cause an error when trying to retrieve the port:
```py
from yarl import URL
url = URL("http://example.com")
true_url = url.with_port(True)
print(true_url.port)  # ValueError here
```

### To Reproduce

1. Install `yarl==1.7.2`
2. Run the following code:
```py
from yarl import URL
url = URL("http://example.com")
print(url.with_port(True))
print(url.with_port(0))
print(url.with_port(-1))
print(url.with_port(99999))
```

### Expected behavior

1. `yarl.URL("http://example.com").with_port(True)` raises a `TypeError`
2. `yarl.URL("http://example.com").with_port(0)` raises a `ValueError`
3. `yarl.URL("http://example.com").with_port(-1)` raises a `ValueError`
4. `yarl.URL("http://example.com").with_port(99999)` raises a `ValueError`

### Logs/tracebacks

Output of the code in `To Reproduce`
```
http://example.com:True
http://example.com
http://example.com:-1
http://example.com:99999
```

### Python Version

```console
CPython 3.9.8
```

### multidict Version

```console
6.0.2
```

### yarl Version

```console
1.7.2
```

### OS

Manjaro Linux

### Code of Conduct

- [X] I agree to follow the aio-libs Code of Conduct

webknjaz:
Thanks! Feel free to send a PR with tests for these cases.

dereckt:
Hi, I'm looking into this issue. Almost got a PR ready.

dereckt:
Hi I've created a draft for the PR, though I'm unsure as how to handle the case with the invalid port 0. How should port 0 be handled?

asvetlov:
Fixed by #793

PR #793, opened 2022-12-05
Fix to URL.with_port() input filtering

dereckt:
## What do these changes do?
These changes fix the improper validation issues with URL.with_port(). Essentially, it restricts ports to be from 0 to 65535 (tcp limit), and fixes the weird handling of the 0 port

The URL.with_port() function will now throw errors when Booleans or invalid ports are arguments.
## Related issue number

#746 

## Implementation Inquiry
Currently, this is set up to allow port 0 to function and not throw errors. From what I looked up it seems to be possible to use the 0 port (officially it does "not exist"), though since it's a special case and issue #746 says that the output should throw an error, I'm unsure if I'm handling it correctly.

## Checklist

- [ ] Documentation reflects the changes - Documentation does not go into specifics for input/output on this function

### DIFF
--- a/CHANGES/793.bugfix.rst
+++ b/CHANGES/793.bugfix.rst
@@ -1,0 +1,1 @@
+Added some input restrictions on with_port() function to prevent invalid boolean inputs or out of valid port inputs.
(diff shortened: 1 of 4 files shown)


asvetlov (approved):
LGTM

Issue #808, opened 2023-01-09
`yarl.URL.build()` fails when `host=None`


paulpapacz:
### Describe the bug

When building a URL using `yarl.URL.build()` the command will fail with exception:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/path/to/virtualenv/lib/python3.10/site-packages/yarl/_url.py", line 248, in build
    cls._validate_authority_uri_abs_path(host=host, path=path)
  File "/path/to/virtualenv/lib/python3.10/site-packages/yarl/_url.py", line 706, in _validate_authority_uri_abs_path
    if len(host) > 0 and len(path) > 0 and not path.startswith("/"):
TypeError: object of type 'NoneType' has no len()
```

### To Reproduce

`yarl` version `1.8.2`

```python
from yarl import URL

URL.build(scheme="http", host=None)
```

### Expected behavior

Either passing without problems or some meaningful error/exception indicating that `host` cannot be `None`.

### Logs/tracebacks

```python-traceback
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/path/to/virtualenv/lib/python3.10/site-packages/yarl/_url.py", line 248, in build
    cls._validate_authority_uri_abs_path(host=host, path=path)
  File "/path/to/virtualenv/lib/python3.10/site-packages/yarl/_url.py", line 706, in _validate_authority_uri_abs_path
    if len(host) > 0 and len(path) > 0 and not path.startswith("/"):
TypeError: object of type 'NoneType' has no len()
```

### Python Version

```console
$ python --version
Python 3.10.6
```

### multidict Version

```console
$ python -m pip show multidict
Name: multidict
Version: 6.0.4
Summary: multidict implementation
Home-page: https://github.com/aio-libs/multidict
Author: Andrew Svetlov
Author-email: andrew.svetlov@gmail.com
License: Apache 2
Location: /path/to/virtualenv/lib/python3.10/site-packages
Requires: 
Required-by: aiohttp, yarl
```

### yarl Version

```console
$ python -m pip show yarl
Name: yarl
Version: 1.8.2
Summary: Yet another URL library
Home-page: https://github.com/aio-libs/yarl/
Author: Andrew Svetlov
Author-email: andrew.svetlov@gmail.com
License: Apache 2
Location: /path/to/virtualenv/lib/python3.10/site-packages
Requires: idna, multidict
Required-by: aiohttp, template-api
```

### OS

[...]

PR #809, opened 2023-01-09
`URL.build()`: Raise `TypeError` if `host` argument is `None`

paulpapacz:
## What do these changes do?

Make `URL.build()` raise a `TypeError` if `host` argument is `None`

No.

## Related issue number

#808

## Checklist

### DIFF
--- a/CHANGES/808.bugfix.rst
+++ b/CHANGES/808.bugfix.rst
@@ -1,1 +1,1 @@
-Made ``URL.build()`` raise a ``TypeError`` if the ``host`` argument is ``None``.
+Made :py:meth:`URL.build` raise a :py:exc:`TypeError` if the ``host`` argument is :py:data:`None` — by :user:`paulpapacz`.
--- a/CHANGES/808.bugfix.rst
+++ b/CHANGES/808.bugfix.rst
@@ -1,0 +1,1 @@
+Made ``URL.build()`` raise a ``TypeError`` if the ``host`` argument is ``None``.
--- a/tests/test_url_build.py
+++ b/tests/test_url_build.py
@@ -239,6 +239,11 @@ def test_build_with_authority_with_path_without_leading_slash():
         URL.build(scheme="http", host="example.com", path="path_without_leading_slash")
 
 
+def test_build_with_none_host():
+    with pytest.raises(TypeError, match="NoneType is illegal for.*host"):
+        URL.build(scheme="http", host=None)
+
+
 def test_build_with_none_path():
     with pytest.raises(TypeError):
         URL.build(scheme="http", host="example.com", path=None)
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -240,12 +240,13 @@ class URL:
         if (
             scheme is None
             or authority is None
+            or host is None
             or path is None
             or query_string is None
             or fragment is None
         ):
             raise TypeError(
-                'NoneType is illegal for "scheme", "authority", "path", '
+                'NoneType is illegal for "scheme", "authority", "host", "path", '
                 '"query_string", and "fragment" args, use empty string instead.'
             )
 


tests/test_url_build.py
```diff
@@ -239,6 +239,11 @@ def test_build_with_authority_with_path_without_leading_slash():
         URL.build(scheme="http", host="example.com", path="path_without_leading_slash")
 
 
+def test_build_with_none_host():
+    with pytest.raises(TypeError):
```
webknjaz, line 243: It is always important to use the `match` argument with `pytest.raises()`.

webknjaz (commented):
FWIW I there are valid cases for not having a hostname. For example, this is how one would represent local filesystem paths or URLs, or UNIX file socket URLs, or abstract socket URLs. Hostnames are needed where a URI would represent an IP-network-accessible resource but a generic case would include things like `http+unix:///var/tmp/aiohttp-server.sock` even for web apps that are intended to operate locally. YARL is a generic lib so we shouldn't ignore use-cases that exist outside web. There's `file:///etc/hosts` or `git+ssh://../another-repo/.git` that are valid URLs. And there are URLs that are used in protocols targeting `xdg-open`/MIME thingies. Like `steam://some-app?params` or `twitter://username` etc.

If we want to force people pass a non-`None` host, we should at least make sure there are tests for these cases allowing the end-users to construct corresponding URLs with something like an empty string, for example.

webknjaz:
But I guess having a `None` value representing “no hostname” seems to be better than an empty string that might have a different semantic meaning.
@Dreamsorcerer WDYT?

tests/test_url_build.py
paulpapacz: Added a `match` now but none of the other test cases uses it (just FYI).

webknjaz: That's unfortunate. And a chance to contribute more refactoring PRs for someone.

CHANGES/808.bugfix.rst
```diff
@@ -0,0 +1 @@
+Made ``URL.build()`` raise a ``TypeError`` if the ``host`` argument is ``None``.
```
webknjaz, line 1: It should be possible to link these, thanks to intersphinx:
```suggestion
Made :py:meth:`URL.build` raise a :py:exc:`TypeError` if the ``host`` argument is :py:data:`None` — by :user:`paulpapacz`.
```

webknjaz:
Oh, and one other case is relative URIs which only need path+qs.

Dreamsorcerer:
> But I guess having a `None` value representing “no hostname” seems to be better than an empty string that might have a different semantic meaning. @Dreamsorcerer WDYT?

Looks like the current situation is that it's not designed for it. At a glance, it looks much the same as the other parameters in that block, and fails if you try to use `None`:
```
>>> yarl.URL.build(host=None)  
  File "<stdin>", line 1, in <module>
  File "/home/ubuntu/.local/lib/python3.10/site-packages/yarl/_url.py", line 246, in build
    cls._validate_authority_uri_abs_path(host=host, path=path)
  File "/home/ubuntu/.local/lib/python3.10/site-packages/yarl/_url.py", line 699, in _validate_authority_uri_abs_path
    if len(host) > 0 and len(path) > 0 and not path.startswith("/"):
TypeError: object of type 'NoneType' has no len()
```

I don't think there's any meaningful different to `""` or `None` representing "no hostname", the result would surely look the same, but would require extra code to convert the `None`.

So, this change looks fine to me. We should get type annotations added though, which would make it more explicit.

Dreamsorcerer:
Nevermind, it use stub files. This is already typed as `str`: https://github.com/aio-libs/yarl/blob/master/yarl/__init__.pyi#L59

webknjaz (approved):
@Dreamsorcerer I'll leave the final merge to you if you don't mind.

Issue #821, opened 2023-02-06
Incorrect handling of empty host in absolute URL


kenballus:
### Describe the bug

Absolute URLs are permitted to have empty hosts in RFC 3986.

Relevant grammar rules:
```
   host          = IP-literal / IPv4address / reg-name
   reg-name      = *( unreserved / pct-encoded / sub-delims )
```

Thus, a URL like `a://:1` conforms to the standard.
However, yarl rejects this URL.

`urllib3`, CPython `urllib`, `rfc3986`, `furl`, and `hyperlink` all correctly handle this situation.

### To Reproduce

Try running the following snippet:
```python3
>>> yarl.URL("a://:1")
```

### Expected behavior

The parse should have succeeded, resulting in `URL('a://:1')`.

### Logs/tracebacks

```python-traceback
This is the output from the snippet:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.10/site-packages/yarl/_url.py", line 163, in __new__
    raise ValueError("Invalid URL: host is required for absolute urls")
ValueError: Invalid URL: host is required for absolute urls
```
```


### Python Version

```console
$ python --version
Python 3.10.9
```

### multidict Version

```console
$ python -m pip show multidict
Name: multidict
Version: 6.0.4
Summary: multidict implementation
Home-page: https://github.com/aio-libs/multidict
Author: Andrew Svetlov
Author-email: andrew.svetlov@gmail.com
License: Apache 2
Location: /home/bkallus/fuzzing/url_differential_fuzzing/url_fuzz_env/lib/python3.10/site-packages
Requires:
Required-by: yarl
```

### yarl Version

```console
$ python -m pip show yarl
Name: yarl
Version: 1.8.2
Summary: Yet another URL library
Home-page: https://github.com/aio-libs/yarl/
Author: Andrew Svetlov
Author-email: andrew.svetlov@gmail.com
License: Apache-2.0
Location: /home/bkallus/fuzzing/url_differential_fuzzing/url_fuzz_env/lib/python3.10/site-packages
Requires: idna, multidict
Required-by:
```

### OS

Arch Linux

### Code of Conduct

- [X] I agree to follow the aio-libs Code of Conduct

webknjaz:
Hi, if you could send a PR with just a test that could then be marked as xfail, it'd be useful.

kenballus:
> Hi, if you could send a PR with just a test that could then be marked as xfail, it'd be useful.

[...]

PR #822, opened 2023-02-06
Add test for empty host #821

kenballus:
Adds a test to ensure that empty hosts are handled in the correct way, as requested in #821.

### DIFF
--- a/tests/test_url_parsing.py
+++ b/tests/test_url_parsing.py
@@ -214,7 +214,7 @@ class TestPort:
         assert u.query_string == ""
         assert u.fragment == ""
 
-    @pytest.mark.xfail
+    @pytest.mark.xfail(reason="https://github.com/aio-libs/yarl/issues/821")
     def test_no_host(self):
         u = URL("//:80")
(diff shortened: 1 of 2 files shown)


tests/test_url_parsing.py
```diff
@@ -239,8 +239,13 @@ def test_no_path(self):
         assert u.fragment == ""
 
     def test_no_host(self):
```
mjpieters, line 241: Can you please mark the test as 'expected to fail'?

```suggestion
    @pytest.mark.xfail
    def test_no_host(self):
```

mjpieters (changes requested):

tests/test_url_parsing.py
```diff
@@ -214,9 +214,15 @@ def test_no_path(self):
         assert u.query_string == ""
         assert u.fragment == ""
 
+    @pytest.mark.xfail
```
webknjaz, line 217:
```suggestion
    @pytest.mark.xfail(reason="https://github.com/aio-libs/yarl/issues/821")
```

PR #848, opened 2023-04-19
test_url_query: Expand tests

mjpieters:
Expanding URL-query tests to cover more behaviors around special chars
when parsing, setting, re-setting and getting values.

Originally created as PR #220.


mjpieters:
Turns out this new PR is not needed, as the branch push has been picked up by the original PR.

PR #849, opened 2023-04-19
Bump the project version to 1.9.0

mjpieters:


mjpieters:
A new release is over-due! Any objections to 1.9.0 being cut in the next few days?

webknjaz (approved):
LGTM!

I wanted to preview the changelog on RTD and it looks like we don't show it in the Sphinx docs. This is not a blocker but would be nice to integrate...

Also, you don't need to skip the change note check — the bot _knows_ to identify the release PRs that remove fragments and update the changelog file.

mjpieters:
The release failed because the Mac wheels builder used an old version of Python 3.7 that doesn't include [this security fix](https://github.com/python/cpython/issues/87133), one that the updated test suite needs to be in place to pass.

I'm out of time tonight to look for a fix for this.

mjpieters:
Right, 3.7.9 is the last version with a MacOS distribution included.

I think it's time to drop 3.7 support :-)

PR #858, opened 2023-04-25
Bump the project version to 1.9.2

mjpieters:


CHANGES.rst
```diff
@@ -14,6 +14,15 @@ Changelog
 
 .. towncrier release notes start
 
+1.8.2 (2023-04-25)
```
webknjaz, line 17: @mjpieters this is why the release failed

mjpieters: Damn, automation fails again. So towncrier looks at the git tags, not the version in `__init__.py`. 

We should either fix towncrier to run `git fetch --tags` first, and / or check for this particular error *before* starting the 80 minute whee building steps. :-D

Even better, put the check into a pre-commit hook..

mjpieters: Correction: it's not from the tags, it _is_ from the project version, but somehow my virtualenv had two yarl versions; 1.8.2 and the editable install with the correct 1.9.2 version. 🤦

PR #859, opened 2023-04-25
Correct changelog

mjpieters:


webknjaz:
Oh, looks like I got there first...

webknjaz:
I've triggered the release too.

PR #891, opened 2023-06-14
Update type hints

mjpieters:
- `explicit_port` and `__bytes__` were missing
- `joinpath()` takes an optional `encoded` argument


mjpieters:
Preferably we detect these issues earlier, but with the current `.pyi` file I am not sure that that's possible. Moving the type hints into `yarl/_url.py` would make it possible to run mypy in strict mode and so catch missing type hints there, but that's a larger undertaking.

PR #893, opened 2023-06-15
📦 Move packaging to PEP 517 in-tree backend

webknjaz:
This essentially allows the cythonization opt-out to be controlled by the `--pure-python` PEP 517 config setting that can be passed to the corresponding build frontends via their respective CLIs.

Yes, the users, and packagers are now supposed to explicitly pass the `--pure-python` config setting to their PEP 517 build frontends in use, when they need to make pure-python wheels.

```console
$ python -m pip install . --config-settings=--pure-python=
$ python -m pip install .

$ python -m pip install -e . --config-settings=--pure-python=
$ python -m pip install -e .

$ python -m pip wheel . --no-deps --config-settings=--pure-python=
$ python -m pip wheel . --no-deps

$ python -m build --config-settings=--pure-python=
$ python -m build

$ python -m build --wheel --config-settings=--pure-python=
$ python -m build --wheel
```

## Related issue number

N/A

## Checklist


pyproject.toml
```diff
@@ -1,8 +1,67 @@
 [build-system]
 requires = [
-  "setuptools>=40",
+  # NOTE: The following build dependencies are necessary for initial
+  # NOTE: provisioning of the in-tree build backend located under
+  # NOTE: `packaging/pep517_backend/`.
+  "expandvars",
+  "setuptools >= 45",
```
webknjaz, line 7: Will need to re-check what's actually the lowest possible setuptools version..

webknjaz (commented):
NOTE: for some reason, pure-python wheels get an `.so` file too. Needs investigation.

webknjaz:
@mjpieters this still needs work, but you're welcome to poke around. Also, I'm not going to fix linting errors right now since the formatters are too invasive and make incorrect changes sometimes.

mjpieters:
> Also, I'm not going to fix linting errors right now since the formatters are too invasive and make incorrect changes sometimes.

That's fine for now. I'd be interested in examples of incorrect changes, if you have any?

webknjaz:
> > Also, I'm not going to fix linting errors right now since the formatters are too invasive and make incorrect changes sometimes.
> 
> That's fine for now. I'd be interested in examples of incorrect changes, if you have any?

@mjpieters like moving `noqa` who-knows-where, so they don't work:

```diff
-from ._backend import (  # noqa: WPS436  # Re-exporting PEP 517 hooks
+from ._backend import (  # Re-exporting PEP 517 hooks
```

webknjaz:
Status update: I made the packaging work. Will have to polish stuff before undrafting the PR. But in general, I'm pretty happy with the interface here.

webknjaz:
So while working on this, I realized that introducing a `--build-c-extensions` inverses the default behavior of building C-extensions, making it an opt-in rather than an opt-out.
With that in mind, I think I'll replace it by `--pure-python` and preserve the opt-out way.

PR #898, opened 2023-06-23
Add new method `without_query_params` to drop query keys

### LINKED ISSUE #774: Ability to remove a query value while retaining others

I have a situation where I would like to get a query parameter, do something with it, and then pass the URL onto a lower layer _without_ that single query param.

The current API is awkward for this.

It would be nice to be able to pass kwargs to `update_query` with `None` value and have that remove those specific query params.

```python
actual = URL('http://test.com/path?a=1&b=2').update_query(a=None)
expected = URL('http://test.com/path?b=2')
assert actual == expected
```

Today this throws a `TypeError`

Today you can accomplish this, but its somewhat awkward:
```python
url = URL('http://test.com/path?a=1&b=2')
query = url.query.copy()
del query['a']
actual = url.with_query(query)
expected = URL('http://test.com/path?b=2')
assert actual == expected
```

### Additional context

_No response_

### Code of Conduct

### LINKED ISSUE #1010: Ability to remove a specific query parameter

I'd like to be able to remove a specific query parameter. Given an URL like `https://domain-name.com/path?param1=value1&param2=value2&param3=value3`, I'd like to be able to derive an URL without the `param2` query parameter, `https://domain-name.com/path?param1=value1&param3=value3` in this case.__

I'd like to be able to do this by providing `None` as the value of the parameter to `update_query`.

```
>>> URL('https://domain-name.com/path?param1=value1&param2=value2&param3=value3').update_query(param2=None)
URL('https://domain-name.com/path?param1=value1&param3=value3')
```

This could be done using a separate method, something like

```
>>> URL('https://domain-name.com/path?param1=value1&param2=value2&param3=value3').remove_query_param('param2')
URL('https://domain-name.com/path?param1=value1&param3=value3')
```

It's possible to do this by grabbing the URLs query dictionary, manipulating it, and replacing the entire query using `with_query`, but providing this behavior would give a cleaner way to express the operation.

### Additional context

_No response_

### Code of Conduct

hongquan:
## What do these changes do?

Add a method, `without_query_params ` to drop some keys from query part.
For example, dropping `"b"` in `"a=1&b=2&c=3"`, we get `"a=1&c=3"`.

No.

## Related issue number

closes #774 
closes #1010

## Checklist


mjpieters:
I really do appreciate that you put in the work for a PR here, but I'd like to first see more discussion about what the API should look like.  Can you please first propose this API change in the issue?

(I'm not 100% convinced that the name `drop_query_keys()` is right here; perhaps `without_query_keys()` is better as you are returning a new URL rather than mutating the existing URL in-place).

hongquan:
Thanks. I think `without_query_keys` is a better name.

dimaqq (approved):

docs/api.rst
```diff
@@ -726,6 +726,12 @@ section generates a new :class:`URL` instance.
       Support subclasses of :class:`int` (except :class:`bool`) and :class:`float`
       as a query parameter value.
 
+.. method:: URL.without_query_keys(*keys)
```
webknjaz, line 729: I think the official term is a "query parameter". Can we use it in the API and texts, please?

CHANGES/898.feature.rst
```diff
@@ -0,0 +1 @@
+Add `without_query_keys` method.
```
webknjaz, line 1: Plz use the past tense and explain what changed, with the end-users as the target audience in mind. This should be clear, informative and accurate. It's possible to write multiple sentences or paragraphs.

Remember to use RST and not Markdown (hence it's double backticks or even Sphinx's RST roles for referencing APIs).

Finally, you can insert a byline as shown in the examples here: https://github.com/aio-libs/yarl/tree/master/CHANGES#adding-change-notes-with-your-prs.

tests/test_url_query.py
```diff
@@ -171,3 +171,22 @@ def test_query_from_empty_update_query(
 
     if "b" in original_url.query:
         assert new_url.query["b"] == original_url.query["b"]
+
+
+@pytest.mark.parametrize(
+    "original_query_string, keys_to_drop, expected_query_string",
```
webknjaz, line 177: nit: it's generally more readable to use the following param declaration:
```suggestion
    ("original_query_string", "keys_to_drop", "expected_query_string"),
```

yarl/_url.py
```diff
@@ -298,7 +299,7 @@ def __bytes__(self):
         return str(self).encode("ascii")
 
     def __eq__(self, other):
-        if not type(other) is URL:
+        if type(other) is not URL:
```
webknjaz, line 302: I think this changes the semantics of the check so I wouldn't include it in this PR.

webknjaz:
@mjpieters @hongquan how would you feel about `URL.without_query_params()`? The RFCs usually use this term...

hongquan:
Agree with `without_query_params` name.

CHANGES/898.feature.rst
```diff
@@ -0,0 +1 @@
+Added ``without_query_params`` method -- by :user:`hongquan`.
```
webknjaz, line 1: Could you make this more detailed? As in, mention where the method is added to, and so on. With end-users being the audience in mind.

yarl/_url.py
```diff
@@ -321,22 +322,22 @@ def __hash__(self):
         return ret
 
     def __le__(self, other):
-        if not type(other) is URL:
+        if type(other) is not URL:
```
webknjaz, line 325: Same as before: if this is desired, it should be a part of a separate PR.

yarl/_url.py
```diff
@@ -1023,6 +1024,16 @@ def update_query(self, *args, **kwargs):
             self._val._replace(query=self._get_str_query(query) or ""), encoded=True
         )
 
+    def without_query_params(self, *keys: str) -> "URL":
```
webknjaz, line 1027: Let's use something like `query_params` since `keys` is ambiguous. There's no concept called “keys” in the HTTP specs.

yarl/_url.py
hongquan: This change is not made by me. It was added by you in a force-push activity:
image

yarl/_url.py
hongquan: This change is not made by me. It was added by you in a force-push activity:

image

yarl/_url.py
```diff
@@ -1023,6 +1024,16 @@ def update_query(self, *args, **kwargs):
...
+        """Remove some keys from query part and return new URL."""
+        valid_keys = set(keys) & set(self.query.keys())
+        if not valid_keys:
+            return self
+        editable_query = cast(MultiDict, self.query.copy())
```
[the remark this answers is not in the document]
hongquan: I have to do `self.query.copy()` because `self.query` is immutable, also `.copy()` returns a shallow copy ([ref](https://multidict.aio-libs.org/en/stable/multidict.html#multidict.MultiDictProxy.copy), it doesn't waste memory much.

The `cast(MultiDict, self.query.copy())` doesn't create new object. It is just for type checking. The returned value is exact the 2nd argument.

hongquan:
The `lint` job in CI/CD failed because it is setup incorrectly. It doesn't use the proposed code, so it breaks when cannot find the symbols that is proposed to add..

yarl/_url.py
webknjaz: @hongquan I didn't touch this branch manually. I only clicked the “Rebase” button on the GitHub UI. This means that the change was already there.
In fact, I specifically remember seeing it earlier. It was added into the very first commit you made: https://github.com/aio-libs/yarl/pull/898/commits/e906337174579cb4d8bac5c1f44ec5ca280a7c3c.

yarl/_url.py
webknjaz: Nope: https://github.com/aio-libs/yarl/pull/898#discussion_r1414784046 / https://github.com/aio-libs/yarl/commit/e906337174579cb4d8bac5c1f44ec5ca280a7c3c.

yarl/_url.py
```diff
@@ -1025,6 +1026,16 @@ def update_query(self, *args, **kwargs):
             self._val._replace(query=self._get_str_query(query) or ""), encoded=True
         )
 
+    def without_query_params(self, *query_params: str) -> "URL":
+        """Remove some keys from query part and return new URL."""
+        params_to_remove = set(query_params) & set(self.query.keys())
+        if not params_to_remove:
+            return URL(self._val._replace(query=self.query_string), encoded=True)
```
webknjaz, line 1033: Why make not just `URL(self)`?

yarl/_url.py
```diff
@@ -1025,6 +1026,16 @@ def update_query(self, *args, **kwargs):
...
+        editable_query = cast(MultiDict, self.query.copy())
+        for k in params_to_remove:
+            editable_query.pop(k, None)
+        return self.with_query(editable_query)
```
webknjaz, line 1037:
```suggestion
        return self.with_query(
            (query_param_name, query_param_value)
            for query_param_name, query_param_value in self.query.items()
            if query_param_name not in params_to_remove
        )
```

yarl/_url.py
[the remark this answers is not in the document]
webknjaz: This should do: https://github.com/aio-libs/yarl/pull/898/files#r1414808748

yarl/_url.py
hongquan: Oh, sorry. I was not aware that I made it by accident.

yarl/_url.py
hongquan: Sorry.

yarl/_url.py
hongquan: Thanks.

hongquan: Tried as you suggested, but `URL(self)` does not create new object, due to [this logic](https://github.com/aio-libs/yarl/blob/0698dc947772786fcff467b12cfb1f3921c7ff31/yarl/_url.py#L169C1-L170)

yarl/_url.py
hongquan: Updated with list comprehension. I use `tuple` as intermediate data, because `with_query()` does not accept `generator` as your code suggested.

hongquan:
@webknjaz I have updated.

webknjaz:
Looks like type stubs need updating..

hongquan:
@webknjaz I updated type stub.

hongquan:
The "Timeline protection" checking job is wrong. It looks for "news" whereas the PR template tells to put to _CHANGES_.

tests/test_url_query.py
```diff
@@ -171,3 +171,23 @@ def test_query_from_empty_update_query(
...
+
+@pytest.mark.parametrize(
+    ("original_query_string", "keys_to_drop", "expected_query_string"),
+    [
+        ("a=10&b=20", ["a"], "b=20"),
+        ("a=10&b=20", ["b"], "a=10"),
+        ("a=10&b=20&c=30", ["b"], "a=10&c=30"),
+        ("a=10&b=20&c=30", ["invalid_key"], "a=10&b=20&c=30"),
+        ("a=10&b=20", ["a", "b"], ""),
+        ("a=10&b=20", [], "a=10&b=20"),
```
bdraco, line 184: This should probably get some tests for encoded query strings to make sure they don't get double encoded

hongquan: @bdraco I updated test cases.

yarl/_url.py
bdraco: Why does it need to be a new object? Since `URL` is immutable we should be fine with returning `self` if there are no changes

yarl/_url.py
```diff
@@ -1112,6 +1112,19 @@ def update_query(self, *args: Any, **kwargs: Any) -> "URL":
             self._val._replace(query=self._get_str_query(query) or ""), encoded=True
         )
 
+    def without_query_params(self, *query_params: str) -> "URL":
+        """Remove some keys from query part and return new URL."""
+        params_to_remove = set(query_params) & set(self.query.keys())
```
bdraco, line 1117:
```suggestion
        params_to_remove = set(query_params) & self.query.keys()
```

Does this need to be wrapped in a set?

tests/test_url_query.py
bdraco: Thanks. Please check for the encoded key case as well. It looks like its only testing values

yarl/_url.py
hongquan: Thanks, I didn't know that `set`'s `&` operator accepts `_KeysView` on the right-hand-side.

yarl/_url.py
```diff
@@ -1112,6 +1112,19 @@ def update_query(self, *args: Any, **kwargs: Any) -> "URL":
             self._val._replace(query=self._get_str_query(query) or ""), encoded=True
         )
 
+    def without_query_params(self, *query_params: str) -> "URL":
+        """Remove some keys from query part and return new URL."""
+        params_to_remove = set(query_params) & self.query.keys()
+        if not params_to_remove:
+            return URL(self._val._replace(query=self.query_string), encoded=True)
```
bdraco, line 1119:
```suggestion
            return self
```

hongquan:
@bdraco I changed the behavior and added test cases for Unicode param keys.

docs/api.rst
```diff
@@ -733,6 +733,12 @@ section generates a new :class:`URL` instance.
       Support subclasses of :class:`int` (except :class:`bool`) and :class:`float`
       as a query parameter value.
 
+.. method:: URL.without_query_params(*keys)
+
+   Return a new URL whose *query* part does not contain specified keys.
+
+   It does nothing if none of specified keys are present in the query.
```
webknjaz, line 740: We should add a `.. versionadded::` directive right before merging.

Here's an example that I'd postpone committing until we know the version this will target.
```suggestion
   It does nothing if none of specified keys are present in the query.

   .. versionadded:: 1.10.0
```

docs/api.rst
```diff
@@ -733,6 +733,12 @@ section generates a new :class:`URL` instance.
       Support subclasses of :class:`int` (except :class:`bool`) and :class:`float`
       as a query parameter value.
 
+.. method:: URL.without_query_params(*keys)
```
webknjaz, line 736: The signature or description should probably explain that the args are of type `str`: https://yarl--898.org.readthedocs.build/en/898/api/#yarl.URL.without_query_params.

webknjaz: Also, the implementation does not call that argument `keys` so the docs shouldn't either.

yarl/_url.py
```diff
@@ -1163,6 +1162,19 @@ def update_query(self, *args: Any, **kwargs: Any) -> "URL":
             self._val._replace(query=self._get_str_query(query) or ""), encoded=True
         )
 
+    def without_query_params(self, *query_params: str) -> "URL":
```
webknjaz, line 1165: @bdraco @mjpieters how would you feel about forcing the args to be positional-only, API-wise?
```suggestion
    def without_query_params(self, *query_params: str, /) -> "URL":
```

yarl/_url.py
```diff
@@ -63,7 +63,6 @@ class CacheInfo(TypedDict):
 
 
 class _SplitResultDict(TypedDict, total=False):
```
webknjaz, line 65: Let's avoid unrelated formatting changes.
```suggestion
class _SplitResultDict(TypedDict, total=False):

```

tests/test_url_query.py
```diff
@@ -171,3 +171,41 @@ def test_query_from_empty_update_query(
...
+@pytest.mark.parametrize(
+    ("original_query_string", "keys_to_drop"),
+    [
+        ("a=10&b=M%C3%B9a+xu%C3%A2n&c=30", ["invalid_key"]),
+        ("a=10&b=M%C3%B9a+xu%C3%A2n", []),
+    ],
+)
+def test_skip_dropping_query_params(
+    original_query_string: str, keys_to_drop: Sequence[str]
+):
```
bdraco, line 208:
```suggestion
) -> None:
```

docs/api.rst
```diff
@@ -733,6 +733,14 @@ section generates a new :class:`URL` instance.
       Support subclasses of :class:`int` (except :class:`bool`) and :class:`float`
       as a query parameter value.
 
+.. method:: URL.without_query_params(*query_params)
+
+   Return a new URL whose *query* part does not contain specified keys.
+
+   It does nothing if none of specified keys are present in the query.
```
bdraco, line 740:
```suggestion
   Return a new URL whose *query* part does not contain specified ``query_params``.

   It does nothing if none of specified ``query_params`` are present in the query.
```

docs/api.rst
```diff
@@ -733,6 +733,14 @@ section generates a new :class:`URL` instance.
       Support subclasses of :class:`int` (except :class:`bool`) and :class:`float`
       as a query parameter value.
 
+.. method:: URL.without_query_params(*query_params)
```
webknjaz, line 736:
```suggestion
.. method:: URL.without_query_params(*query_params, /)
```

docs/api.rst
```diff
@@ -733,6 +733,14 @@ section generates a new :class:`URL` instance.
       Support subclasses of :class:`int` (except :class:`bool`) and :class:`float`
       as a query parameter value.
 
+.. method:: URL.without_query_params(*query_params, /)
+
+   Return a new URL whose *query* part does not contain specified ``query_params``.
+
```
bdraco, line 739:
```suggestion
   Return a new URL whose *query* part does not contain specified ``query_params``.

   Accepts :class:`str` for ``query_params``.

```

yarl/_url.py
bdraco: Oh thats not going to work because its unpacked
```
yarl/_url.py:1166:57: error: / must be ahead of *  [syntax]
```

docs/api.rst
```diff
@@ -733,6 +733,16 @@ section generates a new :class:`URL` instance.
       Support subclasses of :class:`int` (except :class:`bool`) and :class:`float`
       as a query parameter value.
 
+.. method:: URL.without_query_params(*query_params, /)
```
bdraco, line 736:
```suggestion
.. method:: URL.without_query_params(*query_params)
```

yarl/_url.py
```diff
@@ -1163,6 +1163,19 @@ def update_query(self, *args: Any, **kwargs: Any) -> "URL":
             self._val._replace(query=self._get_str_query(query) or ""), encoded=True
         )
 
+    def without_query_params(self, *query_params: str, /) -> "URL":
```
bdraco, line 1166:
```suggestion
    def without_query_params(self, *query_params: str) -> "URL":
```

CHANGES/898.feature.rst
```diff
@@ -0,0 +1 @@
+Added ``without_query_params`` method, to drop some parameters from query string -- by :user:`hongquan`.
```
bdraco, line 1:
```suggestion
Added :meth:`URL.without_query_params() <yarl.URL.without_query_params>` method, to drop some parameters from query string -- by :user:`hongquan`.
```

bdraco (approved):
Thanks @hongquan

PR #902, opened 2023-07-09
 Set up secretless publishing to PyPI

webknjaz:
This also implements sigstore signing of the artifacts.


webknjaz:
The old secrets, tokens, and GitHub Environments have been cleaned up and revoked.

PR #910, opened 2023-07-29
add `/` to protected chars in `_QUERY_REQUOTER`

SafaSafari:
## What do these changes do?

this patch will fix query reqouter skip chars

who users cannot download Instagram posts using aiohttp, after this patch will can =))

## Related issue number

actually no, but I'm wondering why it's not =))

## Checklist


webknjaz (changes requested):
I would be curious to see a quote from the standard that this attempts to comply with, and some explanation of what  doesn't work. With reproducers et al.

If this change is correct indeed, the PR will also need a test covering this specific case. And finally, the linting violations will have to be corrected.

Unfortunately, without understanding of what this patch attempts to fix and what the side effects would be, we can't accept this pull request in its current state.

@SafaSafari could you address the above shortcomings and provide us with all the information necessary to proceed?

SafaSafari:
https://github.com/aio-libs/yarl/assets/29104726/80150edd-0905-47f0-b018-05d3dd06510a

this is what actually occurred and that's my fix for that

also i think linter failed to check because of ` character in CHANGES rst file
@webknjaz

webknjaz:
The video doesn't show the actual URL original, nor can copy it. So it can't be verified.

Besides, the earlier points regarding the low quality and missing bits in the patch still stand.

But if I were to make a wild guess, you're probably hitting reencoding on redirects: https://github.com/aio-libs/aiohttp/issues/5319#issuecomment-1182513707. Using proper arguments (as in, `requote_redirect_url=False`) is the right way to solve this. Also, passing pre-built yarl objects to aiohttp can give more control for the initial input.
This is definitely not a bug. Not in yarl.

webknjaz:
Closing per above.

Issue #880, opened 2023-06-05
basic-auth user and password are converted to lowercase when passed as part of "host" to URL.build


starflows:
### Describe the bug

user and password casing is preserved when using the constructor:

    >>> yarl.URL('httPS://usER:passWORD@hostNAME/paTH')
    URL('https://usER:passWORD@hostname/paTH')
    >>> yarl.URL('httPS://usER:passWORD@hostNAME/paTH').user
    'usER'

user and password casing is lost when using `URL.build`:

    >>> yarl.URL.build(scheme='httPS', host='usER:passWORD@hostNAME', path='/paTH')
    URL('httPS://user:password@hostname/paTH')
    >>> yarl.URL.build(scheme='httPS', host='usER:passWORD@hostNAME', path='/paTH').user
    'user'

Note that the user & password are extracted from the string passed as `host`.

passing user and password explicitly also preserves the casing:

    >>> yarl.URL.build(scheme='httPS', user='usER', password='passWORD', host='hostNAME', path='/paTH')
    URL('httPS://usER:passWORD@hostname/paTH')
    >>> yarl.URL.build(scheme='httPS', user='usER', password='passWORD', host='hostNAME', path='/paTH').user
    'usER'

I'd expect either user and password to preserve casing when passed into `host` (preferably) or them not being extracted into the `user` and `password` attributes and instead remain in the lowercase and encoded `host` attribute.

maybe caused by https://github.com/aio-libs/yarl/issues/386

converting the host to lowercase should be done after extracting username / password.

### To Reproduce

    >>> yarl.URL.build(scheme='httPS', host='usER:passWORD@hostNAME', path='/paTH')
    URL('httPS://user:password@hostname/paTH')
    >>> yarl.URL.build(scheme='httPS', host='usER:passWORD@hostNAME', path='/paTH').user
    'user'

### Expected behavior

    >>> yarl.URL.build(scheme='httPS', host='usER:passWORD@hostNAME', path='/paTH')
    URL('httPS://user:password@hostname/paTH')
    >>> yarl.URL.build(scheme='httPS', host='usER:passWORD@hostNAME', path='/paTH').user
    'usER'

### Logs/tracebacks

```python-traceback
N/A
```

### Python Version

```console
$ python --version
Python 3.10.10
```

### multidict Version

```console
$ python -m pip show multidict
Name: multidict
Version: 6.0.4
Summary: multidict implementation
Home-page: https://github.com/aio-libs/multidict
[...]

PR #920, opened 2023-08-26
Preserve the case of username and password in URL.build

tushar-balwani:
## What do these changes do?
The new changes preserve casing of username and password when extracted from host in URL.build

```python
>>> yarl.URL.build(scheme='httPS', host='usER:passWORD@hostNAME', path='/paTH')
URL('httPS://usER:passWORD@hostname/paTH')
>>> yarl.URL.build(scheme='httPS', host='usER:passWORD@hostNAME', path='/paTH').user
'usER'
>>> yarl.URL.build(scheme='httPS', host='usER:passWORD@hostNAME', path='/paTH').password
'passWORD'
```

No.

## Related issue number

#880 

## Checklist

### DIFF
--- a/CHANGES/880.bugfix.rst
+++ b/CHANGES/880.bugfix.rst
@@ -1,0 +1,1 @@
+Preserve the case of username and password in URL.build when extracted from host.
--- a/tests/test_url_build.py
+++ b/tests/test_url_build.py
@@ -257,3 +257,14 @@ def test_build_with_none_query_string():
 def test_build_with_none_fragment():
     with pytest.raises(TypeError):
         URL.build(scheme="http", host="example.com", fragment=None)
+
+
+def test_build_with_case_sensitive_user_and_password():
+    url = URL.build(scheme='httPS', host='usER:passWORD@hostNAME', path='/paTH')
+    assert url.user == "usER"
+    assert url.password == "passWORD"
+
+
+def test_build_with_case_sensitive_user():
+    url = URL.build(scheme='httPS', host='usER@hostNAME', path='/paTH')
+    assert url.user == "usER"
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -261,6 +261,12 @@ class URL:
         elif not user and not password and not host and not port:
             netloc = ""
         else:
+            if "@" in host and not user and not password:
+                user_pass, host = host.split("@", 1)
+                if ":" in user_pass:
+                    user, password = user_pass.split(":", 1)
+                else:
+                    user = user_pass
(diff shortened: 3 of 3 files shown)


yarl/_url.py
```diff
@@ -261,6 +261,12 @@ def build(
         elif not user and not password and not host and not port:
             netloc = ""
         else:
+            if "@" in host and not user and not password:
```
webknjaz, line 264: This seems like a weird place to extract this data from.. Could you add more detail regarding the problem you're trying to address?

tests/test_url_build.py
```diff
@@ -257,3 +257,14 @@ def test_build_with_none_query_string():
...
+
+def test_build_with_case_sensitive_user_and_password():
+    url = URL.build(scheme='httPS', host='usER:passWORD@hostNAME', path='/paTH')
+    assert url.user == "usER"
+    assert url.password == "passWORD"
+
+
+def test_build_with_case_sensitive_user():
+    url = URL.build(scheme='httPS', host='usER@hostNAME', path='/paTH')
+    assert url.user == "usER"
```
webknjaz, line 270: Looks like these are the same test. So I'd recommend using `@pytest.mark.parametrize` to fold it.

Also, I don't believe that scheme needs to be case-insensitive. Do you have any RFC reference that clarify this corner case?

CHANGES/880.bugfix.rst
```diff
@@ -0,0 +1 @@
+Preserve the case of username and password in URL.build when extracted from host.
```
webknjaz, line 1: Please, use the past tense and RST markup here.

tests/test_url_build.py
```diff
@@ -257,3 +257,14 @@ def test_build_with_none_query_string():
 def test_build_with_none_fragment():
     with pytest.raises(TypeError):
         URL.build(scheme="http", host="example.com", fragment=None)
+
+
+def test_build_with_case_sensitive_user_and_password():
+    url = URL.build(scheme='httPS', host='usER:passWORD@hostNAME', path='/paTH')
```
webknjaz, line 263: Also, why this information would ever be a part of the host? I don't think the API should allow for it at all...

mjpieters:
This is the wrong fix, as discussed in #880. The issue is that `host` should only accept a host string, not an _authority_ string. Pass those to `URL.build(authority=...)` instead.

The right fix is to reject any [`gen-delims` characters](https://www.rfc-editor.org/rfc/rfc3986#section-2.2) in the host value (provided it is not an IP address). Either by raising a deprecation warning first, or just going straight to raising `ValueError`.

PR #959, opened 2023-11-21
Raise `TypeError` when a string is passed for `port` to `URL.build()`

### LINKED ISSUE #883: regression in URL.build(…,port="")
### Describe the bug

Input filtering for with_port #793 changed the logic to render the URL to compare the port to None.
https://github.com/aio-libs/yarl/commit/94b8679da05fd7b2d7bc86ed95e84c09f1946d75#r116844287
as it is possible to create a URL with a port="" via URL.build, this results in urls created with an empty port.

### To Reproduce

```
import yarl
u = yarl.URL.build(scheme="http", host="api.com", port="", path="/v1")
assert str(u) == "http://api.com/v1", str(u)
```

### Expected behavior

Filtering port="" in URL.build.
I'd even prefer to allow "" and convert to None to stay backwards compatible.

### Logs/tracebacks

```python-traceback
AssertionError: http://api.com:/v1
```

### Python Version

```console
$ python --version
Python 3.10.6
```

### multidict Version

```console
Name: multidict
Version: 6.0.2
```

### yarl Version

```console
$ python -m pip show yarl
Name: yarl
Version: 1.9.2
```

### OS

Linux

### Additional context

_No response_

### Code of Conduct

commonism:
## What do these changes do?

Using port="" resulted in malformed urls, now using string for port raises TypeError

URL.build does not accept strings as port any longer.
URL.build does not create invalid urls when using "" as value for port.

## Related issue number

fix #883

## Checklist

### DIFF
--- a/CHANGES/883.bugfix
+++ b/CHANGES/883.bugfix
@@ -1,1 +1,3 @@
-Fix URL.build with port="" creating invalid URLs
+Started treating an empty-string ``port=""`` passed into :py:meth:`~yarl.URL.build` as unset -- :user:`commonism`.
+
+The logic has changed to create URL with no colon and port rendered in its string representation.
--- a/CHANGES/883.bugfix
+++ b/CHANGES/883.bugfix
@@ -1,3 +1,4 @@
-Started treating string ``port`` passed into :py:meth:`~yarl.URL.build` as TypeError -- :user:`commonism`.
+Started raising :py:exc:`TypeError` when a string value is passed into
+:py:meth:`~yarl.URL.build` as the ``port`` argument  -- :user:`commonism`.
 
 The logic has changed to create URL with no colon and port rendered in its string representation.
--- a/CHANGES/883.bugfix
+++ b/CHANGES/883.bugfix
@@ -1,3 +1,3 @@
-Started treating an empty-string ``port=""`` passed into :py:meth:`~yarl.URL.build` as unset -- :user:`commonism`.
+Started treating string ``port`` passed into :py:meth:`~yarl.URL.build` as TypeError -- :user:`commonism`.
 
 The logic has changed to create URL with no colon and port rendered in its string representation.
--- a/CHANGES/883.bugfix
+++ b/CHANGES/883.bugfix
@@ -1,4 +1,4 @@
 Started raising :py:exc:`TypeError` when a string value is passed into
-:py:meth:`~yarl.URL.build` as the ``port`` argument  -- :user:`commonism`.
+:py:meth:`~yarl.URL.build` as the ``port`` argument  -- by :user:`commonism`.
 
 Previously the empty string as port would create malformed URLs when rendered as string representations.
--- a/CHANGES/883.bugfix
+++ b/CHANGES/883.bugfix
@@ -1,4 +1,4 @@
 Started raising :py:exc:`TypeError` when a string value is passed into
 :py:meth:`~yarl.URL.build` as the ``port`` argument  -- :user:`commonism`.
 
-The logic has changed to create URL with no colon and port rendered in its string representation.
+Previously the empty string as port would create malformed URLs when rendered as string representations.
--- a/CHANGES/883.bugfix
+++ b/CHANGES/883.bugfix
@@ -1,0 +1,1 @@
+Fix URL.build with port="" creating invalid URLs
--- a/tests/test_url_build.py
+++ b/tests/test_url_build.py
@@ -33,7 +33,7 @@ def test_build_with_scheme_and_host():
 
 
 @pytest.mark.parametrize(
-    "port,exc,match",
+    ("port", "exc", "match"),
     [
         pytest.param(8000, ValueError, None, id="port-only"),
         pytest.param("", TypeError, "port is required to be int", id="port-str"),
--- a/tests/test_url_build.py
+++ b/tests/test_url_build.py
@@ -39,6 +39,9 @@ def test_build_with_port():
     u = URL.build(scheme="http", host="127.0.0.1", port=8000)
     assert str(u) == "http://127.0.0.1:8000"
 
+    u = URL.build(scheme="http", host="127.0.0.1", port="", path="/v1")
+    assert str(u) == "http://127.0.0.1/v1", str(u)
+
 
 def test_build_with_user():
     u = URL.build(scheme="http", host="127.0.0.1", user="foo")
--- a/tests/test_url_build.py
+++ b/tests/test_url_build.py
@@ -38,10 +38,12 @@ def test_build_with_scheme_and_host():
         pytest.param(
             8000,
             ValueError,
-            r'^Can\'t build URL with "port" but without "host".$',
+            r'^Can\'t build URL with "port" but without "host"\.$',
             id="port-only",
         ),
-        pytest.param("", TypeError, r"^port is required to be int$", id="port-str"),
+        pytest.param(
+            "", TypeError, r"^The port is required to be int\.$", id="port-str"
+        ),
     ],
 )
 def test_build_with_port(port, exc, match):
--- a/tests/test_url_build.py
+++ b/tests/test_url_build.py
@@ -36,7 +36,7 @@ def test_build_with_scheme_and_host():
     ("port", "exc", "match"),
     [
         pytest.param(8000, ValueError, None, id="port-only"),
-        pytest.param("", TypeError, "port is required to be int", id="port-str"),
+        pytest.param("", TypeError, r"^port is required to be int$", id="port-str"),
     ],
 )
 def test_build_with_port(port, exc, match):
--- a/tests/test_url_build.py
+++ b/tests/test_url_build.py
@@ -35,7 +35,12 @@ def test_build_with_scheme_and_host():
 @pytest.mark.parametrize(
     ("port", "exc", "match"),
     [
-        pytest.param(8000, ValueError, None, id="port-only"),
+        pytest.param(
+            8000,
+            ValueError,
+            r'^Can\'t build URL with "port" but without "host".$',
+            id="port-only",
+        ),
         pytest.param("", TypeError, r"^port is required to be int$", id="port-str"),
     ],
 )
--- a/tests/test_url_build.py
+++ b/tests/test_url_build.py
@@ -33,14 +33,14 @@ def test_build_with_scheme_and_host():
 
 
 @pytest.mark.parametrize(
-    "port,exc",
+    "port,exc,match",
     [
-        pytest.param(8000, ValueError, id="port-only"),
-        pytest.param("", TypeError, id="port-str"),
+        pytest.param(8000, ValueError, None, id="port-only"),
+        pytest.param("", TypeError, "port is required to be int", id="port-str"),
     ],
 )
-def test_build_with_port(port, exc):
-    with pytest.raises(exc):
+def test_build_with_port(port, exc, match):
+    with pytest.raises(exc, match=match):
         URL.build(port=port)
 
 
--- a/tests/test_url_build.py
+++ b/tests/test_url_build.py
@@ -32,15 +32,16 @@ def test_build_with_scheme_and_host():
     assert u == URL("http://127.0.0.1")
 
 
-def test_build_with_port():
-    with pytest.raises(ValueError):
-        URL.build(port=8000)
-
-    u = URL.build(scheme="http", host="127.0.0.1", port=8000)
-    assert str(u) == "http://127.0.0.1:8000"
-
-    u = URL.build(scheme="http", host="127.0.0.1", port="", path="/v1")
-    assert str(u) == "http://127.0.0.1/v1", str(u)
+@pytest.mark.parametrize(
+    "port,exc",
+    [
+        pytest.param(8000, ValueError, id="port-only"),
+        pytest.param("", TypeError, id="port-str"),
+    ],
+)
+def test_build_with_port(port, exc):
+    with pytest.raises(exc):
+        URL.build(port=port)
 
 
 def test_build_with_user():
--- a/tests/test_url_build.py
+++ b/tests/test_url_build.py
@@ -40,7 +40,7 @@ def test_build_with_scheme_and_host():
             ValueError,
             r"""(?x)
             ^
-            Can't build URL with "port" but without "host"\.
+            Can't\ build\ URL\ with\ "port"\ but\ without\ "host"\.
             $
             """,
             id="port-only",
@@ -51,6 +51,7 @@ def test_build_with_scheme_and_host():
     ],
 )
 def test_build_with_port(port, exc, match):
+    print(match)
     with pytest.raises(exc, match=match):
         URL.build(port=port)
 
--- a/tests/test_url_build.py
+++ b/tests/test_url_build.py
@@ -38,7 +38,11 @@ def test_build_with_scheme_and_host():
         pytest.param(
             8000,
             ValueError,
-            r'^Can\'t build URL with "port" but without "host"\.$',
+            r"""(?x)
+            ^
+            Can't build URL with "port" but without "host"\.
+            $
+            """,
             id="port-only",
         ),
         pytest.param(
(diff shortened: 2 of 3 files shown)


commonism:
I did not include this in the PR - but the requirements file has a typo.
```
diff --git a/requirements/dev.txt b/requirements/dev.txt
index 62fe39a..3b676e4 100644
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -1,2 +1,2 @@
--r text.txt
+-r test.txt
 towncrier==23.11.0
 ```

**Comment by webknjaz** [comment]
> I did not include this in the PR - but the requirements file has a typo.
> 
> ```
> diff --git a/requirements/dev.txt b/requirements/dev.txt
> index 62fe39a..3b676e4 100644
> --- a/requirements/dev.txt
> +++ b/requirements/dev.txt
> @@ -1,2 +1,2 @@
> --r text.txt
> +-r test.txt
>  towncrier==23.11.0
> ```

Fixed that in https://github.com/aio-libs/yarl/commit/85b53c9f01e0f580405cc160e9f1ab168367b94e. Thanks!

File: yarl/_url.py [hunk 1]
```diff
@@ -233,6 +233,8 @@ def build(
             raise ValueError(
                 'Can\'t mix "authority" with "user", "password", "host" or "port".'
             )
+        if isinstance(port, str) and port == "":
```
webknjaz, line 236: How about
```suggestion
        if isinstance(port, str) and not port:
```

webknjaz: Or even better:
```suggestion
        if port == "":
```
(you don't really need to check the type, since comparing to a string literal would be falsy if it's not)

tests/test_url_build.py
```diff
@@ -39,6 +39,9 @@ def test_build_with_port():
     u = URL.build(scheme="http", host="127.0.0.1", port=8000)
     assert str(u) == "http://127.0.0.1:8000"
 
+    u = URL.build(scheme="http", host="127.0.0.1", port="", path="/v1")
+    assert str(u) == "http://127.0.0.1/v1", str(u)
```
webknjaz, line 43: This and the previous assert are actually the same test. Instead of copying the lines, `@pytest.mark.parametrize` should be used. That previous one shouldn't have been put into the same test as the input validation, anyway.

Could you update this, please?

CHANGES/883.bugfix
```diff
@@ -0,0 +1 @@
+Fix URL.build with port="" creating invalid URLs
```
webknjaz, line 1: This should use RST markup for code, [past tense](https://yarl.aio-libs.org/en/latest/contributing/guidelines/#adding-change-notes-with-your-prs) and a byline with the `:user:` role in it:
```suggestion
Started treating an empty-string ``port=""`` passed into :py:meth:`~yarl.URL.build` as unset -- :user:`commonism`.

The logic has changed to create URL with no colon and port rendered in its string representation.
```

webknjaz (commented):
Actually… I'm not sure if it's a good idea to accept empty strings at all. Should we error out instead, and aim for stricter public API input validation in general?

cc @Dreamsorcerer @mjpieters

Dreamsorcerer:
> Actually… I'm not sure if it's a good idea to accept empty strings at all. Should we error out instead, and aim for stricter public API input validation in general?

Yeah, it's typed as int. I'm not sure we should add code that explicitly supports something the type checker rejects makes sense (and support `str` in general would be odd).
https://github.com/aio-libs/yarl/blob/85b53c9f01e0f580405cc160e9f1ab168367b94e/yarl/__init__.pyi#L61

tests/test_url_build.py
commonism: I'll wait for the decision if str is actually acceptable as argument.

webknjaz: I think we agreed that it's not acceptable: https://github.com/aio-libs/yarl/pull/959#issuecomment-1821506912.

So that `raises()` check is what actually needs to be parametrized.

yarl/_url.py
```diff
@@ -233,6 +233,8 @@ def build(
             raise ValueError(
                 'Can\'t mix "authority" with "user", "password", "host" or "port".'
             )
+        if port is not None and not isinstance(port, int):
```
webknjaz, line 236: Can we collapse this?
```suggestion
        if not isinstance(port, (int, NoneType)):
```

tests/test_url_build.py
```diff
@@ -32,12 +32,16 @@ def test_build_with_scheme_and_host():
...
-    assert str(u) == "http://127.0.0.1:8000"
+@pytest.mark.parametrize(
+    "port,exc",
+    [
+        pytest.param(8000, ValueError, id="port-only"),
+        pytest.param("", TypeError, id="port-str"),
+    ],
+)
+def test_build_with_port(port, exc):
+    with pytest.raises(exc):
```
webknjaz, line 43: It is strongly recommended to always use the `match=` regexp with `pytest.raises()`, especially when testing against stdlib exceptions which may come from a lot of places, silently making the test false-positive over time.

Could you integrate this?

tests/test_url_build.py
```diff
@@ -32,12 +32,16 @@ def test_build_with_scheme_and_host():
...
 
 
-def test_build_with_port():
-    with pytest.raises(ValueError):
-        URL.build(port=8000)
-
-    u = URL.build(scheme="http", host="127.0.0.1", port=8000)
-    assert str(u) == "http://127.0.0.1:8000"
+@pytest.mark.parametrize(
+    "port,exc",
```
webknjaz, line 36: A non-string sequence for the params list is considered more readable, some projects even enforce this with linters:
```suggestion
    ("port", "exc"),
```

yarl/_url.py
commonism: NameError: name 'NoneType' is not defined

I chose None.__class__ instead.

tests/test_url_build.py
commonism: done

CHANGES/883.bugfix
```diff
@@ -0,0 +1,3 @@
+Started treating string ``port`` passed into :py:meth:`~yarl.URL.build` as TypeError -- :user:`commonism`.
```
webknjaz, line 1:
```suggestion
Started raising :py:exc:`TypeError` when a string value is passed into
:py:meth:`~yarl.URL.build` as the ``port`` argument  -- :user:`commonism`.
```

CHANGES/883.bugfix
```diff
@@ -0,0 +1,3 @@
...
+
+The logic has changed to create URL with no colon and port rendered in its string representation.
```
webknjaz, line 3: I suppose this isn't really true anymore?

tests/test_url_build.py
```diff
@@ -32,12 +32,16 @@ def test_build_with_scheme_and_host():
...
-    with pytest.raises(ValueError):
-        URL.build(port=8000)
-
-    u = URL.build(scheme="http", host="127.0.0.1", port=8000)
-    assert str(u) == "http://127.0.0.1:8000"
+@pytest.mark.parametrize(
+    ("port", "exc", "match"),
+    [
+        pytest.param(8000, ValueError, None, id="port-only"),
+        pytest.param("", TypeError, "port is required to be int", id="port-str"),
```
webknjaz, line 39: Let's use raw-strings for regexes:
```suggestion
        pytest.param("", TypeError, r"^port is required to be int$", id="port-str"),
```

tests/test_url_build.py
```diff
@@ -32,12 +32,16 @@ def test_build_with_scheme_and_host():
...
-def test_build_with_port():
-    with pytest.raises(ValueError):
-        URL.build(port=8000)
-
-    u = URL.build(scheme="http", host="127.0.0.1", port=8000)
-    assert str(u) == "http://127.0.0.1:8000"
+@pytest.mark.parametrize(
+    ("port", "exc", "match"),
+    [
+        pytest.param(8000, ValueError, None, id="port-only"),
```
webknjaz, line 38: Could you replace `None` with something meaningful?

yarl/_url.py
```diff
@@ -233,6 +233,8 @@ def build(
             raise ValueError(
                 'Can\'t mix "authority" with "user", "password", "host" or "port".'
             )
+        if not isinstance(port, (int, None.__class__)):
```
webknjaz, line 236: A more idiomatic way would be
```suggestion
        if not isinstance(port, (int, type(None))):
```

webknjaz:
@commonism looks like the PR title and description no longer correspond to its contents. This needs to be corrected.

tests/test_url_build.py
```diff
@@ -32,12 +32,21 @@ def test_build_with_scheme_and_host():
...
-
-    u = URL.build(scheme="http", host="127.0.0.1", port=8000)
-    assert str(u) == "http://127.0.0.1:8000"
+@pytest.mark.parametrize(
+    ("port", "exc", "match"),
+    [
+        pytest.param(
+            8000,
+            ValueError,
+            r'^Can\'t build URL with "port" but without "host".$',
```
webknjaz, line 41: I imagine escaping `'` within a raw-string would produce `\\'`. Not sure what it'd do to the regexp…
In such cases, I tend to split the literal as follows:
```suggestion
            r"^Can't " r'build URL with "port" but without "host".$',
```

webknjaz: Another (better?) option is to use a multiline raw-string combined with a `re.VERBOSE` flag:
```suggestion
            r"""(?x)
            ^
            Can't build URL with "port" but without "host"\.
            $
            """,
```

tests/test_url_build.py
```diff
@@ -32,12 +32,21 @@ def test_build_with_scheme_and_host():
...
+            id="port-only",
+        ),
+        pytest.param("", TypeError, r"^port is required to be int$", id="port-str"),
```
webknjaz, line 44: Looking at the other exception messages, I realized that this one doesn't match the common style. Let's change it to look similar. For example, all the other messages are full sentences, starting with a capital letter and ending with a period.

CHANGES/883.bugfix
webknjaz: Looks like you marked this as resolved by accident, without actually updating anything...

tests/test_url_build.py
commonism: As 0-3 \ infront of a ' are allowed in a re to match a '
```
>>> re.match("^test't$", "test't")
<re.Match object; span=(0, 6), match="test't">
>>> re.match("^test\'t$", "test't")
<re.Match object; span=(0, 6), match="test't">
>>> re.match("^test\\'t$", "test't")
<re.Match object; span=(0, 6), match="test't">
>>> re.match("^test\\\'t$", "test't")
<re.Match object; span=(0, 6), match="test't">
>>> re.match("^test\\\\'t$", "test't")
>>>
```
raw strings end up with 2 \, and any of the proposed changes to the re do not improve read-ability, I'd go with it?

tests/test_url_build.py
commonism: The common pattern does not end with a period.
```
# Ends with period
grep raise yarl/ -R | grep -E "\..\)" | wc -l
# 2
```

```
# Ends with anything but period
grep raise yarl/ -R | grep -E "[^\.].\)$" | wc -l
58
```

tests/test_url_build.py
webknjaz: I disagree — the multiline raw-string is more readable, because in verbose mode, it allows moving the line start/end markers to separate lines. Moving complexity to other lines has a positive cognitive effect. I'd also argue that any escaped things in a string (not even a regexp) do reduce the readability.

CHANGES/883.bugfix
webknjaz, line 1: I checked that this change note is rendered well: https://yarl--959.org.readthedocs.build/en/959/changes/#bug-fixes.

commonism: the missing by?
added.

CHANGES/883.bugfix
commonism: TypeError was an update.

tests/test_url_build.py
commonism: okay than

webknjaz (approved):

tests/test_url_build.py
```diff
@@ -32,12 +32,27 @@ def test_build_with_scheme_and_host():
...
+    ("port", "exc", "match"),
+    [
+        pytest.param(
+            8000,
+            ValueError,
+            r"""(?x)
+            ^
+            Can't build URL with "port" but without "host"\.
+            $
+            """,
```
webknjaz, line 45: So this didn't match for some reason. Need to check why.
```python-traceback
  _______________________ test_build_with_port[port-only] ________________________
  [gw2] linux -- Python 3.7.17 /tmp/tmp.WTVCAbf0yq/venv/bin/python
  
  port = 8000, exc = <class 'ValueError'>
  match = '(?x)\n            ^\n            Can\'t build URL with "port" but without "host"\\.\n            $\n            '
  
      @pytest.mark.parametrize(
          ("port", "exc", "match"),
          [
              pytest.param(
                  8000,
                  ValueError,
                  r"""(?x)
                  ^
                  Can't build URL with "port" but without "host"\.
                  $
                  """,
                  id="port-only",
              ),
              pytest.param(
                  "", TypeError, r"^The port is required to be int\.$", id="port-str"
              ),
          ],
      )
      def test_build_with_port(port, exc, match):
          with pytest.raises(exc, match=match):
  >           URL.build(port=port)
  
  exc        = <class 'ValueError'>
  match      = '(?x)\n            ^\n            Can\'t build URL with "port" but without "host"\\.\n            $\n            '
  port       = 8000
  
  /project/tests/test_url_build.py:55: 
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
  
  cls = <class 'yarl.URL'>
  
      @classmethod
      def build(
          cls,
          *,
          scheme="",
          authority="",
          user=None,
          ***
          host="",
          port=None,
          path="",
          query=None,
          query_string="",
          fragment="",
          encoded=False,
      ):
          """Creates and returns a new URL"""
      
          if authority and (user or password or host or port):
              raise ValueError(
                  'Can\'t mix "authority" with "user", "password", "host" or "port".'
              )
          if not isinstance(port, (int, type(None))):
              raise TypeError("The port is required to be int.")
          if port and not host:
  >           raise ValueError('Can\'t build URL with "port" but without "host".')
  E           ValueError: Can't build URL with "port" but without "host".
  
  authority  = ''
  cls        = <class 'yarl.URL'>
  encoded    = False
  fragment   = ''
  host       = ''
  password   = None
  path       = ''
  port       = 8000
  query      = None
  query_string = ''
  scheme     = ''
  user       = None
  
  ../venv/lib/python3.7/site-packages/yarl/_url.py:239: ValueError
  
  During handling of the above exception, another exception occurred:
  
  port = 8000, exc = <class 'ValueError'>
  match = '(?x)\n            ^\n            Can\'t build URL with "port" but without "host"\\.\n            $\n            '
  
      @pytest.mark.parametrize(
          ("port", "exc", "match"),
          [
              pytest.param(
                  8000,
                  ValueError,
                  r"""(?x)
                  ^
                  Can't build URL with "port" but without "host"\.
                  $
                  """,
                  id="port-only",
              ),
              pytest.param(
                  "", TypeError, r"^The port is required to be int\.$", id="port-str"
              ),
          ],
      )
      def test_build_with_port(port, exc, match):
          with pytest.raises(exc, match=match):
  >           URL.build(port=port)
  E           AssertionError: Regex pattern did not match.
  E            Regex: '(?x)\n            ^\n            Can\'t build URL with "port" but without "host"\\.\n            $\n            '
  E            Input: 'Can\'t build URL with "port" but without "host".'
  
  exc        = <class 'ValueError'>
  match      = '(?x)\n            ^\n            Can\'t build URL with "port" but without "host"\\.\n            $\n            '
  port       = 8000
  
  /project/tests/test_url_build.py:55: AssertionError
```
```console
  FAILED ../../../project/tests/test_url_build.py::test_build_with_port[port-only] - AssertionError: Regex pattern did not match.
```

commonism: > Whitespace within the pattern is ignored, except when in a character class, or when preceded by an unescaped backslash,

webknjaz: Ah.. Thanks!

PR #967, opened 2023-11-27
🧪📦 Pass sdist to `cibuildwheel` @ GHA

webknjaz:
$sbj.


webknjaz:
This is blocked by https://github.com/pypa/cibuildwheel/issues/1683. But it's also a can of worms, so I'll close the PR, postponing the effort until there's a more elegant UX.

PR #991, opened 2024-02-07
Test lone surrogate handling in encoded and parsed URL paths

pajod:
## What do these changes do?

Adds regression tests for lone surrogates produced by decoding invalid
UTF-8 with ``surrogateescape``. Building a URL with ``encoded=True``
preserves them verbatim, while parsing the same path percent-encodes the
valid text and drops the surrogates, identically on the C and
pure-Python quoting backends.

When this PR was opened it demonstrated a divergence between the two
backends: the compiled quoter kept lone surrogates that the pure-Python
quoter dropped. That gap was closed by #1799, #1751 and #1752, so the
tests now assert the shared behaviour and pass both with the compiled
extension and with `YARL_NO_EXTENSIONS=1`.

No, tests only.

No.

## Related issue number

Related to https://github.com/aio-libs/aiohttp/pull/8088 and to #1799,
#1751 and #1752, which resolved the inconsistency these tests originally
demonstrated.

## Checklist

- [x] Documentation reflects the changes (N/A; tests only, no public API change)

### DIFF
--- a/tests/test_url_parsing.py
+++ b/tests/test_url_parsing.py
@@ -580,6 +580,47 @@ class TestStripEmptyParts:
         assert u.query_string == ""
         assert u.fragment == ""
 
+    def test_truncated_utf_sequence(self) -> None:
+        # A truncated UTF-8 sequence decoded with surrogateescape yields lone
+        # surrogates. With encoded=True the caller takes responsibility for
+        # the value, so they are preserved verbatim; parsing the same path
+        # normally runs the quoter, which drops them, identically on the C
+        # and pure-Python backends.
+        raw_path = b"/P\xc3\xbcnktchen\xa0\xef\xb7#"
+        abspath, _hash_separator, frag = raw_path.decode(
+            "utf-8", "surrogateescape"
+        ).partition("#")
+        u = URL.build(path=abspath, query_string="", fragment=frag, encoded=True)
+        u2 = URL("/P%C3%BCnktchen\udca0\udcef\udcb7")
+        assert u.scheme == ""
+        assert u.user is None
+        assert u.password is None
+        assert u.path == "/Pünktchen\udca0\udcef\udcb7"
+        assert u.query_string == ""
+        assert u.fragment == ""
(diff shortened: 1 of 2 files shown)


tests/test_url_parsing.py
```diff
@@ -584,3 +584,34 @@ def test_empty_path(self):
...
+        u = URL.build(path=abspath, query_string="", fragment=frag, encoded=True)
+        u2 = URL("/P%C3%BCnktchen\udca0\udcef\udcb7#Pünktelchen\udca0\udcef\udcb6")
+        assert u.scheme == ""
+        assert u.user is None
+        assert u.password is None
+        assert u.path == "/Pünktchen\udca0\udcef\udcb7"
+        assert u.query_string == ""
+        assert u.fragment == "Pünktelchen\udca0\udcef\udcb6"
+        assert u.path == u2.path
+        # assert u.fragment == u2.fragment
```
webknjaz, line 617:
```suggestion
```

tests/test_url_parsing.py
```diff
@@ -584,3 +584,34 @@ def test_empty_path(self):
         assert u.path == ""
         assert u.query_string == ""
         assert u.fragment == ""
+
+    def test_truncated_utf_sequence(self):
```
webknjaz, line 588: If path is the only difference, could you parametrize the test?

bdraco (approved):

tests/test_url_parsing.py
aiolibsbot: Tests differ beyond path. Frag case adds `raw_fragment`/`fragment` asserts. Parametrize on `(raw_path, expected_path, expected_fragment)`, empty fragment for the first case. Folds cleanly into one test.

tests/test_url_parsing.py
aiolibsbot: Addressed. Two tests folded into one parametrized case. Params: raw_path, parsed_url, expected_raw_fragment, expected_fragment. Path-only case uses empty fragment.

bdraco:
Thanks @pajod

PR #1016, opened 2024-06-06
Drop Python 3.7

Dreamsorcerer:


Dreamsorcerer:
@webknjaz I think the timeline protection is misconfigured here.

webknjaz:
@Dreamsorcerer plz ignore it for now. I've been meaning to fix the bot to take into account standalone configs.

Dreamsorcerer:
> @Dreamsorcerer plz ignore it for now. I've been meaning to fix the bot to take into account standalone configs.

Well, it's currently marked as required with admin override disabled...
Will break Dependabot while it's required either way.

webknjaz:
If you apply the skip label, it should shut up and let you merge..

Dreamsorcerer:
Codecov isn't getting reported either, though it appears to have upload successfully...

.pre-commit-config.yaml
```diff
@@ -164,8 +164,8 @@ repos:
     - --html-report=.tox/.tmp/.mypy/python-3.8
     pass_filenames: false
   - id: mypy
-    alias: mypy-py37
-    name: MyPy, for Python 3.7
+    alias: mypy-py38
```
webknjaz, line 167: This is a duplicate, just drop it.

CHANGES/1016.removal.rst
```diff
@@ -0,0 +1 @@
+Drop support for Python 3.7 -- by :user:`Dreamsorcerer`.
```
webknjaz, line 1:
```suggestion
Dropped support for Python 3.7 -- by :user:`Dreamsorcerer`.
```

.pre-commit-config.yaml
webknjaz: Also, check if it's already possible to run it against 3.13 and add it to the top. In general, it's good to run against every other version where possible. This would need an update in the reusable linters workflow, if added.

webknjaz:
> Codecov isn't getting reported either, though it appears to have uploaded successfully...

Codecov is known to be flaky. It might take a while for it to process uploads. Also, we don't fail on upload errors, I think.

webknjaz:
@Dreamsorcerer oh, wait. You reduced the number of uploads. Now you have to adjust the expectation in the Codecov config file too. It's waiting for more jobs to complete before posting the status.

.pre-commit-config.yaml
Dreamsorcerer: Ah, didn't realise it was running under multiple versions. Seems a bit overkill for Mypy..

webknjaz: This helps cover interpreter-conditional branches. It's cheap enough.

.codecov.yml
```diff
@@ -2,7 +2,7 @@
 
 codecov:
   notify:
-    after_n_builds: 27  # The number of test matrix+lint jobs uploading coverage
+    after_n_builds: 20  # The number of test matrix+lint jobs uploading coverage
```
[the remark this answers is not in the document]
webknjaz: Or 23..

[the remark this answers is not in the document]
webknjaz: Fixed in https://github.com/aio-libs/yarl/commit/43503cc8f068be45aba71a3f72cf237fad57a05b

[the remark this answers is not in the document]
Dreamsorcerer: Setting fail_ci_on_error would also alert to any failed uploads immediately, without needing to try and keep this value in check. We originally set this value just to avoid getting email notifications from codecov before it had full coverage. Setting it too high just complicates maintenance, as observed when it took me a couple of hours before realising why codecov wasn't working.

Ideally, codecov would allow us some way to have the notification sent as soon as the test matrix completes.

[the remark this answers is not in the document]
webknjaz: @Dreamsorcerer Yes, that option also needs to be set. But this specific setting is the way to allow notifications upon matrix completion, as it'd see all the uploads. If it's set lower, Codecov starts polling GH API too early, which consumes more rate limit and contributes to flakiness. Hence, the need for this to match for best stability possible.

PR #1018, opened 2024-06-06
Increase timeout on linter

Dreamsorcerer:


webknjaz:
Why is so huge change needed? I was trying to keep it as tight as possible..

webknjaz:
I only see one job exceeding it by a few seconds so an increment by one is enough. Otherwise, it opens up possibilities for a lot of CPU wasted due to network problems within GitHub itself. It's best to interrupt the workflow runs early. Only if the time increase is permanently observed, should the timeout be raised.

Dreamsorcerer:
15 mins doesn't seem much and aligns with our other projects. Today I've seen atleast 3 cancellations before it even finished installing dependencies. So, I think variability in setup time is too much to make this timeout only 3-5 mins.

webknjaz:
Yeah, I haven't yet updated other workflows to have shorter timeouts, but that's the plan. Looks like 4-5 min could work. We want to bail early for responsiveness.

PR #1028, opened 2024-06-26
Fix setuptools typing errors

Dreamsorcerer:


pytest.ini
```diff
@@ -23,8 +23,8 @@ addopts =
   -p pytest_cov
 
   # `pytest-cov`:
-  --cov
-  --cov-context=test
+  --cov=yarl/
+  --cov=tests/
```
webknjaz, line 27: No, this change is incorrect. The config must be in `.coveragerc` to enable it to be reused outside pytest.

.coveragerc
```diff
@@ -12,10 +12,4 @@ show_missing = true
 
 [run]
 branch = true
-cover_pylib = false
 plugins = Cython.Coverage
-relative_files = true
-source =
-  tests
-source_pkgs =
-  yarl
```
webknjaz, line 21: Keep this as is, it's configured intentionally. Are there any settings you didn't like?

webknjaz (changes requested):
The config changes are incorrect. The typing might be fine, I didn't look into it. Note that the in-tree build-backend changes will need to be copied over into other projects, but I haven't yet copied the previous bugfix commit.

.coveragerc
```diff
@@ -12,10 +12,4 @@ show_missing = true
 
 [run]
 branch = true
-cover_pylib = false
 plugins = Cython.Coverage
-relative_files = true
```
webknjaz, line 17: Hm.. Perhaps this relative files setting could be a workaround for codecov not understanding paths… Could be an adequate repro for a bug report.

commonism:
Can you merge the typing glue to fix the mypy/linter ci failure?

webknjaz:
> Can you merge the typing glue to fix the mypy/linter ci failure?

Once it's re-submitted as a separate PR, I think so.

.coveragerc
Dreamsorcerer: Blindly copying from a repo which still works. I'll just merge the typing fixes and leave it to you.

PR #1029, opened 2024-06-27
🚑🧪 Store absolute paths in coverage database

webknjaz:
This will hopefully act as a workaround for the codecov-action v4 regression that stopped understanding relative paths.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -14,7 +14,7 @@ show_missing = true
 branch = true
 cover_pylib = false
 plugins = Cython.Coverage
-relative_files = true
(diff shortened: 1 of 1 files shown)


webknjaz:
Interesting.. there's `<package name="." line-rate="0.9961" branch-rate="0.9712" complexity="0">` in uploads that can't be processed. Also, https://github.com/ansible/pylibssh/blob/3888862/.coveragerc has relative files and it works just fine. I'll have to dig into this separately.

webknjaz:
This has been fixed by #1074.

PR #1031, opened 2024-07-02
Remove unused type ignores and casts

hauntsaninja:

No

## Checklist


.pre-commit-config.yaml
```diff
@@ -121,6 +121,8 @@ repos:
     args:
     - --install-types
     - --non-interactive
+    - --warn-unused-ignores
```
webknjaz, line 124: Please, move these to a common config instead of duplicating. The existing args are only here because they differ across invocations.

CHANGES/1031.contrib.rst
```diff
@@ -0,0 +1 @@
+Update mypy, remove unused ignores and casts
```
webknjaz, line 1: Don't use imperative mood. Change notes have different style. Optionally, add a byline. Add a trailing period to the sentence. Use proper capitalization for MyPy.

webknjaz (commented):
Thanks! There's a few things to address to make this mergeable.

webknjaz:
Looks like the 3.10 and 3.8 runs also need `types-Pygments` in their additional deps.

CHANGES/1031.contrib.rst
bdraco, line 1:
```suggestion
Updated mypy and removed unused ignores and casts -- by :user:`hauntsaninja`.
```

CHANGES/1031.contrib.rst
```diff
@@ -0,0 +1 @@
+Updated mypy and removed unused ignores and casts -- by :user:`hauntsaninja`.
```
bdraco, line 1:
```suggestion
Updated MyPy and removed unused ignores and casts -- by :user:`hauntsaninja`.
```

bdraco:
skip label added as a workaround since the bot is currently not working as expected

CHANGES/1031.contrib.rst
```diff
@@ -0,0 +1 @@
+Updated MyPy and removed unused ignores and casts -- by :user:`hauntsaninja`.
```
bdraco, line 1:
```suggestion
Removed unused type ignores and casts -- by :user:`hauntsaninja`.
```

bdraco:
mypy already got updated so I removed that

bdraco (approved):
Thanks @hauntsaninja

PR #1033, opened 2024-07-03
Normalize URL string ports per RFC 3986 section 3.2.3

commonism:
## What do these changes do?

The changes add a missing step in rfc3986 normalization - strip default port from URL str representation.

Yes

## Related issue number

https://github.com/aio-libs/yarl/pull/1023#discussion_r1636861213
> In the future, with another PR, I think the only normalisation step currently missing is https://datatracker.ietf.org/doc/html/rfc3986.html#section-6.2.3
> Which would include (for HTTP/HTTPS schemes) setting the path to / when empty and removing the port number when matching the default 80/443.

## Checklist

### DIFF
--- a/CHANGES/1033.bugfix.rst
+++ b/CHANGES/1033.bugfix.rst
@@ -1,4 +1,4 @@
 Normalize default ports according to :rfc:`3986`.
 
-Specified ports are removed from the str representation of an URL if the port matches
-the schemes default port -- by :user:`commonism`.
+Specified ports are removed from the :class:`str` representation of a :class:`~yarl.URL`
+if the port matches the scheme's default port -- by :user:`commonism`.
--- a/CHANGES/1033.bugfix.rst
+++ b/CHANGES/1033.bugfix.rst
@@ -1,4 +1,6 @@
-Normalize default ports according to :rfc:`3986`.
+The default protocol ports of well-known URI schemes are now taken into account
+during the normalization of the URL string representation in accordance with
+:rfc:`3986#section-3.2.3`.
 
 Specified ports are removed from the :class:`str` representation of a :class:`~yarl.URL`
 if the port matches the scheme's default port -- by :user:`commonism`.
--- a/CHANGES/1033.bugfix.rst
+++ b/CHANGES/1033.bugfix.rst
@@ -1,4 +1,4 @@
 Normalize default ports according to :rfc:`3986`.
 
 Specified ports are removed from the str representation of an URL if the port matches
-the schemes default port. -- by :user:`commonism`.
+the schemes default port -- by :user:`commonism`.
--- a/CHANGES/1033.bugfix.rst
+++ b/CHANGES/1033.bugfix.rst
@@ -1,0 +1,4 @@
+Normalize default ports according to :rfc:`3986`.
+
+Specified ports are removed from the str representation of an URL if the port matches
+the schemes default port. -- by :user:`commonism`.
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -219,6 +219,12 @@ def test_authority_full_nonasci() -> None:
     assert url.authority == "степан:пароль@слава.укр:8080"
 
 
+def test_authority_unknown_scheme():
+    v = "scheme://user:password@example.com:43/path/to?a=1&b=2"
+    url = URL(v)
+    assert str(url) == v
+
+
 def test_lowercase():
     url = URL("http://gitHUB.com")
     assert url.raw_host == "github.com"
@@ -1332,6 +1338,7 @@ def test_is_default_port_for_absolute_url_without_port():
 def test_is_default_port_for_absolute_url_with_default_port():
     url = URL("http://example.com:80")
     assert url.is_default_port()
+    assert str(url) == "http://example.com"
 
 
 def test_is_default_port_for_absolute_url_with_nondefault_port():
@@ -1542,7 +1549,7 @@ def test_parent_for_empty_url():
 
 def test_empty_value_for_query():
     url = URL("http://example.com/path").with_query({"a": ""})
-    assert str(url) == "http://example.com/path?a="
+    assert str(url) == "http://example.com/path?a"
 
 
 def test_none_value_for_query():
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -1549,7 +1549,7 @@ def test_parent_for_empty_url():
 
 def test_empty_value_for_query():
     url = URL("http://example.com/path").with_query({"a": ""})
-    assert str(url) == "http://example.com/path?a"
+    assert str(url) == "http://example.com/path?a="
 
 
 def test_none_value_for_query():
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -219,7 +219,7 @@ def test_authority_full_nonasci() -> None:
     assert url.authority == "степан:пароль@слава.укр:8080"
 
 
-def test_authority_unknown_scheme():
+def test_authority_unknown_scheme() -> None:
     v = "scheme://user:password@example.com:43/path/to?a=1&b=2"
     url = URL(v)
     assert str(url) == v
--- a/tests/test_url_update_netloc.py
+++ b/tests/test_url_update_netloc.py
@@ -191,14 +191,14 @@ def test_with_port():
     assert str(url.with_port(8888)) == "http://example.com:8888"
 
 
-def test_with_default_port_normalization():
+def test_with_default_port_normalization() -> None:
     url = URL("http://example.com")
     assert str(url.with_scheme("https")) == "https://example.com"
     assert str(url.with_scheme("https").with_port(443)) == "https://example.com"
     assert str(url.with_port(443).with_scheme("https")) == "https://example.com"
 
 
-def test_with_custom_port_normalization():
+def test_with_custom_port_normalization() -> None:
     url = URL("http://example.com")
     u88 = url.with_port(88)
     assert str(u88) == "http://example.com:88"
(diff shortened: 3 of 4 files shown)


commonism:
@webknjaz @Dreamsorcerer - please review
In addition to getting more RFC3986 compliant by normalizing ports, this is part of a series (1/3) to implement URL.join without using urljoin.
Currently URL.join is implemented using using urllib.parse.urljoin, urljoin [strips](https://github.com/python/cpython/issues/84774). empty segments in violation of rfc3985

webknjaz:
Remind me on Monday evening if I forget, not earlier, plz.

yarl/_url.py
```diff
@@ -727,9 +758,7 @@ def _make_child(self, paths, encoded=False):
                     f"Appending path {path!r} starting from slash is forbidden"
                 )
             path = path if encoded else self._PATH_QUOTER(path)
-            segments = [
-                segment for segment in reversed(path.split("/")) if segment != "."
```
Dreamsorcerer, line 731: Not sure how removing this condition relates to anything?

yarl/_url.py
```diff
@@ -382,10 +393,15 @@ def is_default_port(self):
...
+        Return False for relative URLs
+
         """
-        if self.port is None:
-            return False
-        default = DEFAULT_PORTS.get(self.scheme)
+        if self.explicit_port is None:
+            """
+            a relative URL does not have an implicit port / default port
+            """
```
Dreamsorcerer, line 402:
```suggestion
            # A relative URL does not have an implicit port / default port
```

yarl/_url.py
```diff
@@ -291,6 +292,16 @@ def __str__(self):
         val = self._val
         if not val.path and self.is_absolute() and (val.query or val.fragment):
             val = val._replace(path="/")
+        if (port := self._get_port()) is None:
```
Dreamsorcerer, line 295:
```suggestion
        if (port := self._get_port()) is None:
            # https://datatracker.ietf.org/doc/html/rfc3986.html#section-6.2.3
```

Dreamsorcerer (approved):

yarl/_url.py
commonism: Good catch.
Artifact from the future, was meant to be in the next patches addressing URL.join. Removing "." breaks URL.join using URL.joinpath, it basically removes an empty element.

commonism:
> Remind me on Monday evening if I forget, not earlier, plz.

Remind

yarl/_url.py
```diff
@@ -382,10 +394,13 @@ def is_default_port(self):
         e.g. 'http://python.org' or 'http://python.org:80', False
         otherwise.
 
+        Return False for relative URLs
```
webknjaz, line 397: Let's have a full sentence here, at least:
```suggestion
        Return False for relative URLs.
```

yarl/_url.py
```diff
@@ -435,6 +450,21 @@ def raw_authority(self):
...
         return self._val.netloc
 
+    def _get_default_port(self):
+        port = None
+        if self.scheme:
+            port = DEFAULT_PORTS.get(self.scheme)
+            if port is None:
+                with suppress(OSError):
+                    port = socket.getservbyname(self.scheme)
+        return port
```
webknjaz, line 460: Let's reduce a few levels of nesting:
```suggestion
        if not self.scheme:
            return None

        with suppress(KeyError):
            return DEFAULT_PORTS[self.scheme]

        with suppress(OSError):
            return socket.getservbyname(self.scheme)

        return None
```

yarl/_url.py
```diff
@@ -435,6 +450,21 @@ def raw_authority(self):
         """
         return self._val.netloc
 
+    def _get_default_port(self):
```
webknjaz, line 453: And perhaps add typing:
```suggestion
    def _get_default_port(self) -> int | None:
```
(this probably needs a `from __future__ import annotations` at the top)

yarl/_url.py
```diff
@@ -291,6 +292,17 @@ def __str__(self):
         val = self._val
         if not val.path and self.is_absolute() and (val.query or val.fragment):
             val = val._replace(path="/")
+        if (port := self._get_port()) is None:
+            # https://datatracker.ietf.org/doc/html/rfc3986.html#section-6.2.3
```
webknjaz, line 296: Additionally to a link, it'd be nice to have a summary text right above it. As in, “how this branch is to be interpreted”.

tests/test_url_update_netloc.py
```diff
@@ -191,14 +191,32 @@ def test_with_port():
     assert str(url.with_port(8888)) == "http://example.com:8888"
 
 
+def test_with_port_normalization():
```
webknjaz, line 194: It looks like this contains 3 distinct tests. Perhaps, it's reasonable to split them into 3 functions so they show up separately in the test output.

CHANGES/1033.bugfix.rst
```diff
@@ -0,0 +1,4 @@
+Normalize default ports according to :rfc:`3986`.
+
+Specified ports are removed from the str representation of an URL if the port matches
+the schemes default port. -- by :user:`commonism`.
```
webknjaz, line 4: If you want the byline to be in the same paragraph, don't put a period before it.
```suggestion
the schemes default port -- by :user:`commonism`.
```

Alternatively, you could attach it to the very first sentence, too. Another possibility is to drop the trailing period and add an empty line as a separator so it's not a part of the sentence and the paragraph.

yarl/_url.py
Dreamsorcerer: Can we stick to:
```suggestion
    def _get_default_port(self) -> Optional[int]:
```
So, it still works when evaluating annotations at runtime, and just let pyupgrade change it when the time comes.

yarl/_url.py
```diff
@@ -435,6 +450,24 @@ def raw_authority(self):
         """
         return self._val.netloc
 
+    def _get_default_port(self) -> Optional[int]:
```
webknjaz, line 453: Optional is discouraged because it's considered confusing:
```suggestion
    def _get_default_port(self) -> Union[int, None]:
```

CHANGES/1033.bugfix.rst
```diff
@@ -0,0 +1,4 @@
+Normalize default ports according to :rfc:`3986`.
+
+Specified ports are removed from the str representation of an URL if the port matches
+the schemes default port -- by :user:`commonism`.
```
webknjaz, line 4: Let's link the mentioned structures and fix the grammar:
```suggestion
Specified ports are removed from the :class:`str` representation of a :class:`~yarl.URL`
if the port matches the scheme's default port -- by :user:`commonism`.
```

tests/test_url.py
```diff
@@ -219,6 +219,12 @@ def test_authority_full_nonasci() -> None:
     assert url.authority == "степан:пароль@слава.укр:8080"
 
 
+def test_authority_unknown_scheme():
```
webknjaz, line 222:
```suggestion
def test_authority_unknown_scheme() -> None:
```

tests/test_url_update_netloc.py
```diff
@@ -191,14 +191,37 @@ def test_with_port():
     assert str(url.with_port(8888)) == "http://example.com:8888"
 
 
+def test_with_default_port_normalization():
```
webknjaz, line 194:
```suggestion
def test_with_default_port_normalization() -> None:
```

tests/test_url_update_netloc.py
```diff
@@ -191,14 +191,37 @@ def test_with_port():
...
+    url = URL("http://example.com")
+    assert str(url.with_scheme("https")) == "https://example.com"
+    assert str(url.with_scheme("https").with_port(443)) == "https://example.com"
+    assert str(url.with_port(443).with_scheme("https")) == "https://example.com"
+
+
+def test_with_custom_port_normalization():
```
webknjaz, line 201:
```suggestion
def test_with_custom_port_normalization() -> None:
```

tests/test_url_update_netloc.py
```diff
@@ -191,14 +191,37 @@ def test_with_port():
...
+    url = URL("http://example.com")
+    u88 = url.with_port(88)
+    assert str(u88) == "http://example.com:88"
+    assert str(u88.with_port(80)) == "http://example.com"
+    assert str(u88.with_scheme("https")) == "https://example.com:88"
+
+
+def test_with_explicit_port_normalization():
```
webknjaz, line 209:
```suggestion
def test_with_explicit_port_normalization() -> None:
```

yarl/_url.py
```diff
@@ -435,6 +452,24 @@ def raw_authority(self):
...
+
+        with suppress(KeyError):
+            return DEFAULT_PORTS[self.scheme]
+
+        with suppress(OSError):
+            return socket.getservbyname(self.scheme)
+
+        return None
+
+    def _get_port(self):
```
webknjaz, line 467:
```suggestion
    def _get_port(self) -> Union[int, None]:
```

CHANGES/1033.bugfix.rst
webknjaz: @commonism I didn't commit this myself since I wanted you to preview it and check if I didn't confuse things. But I expect this to be the last bit to correct before merging.

CHANGES/1033.bugfix.rst
```diff
@@ -0,0 +1,4 @@
+Normalize default ports according to :rfc:`3986`.
```
webknjaz, line 1: Urgh.. I somehow missed that this was in imperative mood for some reason. It should refer to the change that has already happened since the previous release.

webknjaz:
```suggestion
The default protocol ports of well-known URI schemes are now taken into account
during the normalization of the URL string representation in accordance with
:rfc:`3986#section-3.2.3`.
```

webknjaz (approved):

PR #1039, opened 2024-07-18
RFC3986 compatible URL.join honoring empty segments

commonism:
## What do these changes do?

The changes re-implement URL.join within yarl.URL instead of using urllib.parse.urljoin to avoid known issues when joining paths with empty segments (https://github.com/python/cpython/issues/84774)

Due to the unit tests for URL.join it was required to align the query parameter rendering to match the unit tests, empty parameters are rendered without value therefore instead of an empty value.

## Checklist


commonism:
Towncrier ain't happy with my intersphinx link to python docs:
```
/home/runner/work/yarl/yarl/docs/[towncrier-fragments]:45: WARNING: py:meth reference target not found: python.urllib.parse.urljoin
```

commonism:
As 
```
:py:function:`urllib.parse.urljoin`
```
does not work either
> /home/runner/work/yarl/yarl/docs/changes.rst:45: ERROR: Unknown interpreted text role "py:function".

But according to
```
python3 -m sphinx.ext.intersphinx https://docs.python.org/3/objects.inv | grep -E "(^py:.*|urllib.parse.urljoin)"
…
py:function
	urllib.parse.urljoin                     library/urllib.parse.html#urllib.parse.urljoin
…
```
this should work and the intersphinx is set up in the job

```
loading intersphinx inventory 'python' from https://docs.python.org/3/objects.inv...
loading intersphinx inventory 'multidict' from https://multidict.aio-libs.org/en/stable/objects.inv...
```

@Dreamsorcerer @webknjaz any ideas?

webknjaz:
@commonism use `:func:`. That table shows object types with differs from the role names sometimes.

In my helper, I tried replacing them with proper reference hints: https://webknjaz.github.io/intersphinx-untangled

webknjaz:
> Towncrier ain't happy with my intersphinx link to python docs:

Technically, it's not Towncrier that's unhappy. Towncrier is just a fancy template renderer. But when that's injected into Sphinx (through my `sphinxcontrib-towncrier` extension), it can emit warnings and we turn them into errors.

commonism:
The python domain maps _function_ to _func_, therefore it is :py:func:
https://github.com/sphinx-doc/sphinx/blob/685e3fdb49c42b464e09ec955e1033e2a8729fff/sphinx/domains/python.py#L845-L881

To many docs for intersphinx show :py:function: as example.

That said, mastered this as well.

yarl/_url.py
```diff
@@ -1153,7 +1166,49 @@ def join(self, url):
...
             raise TypeError("url should be URL")
-        return URL(urljoin(str(self), str(url)), encoded=True)
+        other: URL = url
+        scheme = other.scheme or self.scheme
+        parts = {
+            k: getattr(self, k) for k in ["authority", "path", "query", "fragment"]
+        }
+        parts["scheme"] = scheme
+        from urllib.parse import uses_netloc as uses_authority
+        from urllib.parse import uses_relative
```
Dreamsorcerer, line 1176: These should presumably be at the top of the file.

yarl/_url.py
```diff
@@ -1153,7 +1166,49 @@ def join(self, url):
...
+
+        if scheme != self.scheme or scheme not in uses_relative:
+            return URL(str(other))
+
+        if scheme in uses_authority:
+            if other.authority:
+                parts.update(
+                    {
+                        k: getattr(other, k)
```
Dreamsorcerer, line 1185: Are you sure about this one? It will replace the parts with `None` if the attributes don't exist.

tests/test_update_query.py
```diff
@@ -97,7 +97,7 @@ def test_with_query_empty_str():
 
 def test_with_query_empty_value():
     url = URL("http://example.com/")
-    assert str(url.with_query({"a": ""})) == "http://example.com/?a="
+    assert str(url.with_query({"a": ""})) == "http://example.com/?a"
```
Dreamsorcerer, line 100: What is the rationale for this change? I can't find anything in the URI or HTTP spec that suggests this normalisation.

yarl/_url.py
commonism: Yes.
If the "other" has an authority, we use it's path/query/fragment.

c.f. 
https://github.com/python/cpython/blob/a9bb3c7b3bd8ba90de87d03cd4d614dca764b116/Lib/urllib/parse.py#L555-L557

tests/test_update_query.py
commonism: Rendering empty parameters with assignment does not work with the rfc3986 test cases:

https://github.com/aio-libs/yarl/blob/6b13867652ebcfad668f6e19146cce0b7ae4bc2c/tests/test_url.py#L1604

would be 
```
http://a/b/c/g?y=#s
```
which fails the unit tests

Dreamsorcerer: I don't think it says anything about how to format the query string. The examples simply show the query unmodified. As far as RFC3986 is concerned, a query string is just a sequence of characters, and therefore shouldn't be modified during normalisation.

commonism: Alternative I considered was using ?a= to store "" and ?a to store None and keep rendering the empty string value "" using an empty assignment. Adds quite some complexity for no additional value and using None would break the use case to remove a query parameter as discussed in https://github.com/aio-libs/yarl/issues/1010.

commonism: > a query string is just a sequence of characters, and therefore shouldn't be modified during normalisation.

The unit tests of rfc3986 use parameters without assignment, serialization has to render those without assignment.
yarl does not have a notion for query parameters which been assigned with empty value or without assignment.
Due to the lack of notion if a parameters was assigned with assignment or without, the only option is to serialize all parameters without value as parameters without assignment.

Previously ?a was serialized as ?a=, now it is the other way around.

Dreamsorcerer: > The unit tests of rfc3986 use parameters without assignment, serialization has to render those without assignment.

No, what I'm saying is that they show a query string (`y`) and after applying normalisation to that query string it still results in `y`. The RFC has no concept of parameters or assignments, it's just a string of characters that exclude `#`.

For encoding a set of parameters into a query string, we need to look elsewhere. I've found these steps from the W3C:
https://www.w3.org/TR/2014/REC-html5-20141028/forms.html#url-encoded-form-data

When presented with a form containing an empty 'y' parameter, these don't appear to mention or provide any steps that would result in a query string of `y`. It would always produce `y=`, so I think this is the best approach for serialising parameters, that should ensure best compatibility with other software.

commonism: The unit tests of rfc3986 are join operations, with yarl is this includes encoding the parameters to have a string representation of the final url, this would add an `=` for empty parameters.

Proceed with adjusting the unit tests "the other direction"?

Dreamsorcerer: In this particular example, there's no join, just a replacing of the query string from a dict of parameters. So, yes, I think move to producing `?a=`.

commonism: This not about this particular example, but about encoding parameters in general.
Once it's a URL, the general encoding strategy is applied.

I adjusted the expectations of the rfc3986 unit tests to make it work accordingly.

Dreamsorcerer: Hmm, I'm guessing the issue is that we are decoding the query string to parameters and then reencoding it. Ideally the rfc3986 tests should not need changing as the query string is a string literal in the URL and should therefore remain untouched. But, the .with_query() etc. code that encodes parameters should result in a `y=` format.

@webknjaz Any thoughts on how this is approached.

commonism: Using query_string now, no encode required, query remains.

Dreamsorcerer (approved):
I think this looks right to me. @webknjaz Can you check it over too?

CHANGES/1039.bugfix.rst
```diff
@@ -0,0 +1,25 @@
+:meth:`URL.join <yarl.URL.join>` has been changed to match
```
webknjaz, line 1:
```suggestion
:meth:`URL.join() <yarl.URL.join>` has been changed to match
```
(consistency with another func ref)

CHANGES/1039.bugfix.rst
webknjaz, line 1: Change note preview link for easy access: https://yarl--1039.org.readthedocs.build/en/1039/changes/#bug-fixes

CHANGES/1039.bugfix.rst
```diff
@@ -0,0 +1,25 @@
...
+:rfc:`3986` and align with
+:meth:`/ operation <yarl.URL.__truediv__>` and :meth:`URL.joinpath() <yarl.URL.joinpath>`
+when joining URLs with empty segments.
+Previously :py:func:`urllib.parse.urljoin` was used,
+which has known issues with empty segments
+(`python/cpython#84774 <https://github.com/python/cpython/issues/84774>`_).
+
+Due to the semantics of :meth:`URL.join <yarl.URL.join>`, joining an
```
webknjaz, line 9:
```suggestion
Due to the semantics of :meth:`URL.join() <yarl.URL.join>`, joining an
```

CHANGES/1039.bugfix.rst
```diff
@@ -0,0 +1,25 @@
...
+URL with scheme requires making it relative, prefixing with ./.
```
webknjaz, line 10: did you mean this?
```suggestion
URL with scheme requires making it relative, prefixing with ``/``.
```

webknjaz: Oh, it's probably
```suggestion
URL with scheme requires making it relative, prefixing with ``./``.
```

tests/test_url.py
```diff
@@ -1657,6 +1657,78 @@ def test_join_from_rfc_3986_abnormal(url, expected):
...
+URLLIB_URLJOIN_FAIL = [
+    # FIXME? - host is None
+    ("http:///", "..", "http:///"),
+    # issue 23703: don't duplicate filename
+    # FIXME? path without schema is not normalized
+    ("a", "b", "b"),
+]
+
+
+@pytest.mark.xfail(strict=True)
```
webknjaz, line 1723: plz always specify a `reason=` in xfail and skip markers

webknjaz: Also, what's the current behavior? Does it raise an exception? Use `raises=` to specify which one.

commonism: One problem is - 
> URL("http:///").authority # host is None
https://github.com/aio-libs/yarl/blob/6b13867652ebcfad668f6e19146cce0b7ae4bc2c/yarl/_url.py#L836
> TypeError("unsupported operand type(s) for +: 'NoneType' and 'str'")

Using "" for the host may be valid here too, but this had been evaluated already and was too dangerous - c.f. https://github.com/python/cpython/pull/15642

> URL("a").join(URL("b")) # .join uses URL.build with scheme None - 
> TypeError('NoneType is illegal for "scheme", "authority", "host", "path", "query_string", and "fragment" args, use empty string instead.')

I'm open to ideas how to work those.

commonism: Hi,

we've had 2 remaining issues - 
- abb - `URL("a").urljoin(URL("b")) == URL("b")`
- empty-string-host - URL("http:///").urljoin("..")

abb has been solved.
For empty-string-host - I'd leave this as is, yarl has this inherited from urllib. Can't be changed without breaking lot's of things, neglect able benefit, not worth the effort without an actual use case.

commonism:
linter currently dies with:
```
Extension error (sphinxcontrib.towncrier.ext):
Handler <bound method TowncrierDraftEntriesEnvironmentCollector.get_outdated_docs of <sphinxcontrib.towncrier.ext.TowncrierDraftEntriesEnvironmentCollector object at 0x7fd1132dac90>> for event 'env-get-outdated' threw an exception (exception: find_fragments() takes 3 positional arguments but 4 were given)
make[1]: *** [Makefile:228: spelling] Error 2
updating environment: make[1]: Leaving directory '/home/runner/work/yarl/yarl/docs'
make: *** [Makefile:64: doc-spelling] Error 2
Error: Process completed with exit code 2.
```

Dreamsorcerer:
> linter currently dies with:

Looks like there's a problem with the requirements files. It ends up installing the latest towncrier instead of the pinned version.

yarl/_url.py
```diff
@@ -1153,7 +1158,52 @@ def join(self, url):
...
+        parts = {
+            k: getattr(self, k) or ""
+            for k in ["authority", "path", "query_string", "fragment"]
+        }
+        parts["scheme"] = scheme
+
+        if scheme != self.scheme or scheme not in uses_relative:
+            return URL(str(other))
+
+        if scheme in uses_authority:
```
bdraco, line 1172: It looks like there is a branch miss on coverage for this line.

commonism: > 774, 1172->1182, 1268-1269

I tried, setting a breakpoint at 1173 or 1170 works for me?

Dreamsorcerer: If codecov is correct, then we're also missing lines 1164-1165 an `if not segments` on line 773 is never True anymore.

bdraco:
Added skip label as a workaround since changelog bot is currently not working

yarl/_url.py
Dreamsorcerer: I don't think it's possible for 1164-1165 to not run, so I that's a bug in the coverage...

commonism:
```python
        parts = {
            k: getattr(self, k) or ""
            for k in ["authority", "path", "query_string", "fragment"]
        }
```
guess coverage does not like the `or ""`

bdraco:
close/reopen to rerun ci

yarl/_url.py
```diff
@@ -1153,7 +1156,52 @@ def join(self, url):
         # See docs for urllib.parse.urljoin
         if not isinstance(url, URL):
             raise TypeError("url should be URL")
-        return URL(urljoin(str(self), str(url)), encoded=True)
+        other: URL = url
+        scheme = other.scheme or self.scheme
+        parts = {
+            k: getattr(self, k) or ""
+            for k in ["authority", "path", "query_string", "fragment"]
```
bdraco, line 1163:
```suggestion
            for k in ("authority", "path", "query_string", "fragment")
```

yarl/_url.py
```diff
@@ -1153,7 +1156,52 @@ def join(self, url):
...
+
+        if scheme != self.scheme or scheme not in uses_relative:
+            return URL(str(other))
+
+        # scheme is in uses_authority as uses_authority is a superset of uses_relative
+        if scheme in uses_authority and other.authority:
+            parts.update(
+                {
+                    k: getattr(other, k) or ""
+                    for k in ["authority", "path", "query_string", "fragment"]
```
bdraco, line 1175:
```suggestion
                    for k in ("authority", "path", "query_string", "fragment")
```

yarl/_url.py
```diff
@@ -1153,7 +1156,52 @@ def join(self, url):
...
+        parts = {
+            k: getattr(self, k) or ""
+            for k in ("authority", "path", "query_string", "fragment")
+        }
+        parts["scheme"] = scheme
+
+        if scheme != self.scheme or scheme not in uses_relative:
+            return URL(str(other))
+
+        # scheme is in uses_authority as uses_authority is a superset of uses_relative
```
Dreamsorcerer, line 1170: Sorry, doesn't this comment mean that the left-side is always True? Shouldn't we just remove it then?

commonism: other way around - if you got to 1170, 
* scheme is in uses_relative, 
* uses_relative is a subset of uses_authority, 
* therefore the first part of the check in 1171 is always true
 
just kept it kept for comparison with urllib

CHANGES/1039.bugfix.rst
```diff
@@ -0,0 +1,25 @@
...
+Previously :py:func:`urllib.parse.urljoin` was used,
+which has known issues with empty segments
+(`python/cpython#84774 <https://github.com/python/cpython/issues/84774>`_).
+
+Due to the semantics of :meth:`URL.join() <yarl.URL.join>`, joining an
+URL with scheme requires making it relative, prefixing with ``./``.
+
+.. code-block:: python
+
+   URL("https://web.archive.org/web/").join(URL("./https://github.com/aio-libs/yarl"))
```
webknjaz, line 14: Just got an idea: what if we convert the code blocks here from `python` to `pycon` (as in “python console”: https://pygments.org/docs/lexers/#pygments.lexers.python.PythonConsoleLexer). The syntax would need to match the default REPL. So the first line would be prefixed with `>>> ` and the following one would be the output. This would get highlighted appropriately and demonstrate better what result is expected.
This could later be used as a doctest (if we enable it in sphinx).

tests/test_url.py
```diff
@@ -1657,6 +1657,83 @@ def test_join_from_rfc_3986_abnormal(url, expected):
...
+        "http:///",
+        "..",
+        "http:///",
+        TypeError,
+        "unsupported operand type\\(s\\) for \\+: 'NoneType' and 'str'",
+    ),
+]
+
+
+@pytest.mark.xfail(raises=TypeError, strict=True)
```
webknjaz, line 1726: In https://github.com/aio-libs/yarl/pull/1039#discussion_r1690120472, I was asking to add a `reason=` param. Have you seen that? Every `xfail` and `skip` should have a reason documented. Sometimes, it's a link to a GitHub issue. Though, I prefer some explanatory text in addition to that.

tests/test_url.py
```diff
@@ -1657,6 +1657,83 @@ def test_join_from_rfc_3986_abnormal(url, expected):
...
+@pytest.mark.parametrize("base,url,expected,exc,match", URLLIB_URLJOIN_FAIL)
+def test_join_cpython_urljoin_fail(base, url, expected, exc, match):
+    # non portable tests from cpython urljoin
+    base = URL(base)
+    url = URL(url)
+    with pytest.raises(exc, match=match) as e:
+        base.join(url)
+    raise e.value
```
webknjaz, line 1734: Why have a `raise` statement in tests? What does the test check, then?

webknjaz: If an unexpected exception is raised (or none at all), `pytest.raises()` would display it already.

tests/test_url.py
```diff
@@ -1657,6 +1657,83 @@ def test_join_from_rfc_3986_abnormal(url, expected):
...
+    assert joined == expected
+
+
+URLLIB_URLJOIN_FAIL = [
+    (
+        "http:///",
+        "..",
+        "http:///",
+        TypeError,
+        "unsupported operand type\\(s\\) for \\+: 'NoneType' and 'str'",
```
webknjaz, line 1721: It's customary to use raw-strings for regexes, as it lets us have less confusing escaping:
```suggestion
        r"unsupported operand type\(s\) for \+: 'NoneType' and 'str'",
```

tests/test_url.py
commonism: pytest.raises() is used to match the exceptions message, re-raised to have a xfail.

Dreamsorcerer: I'm not sure if there's some confusion over how xfail works here.

xfail means the test will fail because something in our code is broken, and we should fix it later. Whereas this test just seems to fail for no reason.

pytest.raises() will fail a test if an exception is not raised. So, if the expected case is that exceptions are raised, then removing the `raise` and removing the `xfail` would give us a correct, working test.

CHANGES/1039.bugfix.rst
commonism: testing

commonism: added the expected result

webknjaz: thanks! it's much more visual now!

tests/test_url.py
webknjaz: I reopened this thread since it wasn't addressed. It sounds like this might not even need to be an xfailing test.

@commonism could you explain more verbosely what's the intent and design behind this specific test?

commonism: xfail does not allow to specify a match for the exception string, pytest.raises does.
Combining both allows to create a tight testing scenario for broken behavior, and indicate this current behavior is considered broken and would be great to have addressed `later`.

I hope this explains my intention - I just wanted to include the match for the exception string in the test.

bdraco: It looks like this test is written to verify that the code produces the wrong result.

Test should be written to indicate what we expect the code should do, and than marked with xfail to indicate the code currently does not work as expected.

Dreamsorcerer: > xfail does not allow to specify a match for the exception string, pytest.raises does.

So what's happening here, is that the test is expecting the code to not raise an exception.
We then catch the exception with pytest.raises() in order to use the match argument.
Then we reraise that exception in order to turn it back into an xfail.

This is causing so much confusion, I'd vote for just doing without the match argument and remove the pytest.raises().

Dreamsorcerer: Or possibly, just replace the `raise` with `pytest.xfail()` and xfail it in the test code instead.

tests/test_url.py
```diff
@@ -1657,6 +1657,70 @@ def test_join_from_rfc_3986_abnormal(url, expected):
...
+    assert joined == expected
+
+
+@pytest.mark.xfail(raises=TypeError, strict=True, reason="empty host name")
+def test_join_cpython_urljoin_fail():
+    with pytest.raises(
+        TypeError, match=r"unsupported operand type\(s\) for \+: 'NoneType' and 'str'"
+    ) as e:
+        URL("http:///").join(URL(".."))
+    raise e.value
```
Dreamsorcerer, line 1721: Maybe this is a lot clearer?
```suggestion
def test_join_cpython_urljoin_fail():
    with pytest.raises(
        TypeError, match=r"unsupported operand type\(s\) for \+: 'NoneType' and 'str'"
    ) as e:
        URL("http:///").join(URL(".."))
    pytest.xfail("Shouldn't raise TypeError on empty host name")
```

bdraco:
Ran a profile.  Only thing I noticed is `schema` comes up a lot in `yarl`.  Probably would make sense to put that one in the dict to and avoid the property call.. but thats another issue

bdraco:
Tested one production.  No adverse effect observed

bdraco (approved):

bdraco:
Thanks @commonism

bdraco:
This PR introduced a regression in decoding query string params in joined urls.

I attempted to fix it in https://github.com/aio-libs/yarl/pull/1066 but the fix looks more involved so I'm going to revert this PR and do another release

I think this needs to be reworked to use `URL(self._val._replace(.....), encoded=True)` instead like some of the other methods

PR #954, opened 2023-11-20
Validate ASCII host values

mjpieters:
Host values must match the unreserved, sub-delims or pct-encoded grammar
rules. The IDNA encoder already checks for this, but for ASCII values we
skip IDNA encoding.

If the invalid character is `@`, or `:` followed by `@` further in the
host value, tack on advice about using `authority` instead of `host`.

Fixes #880

Screenshot 2024-09-07 at 5 19 09 PM

## Checklist

- [x] I think the code is well written
- [x] Unit tests for the changes exist
- [ ] Documentation reflects the changes
- [x] Add a new news fragment into the `CHANGES` folder
  * name it `<issue_id>.<type>` (e.g. `588.bugfix`)
  * if you don't have an `issue_id` change it to the pr id after creating the PR
  * ensure type is one of the following:
    * `.feature`: Signifying a new feature.
    * `.bugfix`: Signifying a bug fix.
    * `.doc`: Signifying a documentation improvement.
    * `.removal`: Signifying a deprecation or removal of public API.
    * `.misc`: A ticket has been closed, but it is not of interest to users.
  * Make sure to use full sentences with correct case and punctuation, for example: `Fix issue with non-ascii contents in doctest text files.`

PR #1044, opened 2024-07-29
Make raw_host and port cached_property

bdraco:
## What do these changes do?

aiohttp accesses these properties quite often because of the calls to `is_absolute`/`update_headers` and `urllib` has to reparse `_hostinfo` each time.

No

port
raw_host


bdraco:
https://yarl--1044.org.readthedocs.build/en/1044/changes/#miscellaneous-internal-changes

CHANGES/1044.misc.rst
```diff
@@ -0,0 +1,6 @@
+``raw_host`` and ``port`` are now ``cached_property`` -- by :user:`bdraco`.
+
+``aiohttp`` accesses these properties quite often, which cause
+``urllib`` to build the ``_hostinfo`` property every time.
+``raw_host`` and ``port`` are now cached properties, which will
+improve performance.
```
bdraco, line 6:
```suggestion
``aiohttp`` accesses these properties quite often, which cause ``urllib`` to build the ``_hostinfo`` property every time. ``raw_host`` and ``port`` are now cached properties, which will improve performance.
```

CHANGES/1044.misc.rst
```diff
@@ -0,0 +1,3 @@
+``raw_host`` and ``port`` are now ``cached_property`` -- by :user:`bdraco`.
+
+``aiohttp`` accesses these properties quite often, which cause ``urllib`` to build the ``_hostinfo`` property every time. ``raw_host`` and ``port`` are now cached properties, which will improve performance.
```
webknjaz, line 3: Let's try to link external docs:
```suggestion
:mod:`aiohttp` accesses these properties quite often, which cause :mod:`urllib` to build the ``_hostinfo`` property every time. ``raw_host`` and ``port`` are now cached properties, which will improve performance.
```

yarl/_url.py
```diff
@@ -521,7 +521,7 @@ def password(self):
         """
         return self._UNQUOTER(self.raw_password)
 
-    @property
+    @cached_property
```
webknjaz, line 524: Does this mean that the tests need to be updated to access `__wrapped__`?

bdraco: Since the URL objects are immutable we shouldn't have to worry about the value every changing or needing to test `__wrapped__`.  I don't see any of the other tests using `cached_property` doing/needing that.

CHANGES/1044.misc.rst
```diff
@@ -0,0 +1,3 @@
+``raw_host`` and ``port`` are now ``cached_property`` -- by :user:`bdraco`.
+
+:mod:`aiohttp` accesses these properties quite often, which cause :mod:`urllib` to build the ``_hostinfo`` property every time. ``raw_host`` and ``port`` are now cached properties, which will improve performance.
```
bdraco, line 3:
```suggestion
``aiohttp`` accesses these properties quite often, which cause :mod:`urllib` to build the ``_hostinfo`` property every time. ``raw_host`` and ``port`` are now cached properties, which will improve performance.
```

bdraco: Docs build failed with `:mod:`aiohttp``

bdraco:
I tried to fix the chronographer in https://github.com/aio-libs/yarl/pull/1053 but it looks like its coming in from the org level and not in the github config so its not something I have access to solve.

bdraco:
It looks like changelog checks haven't been working for a while per https://github.com/aio-libs/yarl/pull/954#issuecomment-1819264845

bdraco:
Added skip label as a workaround since changelog bot is currently not working

bdraco:
Screenshot 2024-08-14 at 9 58 09 AM

PR #1052, opened 2024-08-14
Enforce towncrier requirements for docs to fix CI

commonism:
Starting with the release of towncrier 24.7 the CI/CD regulary fails as the linter does not work.

sphinxcontrib-towncrier does have an upper boundary for the towncrier version and is incompatible to 24.7.
c.f. https://github.com/sphinx-contrib/sphinxcontrib-towncrier/issues/92

Installing tc 23.11 for docs to avoid getting 24.7.

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -1,2 +1,2 @@
 -r test.txt
-towncrier==23.11.0
+-r towncrier.txt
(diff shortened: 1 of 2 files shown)


bdraco (approved):
Thanks @commonism 

I started looking into this yesterday but ran out of time.  Appreciate the fix

bdraco:
Looks like Codecov is stuck. Will try close reppen

PR #1053, opened 2024-08-14
Fix chronographer not finding changes

bdraco:
## What do these changes do?
Fix chronographer not finding changes by symlinking `news` to `CHANGES`.

There is likely a much better solution.

### DIFF
--- a/news
+++ b/news
@@ -1,0 +1,1 @@
+CHANGES


bdraco:
Ahh.. It looks like there is no rst support yet here

bdraco:
I don't have access to fix this. It will need to wait for someone who does.

PR #1057, opened 2024-08-26
Stop unquoting %2F in paths

Dreamsorcerer:
I've added ignore so that `/foo/bar%2fbaz` doesn't become `%2Ffoo%2Fbar%2Fbaz`, but maybe the `unsafe` option is just wrong currently?


bdraco (approved):
After looking at history and considering security perspective with this, I think this is the right fix

bdraco:
skip label added as a workaround

bdraco:
I'm going to test this on some Home Assistant production instances with ingress just to be sure.

bdraco:
Just realized I had already updated my primary production with this since it was in my integration branch.

No obvious breakage

CHANGES/1057.breaking.rst
```diff
@@ -0,0 +1,2 @@
+Stopped decoding ``%2F`` (``/``) in ``URL.path``, as this could lead to code incorrectly being treated as a path separator
```
Dreamsorcerer, line 1:
```suggestion
Stopped decoding ``%2F`` (``/``) in ``URL.path``, as this could lead to code incorrectly treating it as a path separator
```

Dreamsorcerer:
Damn, beat me by a couple of seconds.. :P

CHANGES/1057.breaking.rst
bdraco: I'll fix it up when I do the release

ofekashery:
Hi @Dreamsorcerer @bdraco @webknjaz, thanks for your hard work ❤️ Please pay attention that this PR introduces BRAKING changes and might not be considered as a patch.

It led to a bug in the AMQP connection string builder of [mosquito/aio-pika](https://github.com/mosquito/aio-pika), and broke our critical production system.

I believe those changes are great but could be as an optional flag. It also could reflected better in the changelog.

webknjaz:
@ofekashery would you mind opening a new issue and maybe also submitting a PR with xfailing tests per https://blog.ganssle.io/articles/2021/11/pytest-xfail.html? Comments on closed things may end up getting lost. I'm not sure if @Dreamsorcerer and @bdraco got your message in time, for example.

Dreamsorcerer:
Yeah, should probably have been a bigger version bump, but not too much we can do now. As it relates to security, I think it's still important to keep the change going forwards.

webknjaz:
Yeah, security changes are always exempt from SemVer.

Dreamsorcerer:
Sorry about that, we've decided to revert this anyway as it's caused a few more issues. If you need to know the path separators for safety, we're adding a path_safe attribute instead which skips decoding %2F and %25.

PR #1060, opened 2024-08-30
Release 1.9.5

bdraco:
Screenshot 2024-08-30 at 1 34 43 PM
Screenshot 2024-08-30 at 1 34 52 PM
Screenshot 2024-08-30 at 1 35 01 PM
Screenshot 2024-08-30 at 1 35 06 PM


bdraco:
Doing some retesting on a production Home Assistant before I merge

bdraco:
All good. No errors or increase in exceptions/m

webknjaz:
@bdraco FWIW, I usually don't merge the `release/v1.9.5` branch until after the release. When the release succeeded, I can then `git merge --no-ff` it so that the tagged commit becomes a part of `master` (the key here is avoiding squashes or any other fake merges). This also allows having a commit like #1061 in the same branch right before merging it in.
This flow allows force-pushing stuff until I get things right..

PR #1062, opened 2024-08-31
Refactor cached_property to improve performance

bdraco:
## What do these changes do?
stdlib `cached_property` takes advantage of the performance benefits of storing the value in [`__dict__`](https://github.com/python/cpython/blob/d5abd02f36bbee5720944b8906a118a8fb66d75b/Lib/functools.py#L984) which avoids the `__get__` call. This is much faster since everything can happen in native code to fetch the value. When support for python older than 3.12 is removed, the stdlib `cached_property` can be used as versions before 3.12 had an undocumented class level lock that made the performance profile undesirable.

no

### DIFF
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -32,24 +32,27 @@ class cached_property:
     effectively replacing the function it decorates with an instance
     variable.  It is, in Python parlance, a data descriptor.
 
+    This is trimmed down version of the `cached_property` decorator from
+    the standard library which does not have the undocumented locking
+    behavior that was removed in Python 3.12.
     """
+
+    # When support for Python 3.11 is dropped, this can be replaced with
+    # functools.cached_property.
 
     def __init__(self, wrapped):
         self.wrapped = wrapped
-        try:
-            self.__doc__ = wrapped.__doc__
-        except AttributeError:  # pragma: no cover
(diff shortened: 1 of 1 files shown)


yarl/_url.py
```diff
@@ -151,8 +154,6 @@ class URL:
     #               / path-noscheme
     #               / path-empty
     # absolute-URI  = scheme ":" hier-part [ "?" query ]
-    __slots__ = ("_cache", "_val")
```
bdraco, line 154: `__slots__` is not compatible with `__dict__`

yarl/_url.py
```diff
@@ -32,24 +32,27 @@ class cached_property:
...
 
+    # When support for Python 3.11 is dropped, this can be replaced with
+    # functools.cached_property.
+
     def __init__(self, wrapped):
         self.wrapped = wrapped
-        try:
-            self.__doc__ = wrapped.__doc__
-        except AttributeError:  # pragma: no cover
-            self.__doc__ = ""
```
bdraco, line 42: This was actually unreachable as __doc__ will never not be there https://github.com/python/cpython/blob/d5abd02f36bbee5720944b8906a118a8fb66d75b/Lib/functools.py#L966

PR #1066, opened 2024-08-31
Add additional coverage for joining urls

bdraco:
## What do these changes do?

#1039 introduced a regression in decoding query string params with `%2B` (aka `+`). Add coverage to ensure we do not miss this again in the future.

no

## Related issue number
#1039

### DIFF
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -1790,3 +1790,35 @@ def test_requoting():
     u = URL("http://127.0.0.1/?next=http%3A//example.com/")
     assert u.raw_query_string == "next=http://example.com/"
     assert str(u) == "http://127.0.0.1/?next=http://example.com/"
+
+
+def test_join_query_string():
+    """Test that query strings are correctly joined."""
+    original = URL("http://127.0.0.1:62869")
+    path_url = URL(
+        "/api?start=2022-03-27T14:05:00%2B03:00&end=2022-03-27T16:05:00%2B03:00"
+    )
+    assert path_url.query.get("start") == "2022-03-27T14:05:00+03:00"
+    assert path_url.query.get("end") == "2022-03-27T16:05:00+03:00"
+    new = original.join(path_url)
+    assert new.query.get("start") == "2022-03-27T14:05:00+03:00"
+    assert new.query.get("end") == "2022-03-27T16:05:00+03:00"
+
+
+def test_join_query_string_with_special_chars():
+    """Test url joining when the query string has non-ascii params."""
+    original = URL("http://127.0.0.1")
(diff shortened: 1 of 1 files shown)


tests/test_url.py
```diff
@@ -1853,3 +1853,35 @@ def test_requoting():
...
+        "/api?start=2022-03-27T14:05:00%2B03:00&end=2022-03-27T16:05:00%2B03:00"
+    )
+    assert original.query_string == ""
+    assert (
+        path_url.raw_query_string
+        == "start=2022-03-27T14:05:00%2B03:00&end=2022-03-27T16:05:00%2B03:00"
+    )
+    assert (
+        path_url.query_string
+        == "start=2022-03-27T14:05:00%2B03:00&end=2022-03-27T16:05:00%2B03:00"
```
bdraco, line 1871: Is this incorrect, and thats where the regression is coming from?

bdraco:
I think #1039 needs to be reworked to use `URL(
            self._val._replace(.....),
            encoded=True,
        )` instead

tests/test_url.py
```diff
@@ -1790,3 +1790,35 @@ def test_requoting():
...
+    )
```
bdraco, line 1809:
```suggestion
```

tests/test_url.py
bdraco: I took this out as I'm not sure. We only need to test the regression anyways

bdraco:
`@codecov codecov/patch — 0.00% of diff hit (target 62.37%)` isn't passing so this one won't merge as its marked as required

webknjaz:
@bdraco this should probably be force-merged for now. It's likely because of the Codecov v4 bug we saw earlier. I'm yet to file a report upstream.

bdraco:
@webknjaz looks like I don’t have the proper GitHub access to override checks for this repo.

tests/test_url.py
```diff
@@ -1790,3 +1790,35 @@ def test_requoting():
     u = URL("http://127.0.0.1/?next=http%3A//example.com/")
     assert u.raw_query_string == "next=http://example.com/"
     assert str(u) == "http://127.0.0.1/?next=http://example.com/"
+
+
+def test_join_query_string():
+    """Test that query strings are correctly joined."""
+    original = URL("http://127.0.0.1:62869")
+    path_url = URL(
+        "/api?start=2022-03-27T14:05:00%2B03:00&end=2022-03-27T16:05:00%2B03:00"
```
bdraco, line 1799: This line is obviously covered so something still not right with multi-line coverage

Issue #301, opened 2019-04-11
Why does yarl encode query params differently?


bsolomon1124:
**Note:** I'm not filing this issue to say "yarl is wrong, urllib is right"---I know that the author of yarl knows RFC 3986 way, way better than I ever will and it could be yarl that is doing the better job of adhering to it.  But I'm curious what motivates the difference in behavior below.

**Summary:** We have an API endpoint where one of the query string params is `url=`.  `aiohttp` (via `yarl.URL`) and `requests` (via `urllib.parse.urlencode`) encode the query string differently.  

```python
>>> from yarl import URL
>>> url = URL("http://api.example.com")
>>> params = {
...     "url": "https://www.katherinetimes.com.au/story/6005621/wife-of-ex-nissan-boss-ghosn-leaves-japan/?src=rss",
...     "api_key": "XXXX",
... }
>>> url.with_query(params)
URL('http://api.example.com/?url=https://www.katherinetimes.com.au/story/6005621/wife-of-ex-nissan-boss-ghosn-leaves-japan/?src%3Drss&api_key=XXXX')
```

Notice that the **?** in the URL (within the query string) is *not* encoded:

```
... japan/?src%3Drss&api ...
```

With the `?` not encoded, the API endpoint gives us back a 400 response.

Now with `urllib.parse`:

```python
>>> parts = urllib.parse.urlparse(url.human_repr())
>>> parts
ParseResult(scheme='http', netloc='api.example.com', path='/', params='', query='', fragment='')
>>> urllib.parse.urlunparse(
...     (
...         parts.scheme,
...         parts.netloc,
...         parts.path,
...         parts.params,
...         urllib.parse.urlencode(params),
...         parts.fragment
...     )
... )
'http://api.example.com/?url=https%3A%2F%2Fwww.katherinetimes.com.au%2Fstory%2F6005621%2Fwife-of-ex-nissan-boss-ghosn-leaves-japan%2F%3Fsrc%3Drss&api_key=XXXX'
```

Here the **?** in the URL (within the query string) is encoded (as are the `/`).

```
... japan%2F%3Fsrc%3Drss&api ...
```

With the `?` encoded, the API endpoint gives us back a 200 response.

So, `urlencode` has a default of `safe=''` and encodes pretty much everything.  Is there a reason that `yarl` seems to diverge from this?

Now, when I look at RFC 3986, it actually seems like the `/` and `?` should *not* need to be encoded:

```
   query         = *( pchar / "/" / "?" )
[...]

PR #1067, opened 2024-08-31
Revert "RFC3986 compatible URL.join honoring empty segments (#1039)"

bdraco:
This reverts commit e3dd736368790d1e4dd1d637eecc598a60216132.

This change introduced a regression with parsing the query string in joined urls. #1066 attempted to fix the regression, but when it became too involved. Since this doesn't look like there will be a quick fix, the change is being reverted so it can be tried again in the future.

### DIFF
--- a/CHANGES/1067.bugfix.rst
+++ b/CHANGES/1067.bugfix.rst
@@ -1,3 +1,3 @@
-Reverted RFC3986 compatible URL.join honoring empty segments
+Reverted RFC3986 compatible URL.join honoring empty segments which was introduced in :issue:`1039`.
 
-This change introduced a regression handling query string parameters with joined URLs.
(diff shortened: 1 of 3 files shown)


bdraco:
We may need to sort out https://github.com/aio-libs/yarl/issues/301 first before we can re-try #1039

CHANGES/1067.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+Reverted RFC3986 compatible URL.join honoring empty segments which was introduced in :issue:`1039`.
```
webknjaz, line 1: @bdraco FYI, Sphinx has a dedicated role `:rfc:` which should be used to link RFCs. The rendered style would also be more consistent.

bdraco: I did discover that, and I changed it before I published it in the release step.  At least I hope I did as my burnout level was pretty high at that point

PR #1068, opened 2024-08-31
Release 1.9.6

bdraco:
~~https://github.com/aio-libs/yarl/actions/runs/10642760388~~ failed. (forgot to [drop `v`](https://yarl.aio-libs.org/en/latest/contributing/release_guide/#pre-release-activities) from CHANGES.rst)
https://github.com/aio-libs/yarl/actions/runs/10643373937

Re-release to revert https://github.com/aio-libs/yarl/pull/1039 since it introduced a regression in query string handling with `%2B` (aka `+`) and url joins.

See https://github.com/aio-libs/yarl/pull/1066 for test case

Screenshot 2024-08-30 at 7 24 06 PM

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,26 @@ Changelog
 
 .. towncrier release notes start
 
+1.9.6
+=====
+
+*(2024-08-30)*
+
+
+Bug fixes
+---------
+
+- Reverted :rfc:`3986` compatible :meth:`URL.join() <yarl.URL.join>` honoring empty segments which was introduced in :issue:`1039`.
+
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -99,6 +119,11 @@ Removals and backward incompatible breaking changes
   *Related issues and pull requests on GitHub:*
   :issue:`1057`.
 
+- Dropped support for Python 3.7 -- by :user:`Dreamsorcerer`.
```
bdraco, line 122: This was missing from 1.9.5 because it couldn't parse `removal`

PR #1070, opened 2024-08-31
Adapt aiohttp reify implementation to replace internal cached_property

### LINKED ISSUE #1065: Use aiohttp reify as a replacement for yarl's custom cached_property

The custom `cached_property` used for `URL` can be replaced with `aiohttp` `reify` which is faster, better tested, and better suited for properties with small amount of code 

n/a

n/a

### Additional context

_No response_

### Code of Conduct

bdraco:
## What do these changes do?

Replaces yarl's cached_property with an adapted version of aiohttp's reify which is faster and has a cython implementation.  We have a lot of small property accesses in yarl that will benefit from the cython implementation.

Screenshot 2024-08-31 at 8 40 30 AM

no, performance improvement

## Related issue number

closes #1065


yarl/_helpers.pyx
```diff
@@ -0,0 +1,35 @@
...
+                val = self.wrapped(inst)
+                inst._cache[self.name] = val
+                return val
+        except AttributeError:
+            if inst is None:
+                return self
+            raise
+
+    def __set__(self, inst, value):
+        raise AttributeError("reified property is read-only")
```
bdraco, line 35:
```suggestion
        raise AttributeError("cached property is read-only")
```

yarl/helpers.py
```diff
@@ -0,0 +1,59 @@
...
+                val = self.wrapped(inst)
+                inst._cache[self.name] = val
+                return val
+        except AttributeError:
+            if inst is None:
+                return self
+            raise
+
+    def __set__(self, inst: _TSelf[_T], value: _T) -> None:
+        raise AttributeError("reified property is read-only")
```
bdraco, line 48:
```suggestion
        raise AttributeError("cached property is read-only")
```

yarl/helpers.py
```diff
@@ -0,0 +1,59 @@
+"""Various helper functions"""
```
bdraco, line 1:
```suggestion
"""Various helper functions."""
```

yarl/_helpers.pyx
```diff
@@ -0,0 +1,35 @@
...
+        self.name = wrapped.__name__
+
+    @property
+    def __doc__(self):
+        return self.wrapped.__doc__
+
+    def __get__(self, inst, owner):
+        try:
+            try:
+                return inst._cache[self.name]
```
bdraco, line 24: In a future PR we might be able to improve this a bit more as it could use the PyDict functions

Screenshot 2024-08-31 at 9 09 25 AM

.github/workflows/ci-cd.yml
```diff
@@ -316,6 +316,7 @@ jobs:
         !cancelled()
       run: >-
         python -Im coverage report --format=markdown
+        --debug=plugin,self,sys,trace,process,pid,lock,multiop
```
bdraco, line 319: remove this before merge

bdraco, line 319:
```suggestion
```

tests/test_helpers.py
```diff
@@ -0,0 +1,91 @@
...
+            def __init__(self):
+                self._cache = {}
+
+            @self.cached_property  # type: ignore[misc]
+            def prop(self):
+                """Docstring."""
+                return 1
+
+        assert isinstance(A.prop, self.cached_property)
+        assert "Docstring." == A.prop.__doc__
```
bdraco, line 36: I copied these from aiohttp but I think the asserts should be reversed

bdraco:
added skip label to workaround the bot not working

yarl/_helpers_c.pyi
```diff
@@ -0,0 +1,6 @@
+from typing import Any
+
+class cached_property:
+    def __init__(self, wrapped: Any) -> None: ...  # pragma: no cover
+    def __get__(self, inst: Any, owner: Any) -> Any: ...  # pragma: no cover
+    def __set__(self, inst: Any, value: Any) -> None: ...  # pragma: no cover
```
bdraco, line 6:
```suggestion
    def __init__(self, wrapped: Any) -> None: ...
    def __get__(self, inst: Any, owner: Any) -> Any: ...
    def __set__(self, inst: Any, value: Any) -> None: ...
```

This looks like the same problem with multi-line test coverage with the reporting and not a real issue

bdraco:
Updated a few production systems. Everything looks great.

bdraco:
Using the C implementation now

Screenshot 2024-08-31 at 5 28 13 PM

bdraco:
Testing more widely now, this one's very straightforward, though as aiohttp has been using nearly the same code for a long time, so not expecting any surprises

bdraco:
py-spy looks good as well

bdraco:
Wider testing looks good as well.

tests/test_helpers.py
webknjaz, line 1: @bdraco not something immediately actionable, but I hope to get rid of class-based tests and refactor a couple of things with weird C-ext vs. pure-python auto-selection.

Here's what I have in mind as a strategy for testing dual interfaces, for a few other projects: https://github.com/aio-libs/multidict/blob/bc86d23/tests/conftest.py#L24-L218.
Just wanted to share...

yarl/_helpers_c.pyx
```diff
@@ -0,0 +1,37 @@
+# cython: language_level=3
+
+cdef class cached_property:
```
webknjaz, line 3: @bdraco how different is this from https://docs.python.org/3.8/library/functools.html#functools.cached_property? It exists in stdlib since Python 3.8.

bdraco: The functools version stores the cache in self.__dict__, and the yarl version stores in self._cache which means it also works with __slots__.  The cython implementation is faster than the pure python implementation in functools and better thought out and optimized for our use case.

Also the cached property implementation in functools had an undocumented threading lock which was removed in 3.12+

webknjaz: Oh, I see. If this is copied across our projects, would it make sense to move it out into a separate lib?

bdraco: Yes it would make sense to break it out if we don't mind having another dep in the projects that use it.

webknjaz: I don't mind personally. Especially, if it's not predicted to change often. Can't wait for us to be able to use ABI 3 (it's in the works in the new Cython) that would reduce the need for re-releases even more.

bdraco: I am working on the new package at https://github.com/bluetooth-devices/propcache reserved the name at https://pypi.org/project/propcache/

I will transfer it over when its ready

bdraco: - [ ] yarl https://github.com/aio-libs/yarl/pull/1169
- [ ] aiohttp https://github.com/aio-libs/aiohttp/pull/9394

webknjaz: Honestly, I don't see a problem in hosting new projects in aio-libs right away.

bdraco: I was going to do that, but then I realized I didn’t have permissions to create it

webknjaz: @bdraco if you go to the repo settings and click `Transfer`, does it let you select the org?

webknjaz: Or I can just do it myself since you gave me the permissions...

webknjaz: Done!

bdraco: Thank you!

webknjaz: @bdraco I've improved your PyPI privileges in the org. You can try moving the project there too.

webknjaz: @bdraco I'm going to file an issue in the repo with some things I'd like us to try out while the repo is fresh.

webknjaz: https://github.com/aio-libs/propcache/issues/18

bdraco: I think it moved over ok but it presented a "You do not have permission to view this page" after I did it

webknjaz: @bdraco that's caused by https://github.com/pypi/warehouse/issues/13558

PR #1071, opened 2024-08-31
Attempt to fix coverage report

bdraco:

PR #1074, opened 2024-09-01
🧪🚑 Fix coverage.py-included paths @ XML

webknjaz:
This patch fixes the directory misattribution of test file paths by marking the entire repository directory as source.

closes #1071

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -16,6 +16,6 @@ cover_pylib = false
 plugins = Cython.Coverage
 relative_files = true
 source =
-  tests
+  .
 source_pkgs =
   yarl


bdraco (approved):
Works for me.  This is less invasive than the alternative https://github.com/aio-libs/yarl/pull/1071

bdraco:
We are still getting misses for multi-line coverage but its much better that we can see whats covered now 👍

webknjaz:
@bdraco multiline? do you mean branches?

bdraco:
Statements that span more than one line.

bdraco:
Ie

```
x = URL(
   Code in here shows as not covered
)
```

bdraco:
https://app.codecov.io/gh/aio-libs/yarl/blob/master/yarl%2F_url.py#L263

https://app.codecov.io/gh/aio-libs/yarl/blob/master/yarl%2F_url.py#L281

https://app.codecov.io/gh/aio-libs/yarl/blob/master/yarl%2F_url.py#L300

bdraco:
https://github.com/nedbat/coveragepy/issues/571

webknjaz:
@bdraco that shows up as covered for me..

webknjaz:
@bdraco oh... I see what you're talking about! That's not related to coverage.py. It excludes those lines (so they aren't marked as covered or uncovered). However, we *also* upload coverage reports generated by MyPy and it explicitly marks the lines you're referring to as uncovered.

The fix should be done in the context of MyPy, not pytest.

webknjaz:
One way to spot this would be toggling the flags on the UI. For example, if you go to Codecov and select a flag only reported from pytest jobs (like `Py-3.11.9`), you can see that coverage is actually 100% thanks to those lines you referred being excluded: https://app.codecov.io/gh/aio-libs/yarl/pull/1074/blob/tests/test_url.py?flags%5B0%5D=Py-3.11.9.

Meanwhile, I've added a `pytest` flag so that it'd be possible to just select `MyPy` or `pytest` and show one type of coverage vs. the other w/o having to guess flags.

PR #1081, opened 2024-09-02
Improve performance of handling ports in the URL

bdraco:
## What do these changes do?

There were a few places were the port got parsed out of the underlying `SplitResult` object multiple times. Make sure they are all cached to avoid doing the same work over and over.

This solves a minor performance regression that was introduced with the port normalization in #1033

performance improvement

## Related issue number

#1033

For reference, `is_default_port` from `update_headers` in `aiohttp` is showing significantly higher on the profile with recent yarl versions.

before
is_default_port

after
update_headers_after

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -15,6 +15,11 @@ comment:
 coverage:
   range: 99.34..100
   status:
+    patch:
+      default:
+        target: 100%
+        flags:
+        - pytest
     project:
       default:
         target: 100%
--- a/CHANGES/1081.misc.rst
+++ b/CHANGES/1081.misc.rst
@@ -1,1 +1,1 @@
-Improved performance of handling ports -- by :user:.
+Improved performance of handling ports -- by :user:`bdraco`.
--- a/CHANGES/1081.misc.rst
+++ b/CHANGES/1081.misc.rst
@@ -1,0 +1,1 @@
+Improved performance of handling ports -- by :user:.
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -1146,7 +1146,7 @@ class URL:
                     user,
                     password,
                     host,
-                    self._val.port,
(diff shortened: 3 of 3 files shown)


yarl/_url.py
```diff
@@ -1142,7 +1148,7 @@ def human_repr(self):
                     user,
                     password,
                     host,
-                    self._val.port,
+                    self.explicit_port,
```
bdraco, line 1151: This line is actually covered but we currently have a problem with reporting multi-line statements

bdraco, line 1151:
```suggestion
                    self._val.port,
```

Going to leave it as-is since CI is blocking merge otherwise

yarl/_url.py
```diff
@@ -134,6 +134,8 @@ class URL:
     _PATH_UNQUOTER = _Unquoter(ignore="/", unsafe="+")
     _QS_UNQUOTER = _Unquoter(qs=True)
 
+    _val: SplitResult
+
```
bdraco, line 138:
```suggestion
```

bdraco:
The line is covered that `codecov/project/tests-pytest` and `codecov/patch` is complaining about

.codecov.yml
```diff
@@ -15,6 +15,11 @@ comment:
 coverage:
   range: 99.34..100
   status:
+    patch:
+      default:
+        target: 100%
+        flags:
+        - pytest
```
bdraco, line 22: Testing this here to get the CI working.. Will move to another PR

CHANGES/1081.misc.rst
```diff
@@ -0,0 +1 @@
+Improved performance of handling ports -- by :user:.
```
bdraco, line 1:
```suggestion
Improved performance of handling ports -- by :user:`bdraco`.
```

PR #1082, opened 2024-09-02
 RFC3986 compatible URL.join honoring empty segments again

commonism:
This is a revised version of #1039 using encoded arguments to work around issues with + in query segments.
It uses encoded=True for the parts when building the URL and normalizes the path itself.

## What do these changes do?

The changes re-implement URL.join within yarl.URL instead of using urllib.parse.urljoin to avoid known issues when joining paths with empty segments (https://github.com/python/cpython/issues/84774)

Due to the unit tests for URL.join it was required to align the query parameter rendering to match the unit tests, empty parameters are rendered without value therefore instead of an empty value.

## Related issue number

#1066

## Checklist


yarl/_url.py
```diff
@@ -1123,7 +1126,59 @@ def join(self, url):
         # See docs for urllib.parse.urljoin
         if not isinstance(url, URL):
             raise TypeError("url should be URL")
-        return URL(urljoin(str(self), str(url)), encoded=True)
+        other: URL = url
+        scheme = other.scheme or self.scheme
+        parts = {
+            k[len("raw_") :]: getattr(self, k) or ""
+            for k in ("raw_authority", "raw_path", "raw_query_string", "raw_fragment")
```
bdraco, line 1133: Instead of slicing up `k` it might be a bit faster to unpack a tuple of the known value

bdraco:
Thanks for following up 👍

commonism:
rebased

yarl/_url.py
commonism: You are right, the compiler does not bother to optimize.

```python
>>> def x():
...     return "ab"[len("a"):]
... 
>>> import dis
>>> dis.dis(x)
  2           0 LOAD_CONST               1 ('ab')
              2 LOAD_GLOBAL              0 (len)
              4 LOAD_CONST               2 ('a')
              6 CALL_FUNCTION            1
              8 LOAD_CONST               0 (None)
             10 BUILD_SLICE              2
             12 BINARY_SUBSCR
             14 RETURN_VALUE
```

yarl/_url.py
```diff
@@ -1116,7 +1122,59 @@ def join(self, url):
...
+                {
+                    k[4:]: getattr(other, k) or ""
+                    for k in (
+                        "raw_authority",
+                        "raw_path",
+                        "raw_query_string",
+                        "raw_fragment",
+                    )
+                }
+            )
```
bdraco, line 1148: This is getting rather long, it probably more readable to write it out in both places

yarl/_url.py
```diff
@@ -1116,7 +1122,59 @@ def join(self, url):
...
+        scheme = other.scheme or self.scheme
+        parts = {
+            k[4:]: getattr(self, k) or ""
+            for k in ("raw_authority", "raw_path", "raw_query_string", "raw_fragment")
+        }
+        parts["scheme"] = scheme
+
+        if scheme != self.scheme or scheme not in uses_relative:
+            return URL(str(other))
+
```
bdraco, line 1135: Can be moved above building `parts`

yarl/_url.py
```diff
@@ -1116,7 +1122,59 @@ def join(self, url):
...
+        other: URL = url
+        scheme = other.scheme or self.scheme
+        parts = {
+            k[4:]: getattr(self, k) or ""
+            for k in ("raw_authority", "raw_path", "raw_query_string", "raw_fragment")
+        }
+        parts["scheme"] = scheme
+
+        if scheme != self.scheme or scheme not in uses_relative:
+            return URL(str(other))
```
bdraco, line 1134:
```suggestion
            return URL(other._val._replace(), encoded=True)
```

something like this should work to copy without having to go through `__str__`

commonism: yes

yarl/_url.py
commonism: I propose a function instead

yarl/_url.py
```diff
@@ -1124,28 +1124,25 @@ def join(self, url):
...
+            return URL(other._val._replace(), encoded=True)
+
+        def raw(uo: URL):
+            attrs = (
+                "raw_authority",
+                "raw_path",
+                "raw_query_string",
+                "raw_fragment",
+            )
+            return {k[4:]: getattr(uo, k) or "" for k in attrs}
```
bdraco, line 1138:
```diff
diff --git a/yarl/_url.py b/yarl/_url.py
index fb75339..7e4cfeb 100644
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -1107,6 +1107,14 @@ class URL:
             name = name[: -len(old_suffix)] + suffix
         return self.with_name(name)
 
+    def _raw_dict(self):
+        return {
+            "authority": self.raw_authority,
+            "path": self.raw_path,
+            "query_string": self.raw_query_string,
+            "fragment": self.raw_fragment
+        }
+
     def join(self, url):
         """Join URLs
 
@@ -1128,21 +1136,12 @@ class URL:
         if scheme != self.scheme or scheme not in uses_relative:
             return URL(other._val._replace(), encoded=True)
 
-        def raw(uo: URL):
-            attrs = (
-                "raw_authority",
-                "raw_path",
-                "raw_query_string",
-                "raw_fragment",
-            )
-            return {k[4:]: getattr(uo, k) or "" for k in attrs}
-
-        parts = raw(self)
+        parts = self._raw_dict()
         parts["scheme"] = scheme
 
         # scheme is in uses_authority as uses_authority is a superset of uses_relative
         if scheme in uses_authority and other.authority:
-            parts.update(raw(other))
+            parts.update(other._raw_dict())
             return URL.build(**parts, encoded=True)
 
         if other.path or other.fragment:
```

It be a lot more readable (and probably a bit faster) if you wrote it something like this.

webknjaz: @bdraco I'd make it a property, then.

webknjaz: Plus a docstring and type annotations..

bdraco: Looks like we are missing a lot of types internally .. but we have them in the pyi.... probably nicer to clean that up so the code tools give us better visiblity.

I'll do that in another PR

yarl/_url.py
```diff
@@ -427,7 +431,14 @@ def _default_port(self) -> Union[int, None]:
...
             return None
-        return DEFAULT_PORTS.get(scheme)
+        return self._default_port(self.scheme)
+
+    @staticmethod
+    def _default_port(scheme: str) -> Union[int, None]:
+        with suppress(KeyError):
+            return DEFAULT_PORTS[scheme]
+
+        return None
```
bdraco, line 441:
```suggestion
        return DEFAULT_PORTS.get(scheme)

    @staticmethod
    def _default_port(scheme: str) -> Union[int, None]:
        return DEFAULT_PORTS.get(scheme)
```

yarl/_url.py
```diff
@@ -228,12 +230,14 @@ def build(
...
                 tmp = SplitResult("", authority, "", "", "")
+                port = None if tmp.port == cls._default_port(scheme) else tmp.port
                 netloc = cls._make_netloc(
-                    tmp.username, tmp.password, tmp.hostname, tmp.port, encode=True
+                    tmp.username, tmp.password, tmp.hostname, port, encode=True
                 )
         elif not user and not password and not host and not port:
             netloc = ""
         else:
+            port = None if port == cls._default_port(scheme) else port
```
bdraco, line 240:
```suggestion
            port = None if port == DEFAULT_PORTS.get(scheme) else port
```

yarl/_url.py
```diff
@@ -429,6 +433,10 @@ def _default_port(self) -> Union[int, None]:
             return None
         return DEFAULT_PORTS.get(scheme)
 
+    @staticmethod
+    def _default_port(scheme: str) -> Union[int, None]:
+        return DEFAULT_PORTS.get(scheme)
+
```
bdraco, line 439:
```suggestion
```

bdraco:
I think I have fixed the conflicts correctly, please double check

yarl/_url.py
```diff
@@ -1123,7 +1132,47 @@ def join(self, url):
...
+                URL(self.path).joinpath(other.path).path
+            )
+        else:
+            # …
+            # and relativizing ".."
+            path = URL("/".join([*self.parts[1:-1], ""])).joinpath(other.path).path
+            if parts["authority"]:
+                parts["path"] = "/" + path
+            else:
+                parts["path"] = path
```
bdraco, line 1173: Can we call `_make_child` here instead of creating another URL object?

yarl/_url.py
```diff
@@ -1160,9 +1160,7 @@ def join(self, url):
 
         if self.path[-1] == "/":
             # using an intermediate to avoid URL.joinpath dropping query & fragment
-            parts["path"] = self._normalize_path(
-                URL(self.path).joinpath(other.path).path
-            )
+            parts["path"] = self._normalize_path(self._make_child([other.path]).path)
         else:
             # …
             # and relativizing ".."
```
bdraco, line 1166: Can we use make child for this branch as well?

tests/test_url.py
```diff
@@ -1664,6 +1664,66 @@ def test_join_from_rfc_3986_abnormal(url, expected):
...
+    # tests from cpython urljoin
+    base = URL(base)
+    url = URL(url)
+    expected = URL(expected)
+    joined = base.join(url)
+    assert joined == expected
+
+
+def test_empty_authority():
+    assert URL("http:///").authority == ""
```
bdraco, line 1724: It looks like I accidentally fixed this when I added typing

yarl/_url.py
```diff
@@ -1199,7 +1218,45 @@ def join(self, url: "URL") -> "URL":
         # See docs for urllib.parse.urljoin
         if not isinstance(url, URL):
             raise TypeError("url should be URL")
-        return URL(urljoin(str(self), str(url)), encoded=True)
+        other: URL = url
```
bdraco, line 1221: Note to self: do performance testing with this change

bdraco: I'm preparing an RC release to make sure the added typing doesn't cause any problems.  Once thats done I'll try to do some production testing with this tonight if I don't run out of waking hours.

bdraco:
join after the change
join_after

bdraco:
before

join_before

bdraco:
I think we can get rid of the function and optimize it a bit more

https://github.com/aio-libs/yarl/commit/f4d4d339c6f2ab73d9cb6cadc02cf2c3bfd4c481

bdraco:
https://github.com/commonism/yarl/pull/1 should optimize it quite a bit more.  The tests pass, but maybe we need more coverage

yarl/_url.py
bdraco, line 1221:
```suggestion
        other = url
```

hint should be unneeded as its inferred from the isinstance

yarl/_url.py
commonism: I was wrong here.
But as I just saw you've figured this out yourself already.

commonism:
I've merged your optimizations so we can proceed working them here.

I've added tests which fail due to the folding of path.
https://github.com/aio-libs/yarl/pull/1082/commits/cb8a13c06aa3f93969ed667d9b09451cb640e23c

bdraco:
Pushed a fix for the new test cases. 

If there are any other cases to test, that you can think of that might break it, it would be great to add them.  I think we are a bit light on tests for joining encoded paths.

 I've got to get some sleep now.

commonism:
> Pushed a fix for the new test cases.

possible without make_child

bdraco:
> > Pushed a fix for the new test cases.
> 
> possible without make_child

Nice!

Now I really have to get some sleep. Cheers

CHANGES/1082.bugfix.rst
```diff
@@ -0,0 +1,27 @@
...
+
+
+Empty segments are honored in the base as well as the joined part.
+
+.. code-block:: pycon
+
+   >>> URL("https://web.archive.org/web/https://").join(URL("github.com/aio-libs/yarl"))
+   URL('https://web.archive.org/web/https://github.com/aio-libs/yarl')
+
+
```
bdraco, line 25: We should add a comment that this change first appeared in release X, and this is a rework of it

tests/test_url.py
```diff
@@ -1664,6 +1664,82 @@ def test_join_from_rfc_3986_abnormal(url, expected):
...
+
+
+OPTIMIZE_URLJOIN = [
+    ("a/", "b", "a/b"),
+    ("a/b", "c", "a/c"),
+]
+
+
+@pytest.mark.parametrize("base,url,expected", OPTIMIZE_URLJOIN)
+def test_join_optimize_urljoin(base, url, expected):
```
bdraco, line 1730: We should rename this block as we won’t know what optimize means in this context 5 years from now

bdraco:
I think this is ready to go. So small nits above, and maybe there’s more coverage to add?

bdraco:
I added some more coverage and cleaned up the message a bit

bdraco (approved):

bdraco:
I should have time to do another release for this fix later today

bdraco:
Still doing testing with this, but I think its good to go if the CI passes and the performance testing looks ok

bdraco (approved):

bdraco:
Performance testing looks good.

bdraco:
Thanks for sticking with this one @commonism

commonism:
It was not easy, but a pleasure.

PR #1083, opened 2024-09-02
Fix codecov blocking merges

bdraco:
## What do these changes do?

- prevent mypy coverage from being considered in the `patch` coverage since it decreases the coverage and blocks merges
- codecov is rounding down sometimes for `tests-pytest` so drop it to .01

no CI fix

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -15,6 +15,11 @@ comment:
 coverage:
   range: 99.34..100
   status:
+    patch:
+      default:
+        target: 100%
+        flags:
+        - pytest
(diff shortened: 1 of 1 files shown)


webknjaz:
@bdraco this shouldn't be necessary FWIW. Sometimes, that last 0.01% of coverage isn't added until more jobs send in their coverage reports. I saw this happen on master when I updated the config. It was 0.01% lower than the threshold, then I waited for another minute to let Codecov finish processing the reports, and it matched the expectation..

PR #1084, opened 2024-09-02
Add typing to URL

bdraco:
## What do these changes do?

Add typing to URL

no runtime changes


CHANGES/1084.contrib.rst
```diff
@@ -0,0 +1 @@
+Add typing to the ``URL`` object -- by :user:`bdraco`.
```
webknjaz, line 1:
```suggestion
Covered the :class:`~yarl.URL` object with types -- by :user:`bdraco`.
```

bdraco: Thanks

CHANGES/1084.contrib.rst
webknjaz, line 1: Not sure if it's just a `contrib`. Having types is like a public API for dev experience, for the end-users of yarl, I think. Eh?

Perhaps, it needs to be renamed or symlinked to another category too.

bdraco: I made it a feature as well

webknjaz: Thanks! Marking the thread as resolved.

yarl/_url.py
```diff
@@ -17,16 +30,27 @@
...
+_SimpleQuery = Union[str, int, float]
+_QueryVariable = Union[_SimpleQuery, "Sequence[_SimpleQuery]"]
+_Query = Union[
+    None, str, "Mapping[str, _QueryVariable]", "Sequence[Tuple[str, _QueryVariable]]"
+]
+
+if sys.version_info >= (3, 11):
+    from typing import Self
+else:
+    Self = Any
```
webknjaz, line 42: I think it's common to rely on [`typing-extensions`](https://typing-extensions.rtfd.io/en/latest/#typing_extensions.Self) in such cases… It'd need to go into `setup.cfg` with an env marker.

bdraco: There has been an effort to remove typing-extensions https://github.com/aio-libs/aiohttp/pull/8352

.coveragerc
```diff
@@ -9,6 +9,11 @@ source =
 skip_covered = true
 skip_empty = true
 show_missing = true
+exclude_also =
+    if TYPE_CHECKING
+    assert False
+    : \.\.\.(\s*#.*)?$
```
webknjaz, line 15: @bdraco Are these tripping coverage.py and not MyPy? Could you add code comments to justify each exclusion?

.coveragerc
```diff
@@ -9,6 +9,11 @@ source =
 skip_covered = true
 skip_empty = true
 show_missing = true
+exclude_also =
+    if TYPE_CHECKING
+    assert False
```
webknjaz, line 14: Where have you seen this one?

bdraco: I copied the block from `aiohttp`.  We don't need it here.  I'll clean it up once the CI passes

.coveragerc
```diff
@@ -9,6 +9,11 @@ source =
 skip_covered = true
 skip_empty = true
 show_missing = true
+exclude_also =
```
webknjaz, line 12: It looks like `covdefaults` handles all of these even better. I've been meaning to integrate it for a while...

yarl/_url.py
webknjaz: I think that's because it was no longer used. I don't know why we'd want to ban it for good. Perhaps @Dreamsorcerer could explain..

.coveragerc
webknjaz: I'll test it with #1085 and get in, once I know it doesn't break the current setup.

.coveragerc
webknjaz: I'm replacing it in #1085. You can just drop it then.

.coveragerc
Dreamsorcerer: Personally, I prefer the control here. There's atleast one match in covdefaults that I don't want enabled. Seems like overkill for something that can be done with 3 simple regexes.

yarl/_url.py
Dreamsorcerer: I think it's a good idea to avoid reintroducing it as a dependency, many users prefer fewer dependencies.
Typing in Python has reached a point where it's not really needed for normal use cases, newer versions of Python have small improvements, but nothing that really makes a difference to the end user.

Note this can also be done with a TypeVar, but given that it's basically only the oldest 2 versions of Python that would see the Any, most users will end up type checking with mypy against a more recent release (even if runtime might be on an older release).

Also, are we intending to ditch the .pyi files? If not, then none of this is even exposed to the users...

bdraco: > Also, are we intending to ditch the .pyi files? If not, then none of this is even exposed to the users...

I had originally marked this as `contrib` only because of that, but it probably make more sense to ditch the `pyi` instead of trying to keep them in sync

.coveragerc
Dreamsorcerer: Reviewing it again, I think there's multiple rules I don't want from covdefaults, and all the others we don't already use, we probably will never need.

But, this rule looks like a really bad idea to me:
`    -> ['"]?NoReturn['"]?:`
I'm pretty sure that disables coverage of entire functions just because they are marked as not returning.
Just because they don't return, doesn't mean they shouldn't be run. We have dozens of such functions across our codebases and they should not be ignored.

yarl/_url.py
webknjaz: My understanding is that the pyi files are needed for C-extensions. Is Cython capable of producing the proper annotations for MyPy to consume?

.coveragerc
webknjaz: @Dreamsorcerer that's an interesting corner case, I'll compare the coverage in that PR and see if it affects anything...

yarl/_url.py
bdraco: We need the pyi for the C-extensions, but it looks like everything that in the `__init__.pyi` `__all__` is pure python unless I've overlooked something

webknjaz: Agreed.

.coveragerc
```diff
@@ -14,6 +14,9 @@ skip_covered = true
 skip_empty = true
 show_missing = true
 exclude_also =
+  if TYPE_CHECKING    # never executes at runtime
```
webknjaz, line 17: I'm not sure if coverage.py handles inline comments following regexes that can contain `#` in them, too. Have you checked?

yarl/_url.py
bdraco: We do need to pull in the typed dict for the cache here as well.. I'll do that after #1086 and remove the `__init__.pyi` so we can rely on the inline types instead

.coveragerc
webknjaz: FYI, I ended up merging it, will re-evaluate later if needed. Marking this thread as resolved.

.coveragerc
```diff
@@ -14,6 +14,9 @@ skip_covered = true
...
+  : \.\.\.(\s*#.*)?$  # coverage.py will never report ... as covered
+  ^ +\.\.\.$          # coverage.py will never report ... as covered
```
webknjaz, line 19: let's drop these for now
```suggestion
```

yarl/_helpers_c.pyi
```diff
@@ -1,6 +0,0 @@
-from typing import Any
```
bdraco, line 1: No longer needed since `_helpers.py` will always use `_helpers_py` for typing which it better typed with a Generic

.coveragerc
```diff
@@ -10,7 +10,7 @@ source =
   */Lib/site-packages/yarl
 
 [report]
-fail_under = 98.97
+fail_under = 98.96
```
bdraco, line 13: 2024-09-03T05:38:12.7709570Z [31m[1mFAIL Required test coverage of 98.96% not reached. Total coverage: 98.96%

Well thats a bit of an unexpected error

bdraco, line 13:
```suggestion
fail_under = 98.95
```

bdraco: This fails before coverage upload so its impossible to see whats missing on python 3.8 without installing it locally

bdraco:
path shows 100% diff hit on codecov but is reporting at 99.13% here...

https://app.codecov.io/gh/aio-libs/yarl/pull/1084?flags%5B0%5D=pytest&dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr%20comments&utm_term=aio-libs

.coveragerc
webknjaz: @bdraco yes, `covdefaults` sets this to 100% but since not every job we have runs all the tests, it's impossible to have this value. I lowered it per what I saw in the CI while integrating. It's okay to lower it further when needed, since Codecov looks at the combined numbers anyway. It might be possible to pass more granular values via CLI args to pytest, but I haven't experimented with that, it's probably not worth it.

webknjaz:
> path shows 100% diff hit on codecov but is reporting at 99.13% here...

Yeah, I suppose this was due to the proportion being different because the overall number of lines changed. I'm not seeing 99.13% there, though. Perhaps, it processed more reports after your comment.
For me, it shows 94.94% for the entire `pytest` flag and 99.88% for `tests/` flagged with `pytest` and 98.15% for `yarl/`+`pytest`.

I suppose, it's okay to raise the `codecov/project/lib` expectation to 98.1% (leaving that 0.05% margin for future line count changes, until we can go for 100%).

yarl/_helpers_c.pyi
webknjaz, line 1: Not sure if this one can be removed: direct imports of `yarl/_helpers_c` won't handle this properly. Especially, once the tests are restructured to deal with the dual-variant testing better (similar to frozenlist).

yarl/_url.py
```diff
@@ -186,18 +229,18 @@ def __new__(cls, val="", *, encoded=False, strict=None):
...
-        port=None,
-        path="",
-        query=None,
-        query_string="",
-        fragment="",
-        encoded=False,
-    ):
+        scheme: str = "",
+        authority: str = "",
+        user: Optional[str] = None,
```
webknjaz, line 234: Please, avoid using `Optional` in favor of `| None` since it's considered less confusing:
```suggestion
        user: str | None = None,
```

Dreamsorcerer: But, as mentioned before, this breaks runtime type checking on older Python versions, if the user tries to use those types. We can leave this and just let pyupgrade change next year.

yarl/_url.py
```diff
@@ -276,14 +319,14 @@ def __str__(self):
...
+    def __repr__(self) -> str:
         return f"{self.__class__.__name__}('{str(self)}')"
 
-    def __bytes__(self):
+    def __bytes__(self) -> bytes:
         return str(self).encode("ascii")
 
-    def __eq__(self, other):
-        if not type(other) is URL:
+    def __eq__(self, other: Any) -> bool:
```
Dreamsorcerer, line 328: This should be type safe with any object.

```suggestion
    def __eq__(self, other: object) -> bool:
```

Dreamsorcerer: Same for the other comparison methods.

yarl/_url.py
bdraco: Yup, we have to keep it for 3.8   Once we drop 3.8, we can use `from __future__ import annotations` to turn it on in 3.9 and than run pyupgrade

yarl/_helpers_c.pyi
bdraco: I'll leave it for now, we can always remove it later

yarl/_url.py
bdraco: This should be fine to use the `type()` check as its faster and subclassing is forbidden https://github.com/aio-libs/yarl/blob/d4e76015f31d8133ba37d1d2e1a09fbbeb05538f/tests/test_url.py#L18 and blocked in `__init_subclass__`

yarl/_url.py
```diff
@@ -1205,21 +1275,22 @@ def cache_clear() -> None:
...
-def cache_info():
+def cache_info() -> CacheInfo:
     """Report cache statistics."""
-    return {
+    cache_info: CacheInfo = {
         "idna_encode": _idna_encode.cache_info(),
         "idna_decode": _idna_decode.cache_info(),
         "ip_address": _ip_compressed_version.cache_info(),
     }
+    return cache_info
```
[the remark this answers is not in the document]
bdraco: removed in 915ab95bf8245aa8ab5975ca297602350b80e32a

yarl/_url.py
Dreamsorcerer: Still a runtime error on 3.9 if you use get_type_hints() or similar..

bdraco:
I'm going to do a RC0 release so I can run this in the CI of a few projects since typing changes are always hard to predict the impact

yarl/_url.py
webknjaz: @Dreamsorcerer don't we already do this elsewhere? Though, if `|` is not an option, a compatible variant would be `Union[str, None]`.

bdraco: #1095 Replace all `Optional[str]` with `Union[str, None]`

sergeniushpe:
Hello, I think this change broke python 3.8 support:

```
../lib/python3.8/site-packages/yarl/_url.py:606: in URL
    def query(self) -> MultiDictProxy[str]:
E   TypeError: 'type' object is not subscriptable
```
If I downgrade to release 1.9.7 it works again

bdraco:
> Hello, I think this change broke python 3.8 support:
> 
> ```
> ../lib/python3.8/site-packages/yarl/_url.py:606: in URL
>     def query(self) -> MultiDictProxy[str]:
> E   TypeError: 'type' object is not subscriptable
> ```
> 
> If I downgrade to release 1.9.7 it works again.
> 
> You might want to update python requirements to indicate python>=3.9 is required

Would you please open an issue report, I will fix it shortly

bdraco:
Please include the multi dict version you are using in the issue

Dreamsorcerer:
> Please include the multi dict version you are using in the issue

As far as I can see, multidict has supported that syntax for ~6 years...
But, maybe we should increase the minimum version of multidict that yarl supports.

bdraco:
I had to take the python 3.8 install back to multidict 4 to get it to fail.  There might be a newer version that fails though as I stopped trying them once I confirmed the issue.  Also even with 4 it didn’t fail on python 3.9 so I suspect the issue will go away on its own soon when we drop 3.8

yarl/_url.py
bdraco: I think we shouldn't change `aiohttp`. This one is has `Union` and `Optional` types quite a bit so we should go one way or the other for consistency, and since @webknjaz has a strong preference for `Union` I'm fine with that here knowing that they will eventually be the `type | None` syntax I personally much prefer.

PR #1085, opened 2024-09-02
🧪 Integrate `covdefaults` into coverage.py

webknjaz:
## What do these changes do?

$sbj.

Nope.

## Related issue number

https://github.com/aio-libs/yarl/pull/1084#discussion_r1741257263

## Checklist

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -22,7 +22,9 @@ cover_pylib = false
 # https://coverage.rtfd.io/en/latest/contexts.html#dynamic-contexts
 # dynamic_context = test_function  # conflicts with `pytest-cov` if set here
 parallel = true
-plugins = Cython.Coverage
+plugins =
+  covdefaults
+  Cython.Coverage
 relative_files = true
 source =
   .
--- a/.coveragerc
+++ b/.coveragerc
@@ -10,6 +10,7 @@ source =
   */Lib/site-packages/yarl
 
 [report]
+fail_under = 98.97
 skip_covered = true
 skip_empty = true
 show_missing = true
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -28,7 +28,7 @@ coverage:
         - pytest
         paths:
         - yarl/
-        target: 97.74%  # 97.83%
+        target: 97.91%
       tests:
         flags:
         - pytest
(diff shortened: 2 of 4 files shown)


.coveragerc
```diff
@@ -22,7 +22,9 @@ cover_pylib = false
 # https://coverage.rtfd.io/en/latest/contexts.html#dynamic-contexts
 # dynamic_context = test_function  # conflicts with `pytest-cov` if set here
 parallel = true
-plugins = Cython.Coverage
+plugins =
+  covdefaults
```
Dreamsorcerer, line 26: See https://github.com/aio-libs/yarl/pull/1084#discussion_r1741265935

webknjaz: @Dreamsorcerer we don't have a `NoReturn` in this project. And I checked that the coverage change is meaningful. So I think it's fine to have this plugin enabled in at least this repo.

webknjaz: But you're right that it'd exclude the entire functions, which is harmful where `NoReturn` is used. I filed an issue upstream @ https://github.com/asottile/covdefaults/issues/152 to see if this can be fixed.

webknjaz: Yeah, Anthony can be like that. But there's a chance he'd accept it if somebody else did the work.

.coveragerc
```diff
@@ -22,7 +23,9 @@ cover_pylib = false
 # https://coverage.rtfd.io/en/latest/contexts.html#dynamic-contexts
 # dynamic_context = test_function  # conflicts with `pytest-cov` if set here
 parallel = true
-plugins = Cython.Coverage
+plugins =
```
webknjaz, line 26: So the PyPy jobs have `sed -i.bak 's/^plugins = .*//g' .coveragerc` which was fine when the (incompatible) `Cython.Coverage` plugin was on the same line. However now that it's on a dedicated line, it just breaks the config.

.coveragerc
webknjaz: Given the above, I'm going to integrate the plugin here and maybe other libraries, but not aiohttp for now. I may try to get some fixes upstream and re-evaluate later.

Marking the thread as resolved.

PR #1086, opened 2024-09-02
Cache IP Address parsing

bdraco:
## What do these changes do?

In `_encode_host`, the most expensive thing is converting the string to an ip address to normalize it when the URL has an ip address instead of a hostname. Its much more expensive than [`_idna_encode` which we already cache](https://github.com/aio-libs/yarl/pull/476).

I always wondered the aiohttp benchmarking we did for Home Assistant performed worse when using 127.0.0.1 vs a hostname as it was confusing that that the dns lookup it would be faster, but its cached and avoids the ip address object creation for every iteration in yarl.  I didn't look deep enough in the stack to look at yarl when we did that.  

In hindsight, it should have been a bit more obvious since IP addresses are more common than names for Home Assistant because there is so much local network traffic and ips are usually discovered via mdns or some other method.

Screenshot 2024-09-02 at 1 16 25 PM

no

before
_encode_host

after
encode_host_after


yarl/_url.py
```diff
@@ -769,25 +769,27 @@ def _normalize_path(cls, path):
...
     @classmethod
-    def _encode_host(cls, host, human=False):
-        try:
-            ip, sep, zone = host.partition("%")
-            ip = ip_address(ip)
-        except ValueError:
+    def _encode_host(cls, host: str, human: bool = False) -> str:
+        raw_ip, sep, zone = host.partition("%")
+        # IP parsing is slow, so its wrapped in an LRU
+        ip_compressed_version = _ip_compressed_version(raw_ip)
```
bdraco, line 775: A better design would be to only store hits since the ip address parse code is quick to reject

bdraco:
Ran for a few hours, manual testing and LRU hit rate look good. Working as expected

PR #1094, opened 2024-09-04
Preload netloc properties into the cache if they are known

bdraco:
## What do these changes do?

Preload netloc properties into the cache if they are known

no

## Related issue number

n/a


bdraco:
need to do some more real world testing with this to see if its worth it, or its better to keep it lazy

bdraco:
won't work because we need to requote and we don't have that so we have to parse anyways

PR #1095, opened 2024-09-04
Replace all `Optional` with `Union`

bdraco:
## What do these changes do?

Replace all `Optional` with `Union`

We have both in this project: `Optional[int]` on `port` and `Union[int, None]` on `explict_port` is not so great.  This is a short lived change since `pyupgrade` is going to change them all to `type | None` once we remove support for older python.

no

## Related issue number

https://github.com/aio-libs/yarl/pull/1084#discussion_r1743316773
https://github.com/aio-libs/yarl/pull/1112#discussion_r1746859083


bdraco:
Not sure what kind of changelog message this should get.

Also near midnight here so I'll have to pick this up tomorrow.

webknjaz:
> Not sure what kind of changelog message this should get.

I'd say “contrib” since it's improving the readability for contributors w/o any influence on how the type checkers would perceive this.

yarl/_url.py
webknjaz, line 1: Time to stop importing `Optional`?

CHANGES/1095.contrib.rst
```diff
@@ -0,0 +1 @@
+Replace all :type:`Optional` with :type:`Union` -- by :user:`bdraco`.
```
webknjaz, line 1:
```suggestion
Replaced all :type:`Optional` with :type:`Union` -- by :user:`bdraco`.
```

Oh, I didn't realize you can link types like that..

webknjaz (approved):

CHANGES/1095.contrib.rst
bdraco: It was an attempt but actually I don't think you can.

```
/home/docs/checkouts/readthedocs.org/user_builds/yarl/checkouts/1095/docs/changes.rst:10: WARNING: py:type reference target not found: Optional [ref.type]
/home/docs/checkouts/readthedocs.org/user_builds/yarl/checkouts/1095/docs/changes.rst:10: WARNING: py:type reference target not found: Union [ref.type]
```

CHANGES/1095.contrib.rst
```diff
@@ -0,0 +1 @@
+Replaced all :type:`Optional` with :type:`Union` -- by :user:`bdraco`.
```
bdraco, line 1:
```suggestion
Replaced all ``Optional`` with ``Union`` -- by :user:`bdraco`.
```

was worth a shot but didn't actually work

CHANGES/1095.contrib.rst
webknjaz: Oh, okay. That's what I assumed :)

You'd need to declare the corresponding role. You need to go for 
```rst
:class:`~typing.Optional`
```

In general, the best way to know if an object exists is to call intersphinx from command line or view as html on my helper static site..

webknjaz: @bdraco missed this comment with the working syntax?

bdraco: Yup. Just saw it now

bdraco: https://github.com/aio-libs/yarl/actions/runs/10746477913 is still going.  If it gets past the wheel combine this time, I'll cancel it and rework the changelog message and start the release process over

bdraco: https://github.com/aio-libs/yarl/pull/1121

PR #1100, opened 2024-09-04
Introduce absolute cached property

bdraco:
## What do these changes do?

We call `is_absolute` quite often internally and externally. Currently this requires re-parsing the `netloc`. To make the API more consistent and perform better, `absolute` is now preferred over `is_absolute`. `is_absolute` is retained for backwards compatibility.

`absolute` is now preferred over `is_absolute`, however there are no plans to remove `is_absolute`.

Screenshot 2024-09-04 at 12 50 53 PM

is_absolute

Probably not the best representation of a microbenchmark since `raw_host` would previous be a cache miss most of the time and this change will perform much better either way:
```
>>> timeit.timeit("x.is_absolute()", globals=locals())
0.0817429160233587
>>> timeit.timeit("x.absolute", globals=locals())
0.06663041608408093
```

On previous version when `raw_host` is a cache miss
```
>>> from yarl import URL
>>> x=URL("")
>>> timeit.timeit("x.is_absolute()", globals=locals())
0.18370608310215175
```

### DIFF
--- a/CHANGES/1100.feature.rst
+++ b/CHANGES/1100.feature.rst
@@ -1,1 +1,1 @@
-Added :attr:`~yarl.URL.absolute` which is now preferred over ``is_absolute`` -- by :user:`bdraco`.
+Added :attr:`~yarl.URL.absolute` which is now preferred over ``URL.is_absolute()`` -- by :user:`bdraco`.
--- a/CHANGES/1100.feature.rst
+++ b/CHANGES/1100.feature.rst
@@ -1,1 +1,1 @@
-Added :meth:`~yarl.URL.absolute` which is now preferred over ``is_absolute`` -- by :user:`bdraco`.
+Added :attr:`~yarl.URL.absolute` which is now preferred over ``is_absolute`` -- by :user:`bdraco`.
--- a/CHANGES/1100.feature.rst
+++ b/CHANGES/1100.feature.rst
@@ -1,0 +1,1 @@
+Added :meth:`~yarl.URL.absolute` which is now preferred over ``is_absolute`` -- by :user:`bdraco`.
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -470,7 +470,7 @@ The module supports both absolute and relative URLs.
 Absolute URL should start from either *scheme* or ``'//'``.
 
 
-.. method:: URL.is_absolute()
+.. attribute:: URL.absolute
 
     A check for absolute URLs.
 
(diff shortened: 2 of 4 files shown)


CHANGES/1100.feature.rst
```diff
@@ -0,0 +1 @@
+Added :meth:`~yarl.URL.absolute` which is now preferred over ``is_absolute`` -- by :user:`bdraco`.
```
bdraco, line 1:
```suggestion
Added :attr:`~yarl.URL.absolute` which is now preferred over ``is_absolute`` -- by :user:`bdraco`.
```

docs/api.rst
```diff
@@ -479,15 +479,19 @@ Absolute URL should start from either *scheme* or ``'//'``.
...
-      >>> URL('/path/to').is_absolute()
+      >>> URL('/path/to').absolute
       False
-      >>> URL('path').is_absolute()
+      >>> URL('path').absolute
       False
 
+   .. versionchanged:: 1.9.10
+
+      The ``absolute`` property is preferred over the ``is_absolute`` method.
```
bdraco, line 493:
```suggestion
      The :attr:`~yarl.URL.absolute` property is preferred over the ``is_absolute()`` method.
```

CHANGES/1100.feature.rst
```diff
@@ -0,0 +1 @@
+Added :attr:`~yarl.URL.absolute` which is now preferred over ``is_absolute`` -- by :user:`bdraco`.
```
bdraco, line 1:
```suggestion
Added :attr:`~yarl.URL.absolute` which is now preferred over ``URL.is_absolute()`` -- by :user:`bdraco`.
```

bdraco:
Getting closer to the point where `_url.py` is below `client_reqrep.py` in the profile

tests/test_url.py
```diff
@@ -1307,26 +1307,31 @@ def test_with_suffix_replace():
 def test_is_absolute_for_relative_url():
     url = URL("/path/to")
     assert not url.is_absolute()
+    assert not url.absolute
```
webknjaz, line 1310: @bdraco cached things are usually tested through their `__wrapped__` attributes to prevent caching from influencing behavior in tests. Have you considered doing something like this?

bdraco: Since the URL object is immutable and the state is already predefined we shouldn’t have much risk of the cached value being calculated incorrectly.

Was there a specific case you are concerned about?

webknjaz: No, just got surprised seeing this and thought I'd mention it.

PR #1107, opened 2024-09-05
Fix `TypeError` with `MultiDictProxy` and Python 3.8

### LINKED ISSUE #1105: Hello, I think this change broke python 3.8 support:
Hello, I think this change broke python 3.8 support:

```
../lib/python3.8/site-packages/yarl/_url.py:606: in URL
    def query(self) -> MultiDictProxy[str]:
E   TypeError: 'type' object is not subscriptable
```
If I downgrade to release 1.9.7 it works again.

You might want to update python requirements to indicate python>=3.9 is required

_Originally posted by @sergeniushpe in https://github.com/aio-libs/yarl/issues/1084#issuecomment-2330345023_

bdraco:
## What do these changes do?

Fix non-subscriptable MultiDictProxy error as seen in #1105 when using older versions of multidict

Screenshot 2024-09-04 at 4 00 13 PM

no

## Related issue number
fixes #1105

### DIFF
--- a/CHANGES/1084.bugfix.rst
+++ b/CHANGES/1084.bugfix.rst
@@ -1,0 +1,1 @@
+1107.bugfix.rst
--- a/CHANGES/1105.bugfix.rst
+++ b/CHANGES/1105.bugfix.rst
@@ -1,0 +1,1 @@
+1107.bugfix.rst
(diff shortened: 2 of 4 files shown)


bdraco:
Tested locally, confirmed fixed

PR #1111, opened 2024-09-05
📦 Start building wheels with free-threading

webknjaz:
## What do these changes do?

Fancy 3.13t stuff.

Being able to install from wheels under free-threaded Python builds.

## Related issue number

N/A

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,6 +73,7 @@ before-test = [
   # Ref: https://github.com/pypa/cibuildwheel/issues/1666
   "PIP_CONSTRAINT= pip install PyYAML",
 ]
+free-threaded-support = true
 test-requires = "-r requirements/test.txt"
 test-command = "pytest -v --no-cov {project}/tests"
 # don't build PyPy wheels, install from source instead


webknjaz:
Looks like this needs an unreleased Cython version from Git.

bdraco:
I think we need to wait for Cython 3.1+ to do this 

https://docs.cython.org/en/latest/src/userguide/freethreading.html

bdraco:
Looks like 3.1 is getting closer https://github.com/cython/cython/milestone/65

webknjaz:
This is fine, we're not in a hurry. It's always been like this.

asvetlov:
The target cannot be solved by upgrading Cython version only.

IIRC the quoter uses pre-allocated memory buffer for small strings: https://github.com/aio-libs/yarl/blob/master/yarl/_quoting_c.pyx#L28

Now it's access is protected by GIL, gil-less solution requires some sort of explicit synchronization.

pyproject.toml
```diff
@@ -73,6 +73,7 @@ before-test = [
   # Ref: https://github.com/pypa/cibuildwheel/issues/1666
   "PIP_CONSTRAINT= pip install PyYAML",
 ]
+free-threaded-support = true
```
webknjaz, line 76: This got replaced with `
enable = ["cpython-freethreading"]` which https://iscinumpy.dev/post/cibuildwheel-3-0-0/ also emphasizes.

webknjaz:
Closing as it's been implemented in #1456 already.

Issue #196, opened 2018-05-08
Use custom URL parser


asvetlov:
The library uses a parser from stdlib.
It works but the implementation is not as fast as possible.
I've added tests for parser already.

The plan is:
1. Implement a custom pure python parser, all existing tests should be passed.
2. Add an optional Cythonized parser as the second step

PR #1112, opened 2024-09-05
Significantly speed up parsing netloc in URL objects

bdraco:
## What do these changes do?

Currently to make a URL object that is not already encoded we would call `urlsplit` to generate a `SplitResult`, and than we would always call `.hostname`, `.port`,`.username`, and `.password`, which all have to reparse the `netloc` each time. 

To improve the performance
 - parse out the `hostname`, `port`,`username`, and `password` one time in a new function `_split_netloc` which is a slimmed down version of what `SplitResult` does that only does what we need.
 - wrap `_split_netloc` in `lru_cache` the same size as the one `urlsplit` uses since we tend to see the same `netloc` over and over again. 
 - When `URL` is created with `encoded=False`, save the parsed values from the `netloc` and `SplitResult` into the cache so they do not have to be calculated again the first time the properties are accessed as it avoids a second call into `SplitResult`'s parsers of the same data.
 - When `URL` is created with `encoded=True`, parse the `netloc` as soon as `raw_password`, `raw_user`, `raw_host` or `explict_port` are accessed and cache the result of the other values to avoid having to parse again when they are accessed  since is almost guaranteed that we will access raw_password if we access raw_user and vise-versa. Same with `explict_port` and `raw_host`.

Even on the cache miss case this is much faster because we avoid so many reparses of the same data. For an `aiohttp` request with url passed as string we would previously have to parse the `netloc` 8x (4x for the properties to build the netloc, and 4x for the reading the properties back from `yarl`)

The additional LRU will use a tiny bit more memory, but is limited to 128 keys which will make it negligible

```
# timeit.timeit('URL("http://user:pass@example.com?a+b=c+d").raw_host', globals=globals(), number=100000)
1.10.0.dev0: 0.13812883291393518
1.9.11: 0.2996841249987483

# timeit.timeit('URL("http://example.com?a+b=c+d").raw_host', globals=globals(), number=100000)
1.10.0.dev0: 0.11170849995687604
1.9.11: 0.22350204200483859

# timeit.timeit('URL("http://example.com").raw_host', globals=globals(), number=100000)
1.10.0.dev0: 0.09903637506067753
1.9.11: 0.22337879217229784

# timeit.timeit('URL("http://example.com")', globals=globals(), number=100000)
1.10.0.dev0: 0.09579283301718533
1.9.11: 0.16477595805190504
```

before
new_before

after
new_after

## Related issue

#196 (only partially solves the issues mentioned there)


bdraco:
Need to look at the aiohttp server case a bit more. In some cases sending an unencoded url beats the pre encoded case depending on what we access because of cache misses and reparse of the same data

yarl/_url.py
```diff
@@ -925,6 +930,48 @@ def _make_netloc(
...
+            port: Optional[int] = None
+        else:
+            try:
+                port = int(port_str)
+            except ValueError:
+                raise ValueError("Invalid URL: port can't be converted to integer")
+            if not (0 <= port <= 65535):
+                raise ValueError("Port out of range 0-65535")
+
+        return username, password, cls._encode_host(hostname), port
```
bdraco, line 973: Might make more sense to move the encode to __new__ since it's the only place we this. Than we can reuse this function as a basis to parse when one of the raw netloc properties is accessed so we can lazy fill the cache for the pre encoded case

bdraco:
3 min up time
`2024-09-05 20:21:33.691 CRITICAL (SyncWorker_26) [homeassistant.components.profiler] Cache stats for lru_cache <function URL._split_netloc at 0x7ff8fea02700> at /usr/local/lib/python3.12/site-packages/yarl/_url.py: CacheInfo(hits=2086, misses=123, maxsize=128, currsize=123)`

yarl/_url.py
```diff
@@ -413,6 +423,16 @@ def __setstate__(self, state):
...
         self._cache = {}
 
+    def _cache_netloc(self) -> None:
+        """Cache the netloc parts of the URL."""
+        cache = self._cache
+        (
+            cache["raw_user"],
+            cache["raw_password"],
+            cache["raw_host"],
+            cache["explicit_port"],
```
bdraco, line 433: These lines are covered with coverage and codecov is only reporting mypy doesn't see them

bdraco:
`_url.py` is just barely above `client_proto.py` in the profile now.  It was far above it before.

bdraco:
Will sleep on this one to see if there is any more coverage that would be good to add

yarl/_url.py
```diff
@@ -925,6 +949,42 @@ def _make_netloc(
...
+
+        if "[" in hostinfo:
+            _, _, bracketed = hostinfo.partition("[")
+            hostname, _, port_str = bracketed.partition("]")
+            _, _, port_str = port_str.partition(":")
+        else:
+            hostname, _, port_str = hostinfo.partition(":")
+
+        if not port_str:
+            port: Optional[int] = None
```
webknjaz, line 977:
```suggestion
            port: Union[int, None] = None
```

bdraco: I'm so used to writing `Optional`. Let me do https://github.com/aio-libs/yarl/pull/1095 first so I won't add any more of these and remove `Optional` from the imports

webknjaz: Would https://pypi.org/p/no-optional help?

bdraco: Once the import is gone I'll remember.

webknjaz: Yeah, but in general, it might be better for other contributors. That's for the future, though.

yarl/_url.py
```diff
@@ -209,38 +210,50 @@ def __new__(
...
-                try:
-                    port = val.port
-                except ValueError as e:
-                    raise ValueError(
-                        "Invalid URL: port can't be converted to integer"
-                    ) from e
-
+                host = cls._encode_host(host)
+                raw_user = None if username is None else cls._REQUOTER(username)
+                raw_password = None if password is None else cls._REQUOTER(password)
```
bdraco, line 225: Should make sure we have an explicit test here. Not sure if an empty string is a valid password

yarl/_url.py
```diff
@@ -209,38 +210,50 @@ def __new__(
         else:
             raise TypeError("Constructor parameter should be str")
 
+        cache: Dict[str, Union[str, int, None]] = {}
```
bdraco, line 213: It would be better to make this a typed dict so any mistakes get caught early

yarl/_url.py
bdraco: Already covered by `test_empty_password_without_user`

yarl/_url.py
bdraco: Probably something for a future PR though as I don't want to make this one larger

bdraco:
Retest on Home Assistant production installs looks good

newnew

bdraco:
3 hours
2024-09-06 23:01:09.034 CRITICAL (SyncWorker_30) [homeassistant.components.profiler] Cache stats for lru_cache <function URL._split_netloc at 0x7f23be2e5940> at /usr/local/lib/python3.12/site-packages/yarl/_url.py: CacheInfo(hits=37364, misses=104, maxsize=128, currsize=104)

bdraco:
13 hours

2024-09-07 07:01:46.909 CRITICAL (SyncWorker_32) [homeassistant.components.profiler] Cache stats for lru_cache <function URL._split_netloc at 0x7f4f1c6e6700> at /usr/local/lib/python3.12/site-packages/yarl/_url.py: CacheInfo(hits=157229, misses=129, maxsize=128, currsize=128)

bdraco:
17 hours on another HA install
2024-09-07 12:08:06.055 CRITICAL (SyncWorker_7) [homeassistant.components.profiler] Cache stats for lru_cache <function URL._split_netloc at 0x7f23be2e5940> at /usr/local/lib/python3.12/site-packages/yarl/_url.py: CacheInfo(hits=173706, misses=105, maxsize=128, currsize=105)


So its turning out to be a really great use of an LRU

PR #1113, opened 2024-09-06
Make ``SimpleQuery``, ``QueryVariable``, and ``Query`` types available

### LINKED ISSUE #1050: Export Query type
In aiohttp we define `params` which is simply passed to URL.with_query(). We should have this parameter typed to allow the same things, but the type is rather complex:
https://github.com/aio-libs/yarl/blob/efea830d088309214450b359f0d02792ba2eaa50/yarl/__init__.pyi#L18-L22

Therefore, it'd be useful to export this type from yarl in order to be reused.

bdraco:
## What do these changes do?

The previously protected types `_SimpleQuery`, `_QueryVariable`, and `_Query` are now available for use externally.

no

## Related issue number
closes #1050

### DIFF
--- a/tests/test_url_query.py
+++ b/tests/test_url_query.py
@@ -134,10 +134,10 @@ def test_query_separators_from_parsing(
     URLS_WITH_RESERVED_CHARS_IN_QUERY_VALUES_W_XFAIL,
 )
 def test_query_separators_from_update_query(
-    original_url,
-    expected_query_len,
-    expected_value_a,
-):
+    original_url: URL,
+    expected_query_len: int,
+    expected_value_a: str,
+) -> None:
     new_url = original_url.update_query({"c": expected_value_a})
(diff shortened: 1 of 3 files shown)


CHANGES/1113.contrib.rst
```diff
@@ -0,0 +1 @@
+The previously protected types ``_SimpleQuery``, ``_QueryVariable``, and ``_Query`` are now available for use externally as ``SimpleQuery``, ``QueryVariable``, and ``Query`` -- by user:`bdraco`.
```
bdraco, line 1:
```suggestion
The previously protected types ``_SimpleQuery``, ``_QueryVariable``, and ``_Query`` are now available for use externally as ``SimpleQuery``, ``QueryVariable``, and ``Query`` -- by :user:`bdraco`.
```

PR #1117, opened 2024-09-06
Add types to the internal cache

bdraco:
## What do these changes do?

The goal is to avoid errors when pre-populating the cache

no

## Related issue number

https://github.com/aio-libs/yarl/pull/1112#discussion_r1747515069

### DIFF
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -9,7 +9,6 @@ from typing import (
     TYPE_CHECKING,
     Any,
     Callable,
-    Dict,
     Iterable,
     Iterator,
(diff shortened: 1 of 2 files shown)


yarl/_url.py
```diff
@@ -505,6 +526,20 @@ def relative(self) -> "URL":
         val = self._val._replace(scheme="", netloc="")
         return URL(val, encoded=True)
 
+    @cached_property
```
bdraco, line 529: No changes here, only a relocation to keep all the properties together

PR #1125, opened 2024-09-07
Small optimization to encoding hosts

bdraco:
## What do these changes do?
Avoid a partition call for most cases since its rare that the host will have a IPv6 zone.  We guard all the other partitions but this one was missed, and its one of the most common in the hot path.

before total time = 0.040
   100000    0.040    0.000    0.068    0.000 _url.py:907(_encode_host)

after total time = 0.031
   100000    0.031    0.000    0.051    0.000 _url.py:907(_encode_host)

```
print(timeit.timeit('URL("http://user:pass@example.com?a+b=c+d").raw_host', globals=globals(), number=100000))
```

no

### DIFF
--- a/CHANGES/1125.misc.rst
+++ b/CHANGES/1125.misc.rst
@@ -1,1 +1,1 @@
-Improve performance of encoding non IPv6 hosts -- by :user:`bdraco`.
+Improved performance of encoding non IPv6 hosts -- by :user:`bdraco`.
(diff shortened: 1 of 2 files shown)


CHANGES/1125.misc.rst
```diff
@@ -0,0 +1 @@
+Improve performance of encoding non IPv6 hosts -- by :user:`bdraco`.
```
bdraco, line 1:
```suggestion
Improved performance of encoding non IPv6 hosts -- by :user:`bdraco`.
```

Issue #1127, opened 2024-09-08
Passing params to a new client request generates multiple urls because yarl does not support merging query params


bdraco:
https://github.com/aio-libs/aiohttp/blob/11a96fcc7b62b33f0088640a05237e262cac5691/aiohttp/client_reqrep.py#L231

Maybe a new method `.extend_query`

Need to see how this behaves differently from `.update_query` and why it wasn't used
<img width="1492" alt="Screenshot 2024-09-07 at 10 27 34 PM" src="https://github.com/user-attachments/assets/0667c6dc-5da2-497e-81a7-0c73aacf1d1d">
<img width="1325" alt="Screenshot 2024-09-07 at 10 40 40 PM" src="https://github.com/user-attachments/assets/92b3eb53-402e-43e3-bf8c-19726c664b49">

bdraco:
It looks like the intent in aiohttp is to not overwrite existing values

bdraco:
Fixed by #1128

PR #1128, opened 2024-09-08
Add extend_query method

bdraco:
## What do these changes do?

Adds a new method called `extend_query` that will add additional keys instead of replacing them like `update_query`.  [The goal is to replace the hand rolled version in `aiohttp`](https://github.com/aio-libs/aiohttp/blob/11a96fcc7b62b33f0088640a05237e262cac5691/aiohttp/client_reqrep.py#L231)

no

## Related issue number
#1127

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -659,6 +659,48 @@ section generates a new :class:`URL` instance.
       Support subclasses of :class:`int` (except :class:`bool`) and :class:`float`
       as a query parameter value.
 
+.. method:: URL.extend_query(query)
+            URL.extend_query(**kwargs)
+
+   Returns a new URL with *query* part extended.
+
+   Unlike :meth:`update_query` the method keeps duplicate keys.
+
+   Returned :class:`URL` object will contain query string which extends
+   parts from passed query parts (or parts of parsed query string).
+
+   Accepts any :class:`~collections.abc.Mapping` (e.g. :class:`dict`,
+   :class:`~multidict.MultiDict` instances) or :class:`str`,
+   auto-encode the argument if needed.
+
+   A sequence of ``(key, value)`` pairs is supported as well.
+
+   Also it can take an arbitrary number of keyword arguments.
+
+   Returns the same :class:`URL` if *query* of ``None`` is passed.
+
+   .. note::
+
+      The library accepts :class:`str`, :class:`float`, :class:`int` and their
+      subclasses except :class:`bool` as query argument values.
+
+      If a mapping such as :class:`dict` is used, the values may also be
+      :class:`list` or :class:`tuple` to represent a key has many values.
+
+      Please see :ref:`yarl-bools-support` for the reason why :class:`bool` is not
+      supported out-of-the-box.
+
+   .. doctest::
+
+      >>> URL('http://example.com/path?a=b&b=1').extend_query(b='2')
+      URL('http://example.com/path?a=b&b=1&b=2')
+      >>> URL('http://example.com/path?a=b&b=1').extend_query([('b', '2')])
+      URL('http://example.com/path?a=b&b=1&b=2')
+      >>> URL('http://example.com/path?a=b&c=e&c=f').extend_query(c='d')
+      URL('http://example.com/path?a=b&c=e&c=f&c=d')
+
+   .. versionadded:: 1.11.0
+
 .. method:: URL.update_query(query)
             URL.update_query(**kwargs)
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -664,7 +664,7 @@ section generates a new :class:`URL` instance.
 
    Returns a new URL with *query* part extended.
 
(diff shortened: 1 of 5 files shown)


bdraco:
Verified this works as expected in `aiohttp` with this change

```diff
diff --git a/aiohttp/client_reqrep.py b/aiohttp/client_reqrep.py
index 6eaba561b..6e0f76a2e 100644
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -229,10 +229,7 @@ class ClientRequest:
         # assert session is not None
         self._session = cast("ClientSession", session)
         if params:
-            q = MultiDict(url.query)
-            url2 = url.with_query(params)
-            q.extend(url2.query)
-            url = url.with_query(q)
+            url = url.extend_query(params)
         self.original_url = url
         self.url = url.with_fragment(None)
         self.method = method.upper()
```

bdraco:
Doing a bit of production testing before merge. If that all goes well, I'll bump the version, do a release and than open a PR to aiohttp to replace the hand rolled method

yarl/_url.py
```diff
@@ -1250,17 +1280,35 @@ def update_query(self, query: Query) -> "URL": ...
...
+        >>> url.update_query(a=3, c=4)
+        URL('http://example.com/?a=3&b=2&c=4')
+        """
+        new_query_string = self._get_str_query(*args, **kwargs)
+        if new_query_string is None:
+            return URL(self._val._replace(query=""), encoded=True)
+        return self._merge_query(new_query_string, update=True)
+
+    def _merge_query(self, new_query_string: Union[str, None], update: bool) -> "URL":
+        """Return a new URL with query part merged or extended."""
```
bdraco, line 1298:
```suggestion
        """Return a new URL with query part updated or extended."""
```

bdraco:
In turns out I didn't need to split update_query so I can split this PR up a bit

yarl/_url.py
```diff
@@ -1204,29 +1204,29 @@ def _get_str_query(self, *args: Any, **kwargs: Any) -> Union[str, None]:
             raise ValueError("Either kwargs or single query parameter must be present")
 
         if query is None:
-            query = None
-        elif isinstance(query, Mapping):
+            return None
+        if isinstance(query, (MultiDict, MultiDictProxy)):
+            return self._get_str_query_from_iterable(query.items())
```
bdraco, line 1209:
```suggestion
```

Would be nice to do this, and its probably a bugfix, but its already a breaking change since even though the docs don't say we do, we allowed lists/tuples inside of multidict

yarl/_url.py
```diff
@@ -1191,7 +1191,7 @@ def _get_str_query_from_iterable(
         return "&".join([f"{quoter(k)}={quoter(self._query_var(v))}" for k, v in items])
 
     def _get_str_query(self, *args: Any, **kwargs: Any) -> Union[str, None]:
-        query: Union[str, Mapping[str, QueryVariable], None]
+        query: Union[str, "Mapping[str, QueryVariable]", None]
```
bdraco, line 1194:
```suggestion
        query: Union[str, Mapping[str, QueryVariable], None]
```

docs/api.rst
```diff
@@ -659,6 +659,48 @@ section generates a new :class:`URL` instance.
       Support subclasses of :class:`int` (except :class:`bool`) and :class:`float`
       as a query parameter value.
 
+.. method:: URL.extend_query(query)
+            URL.extend_query(**kwargs)
+
+   Returns a new URL with *query* part extended.
+
+   Unlike :meth:`update_query` the method keeps duplicate keys.
```
bdraco, line 667:
```suggestion
   Unlike :meth:`update_query`, this method keeps duplicate keys.
```

yarl/_url.py
```diff
@@ -1250,14 +1250,50 @@ def with_query(self, *args: Any, **kwargs: Any) -> "URL":
...
+        Example:
+        >>> url = URL('http://example.com/?a=1&b=2')
+        >>> url.extend_query(a=3, c=4)
+        URL('http://example.com/?a=1&b=2&a=3&c=4')
+        """
+        new_query_string = self._get_str_query(*args, **kwargs)
+        if not new_query_string:
+            return self
+        if current_query := self.raw_query_string:
+            if current_query[-1] == "&":
```
bdraco, line 1273:
```suggestion
            # both strings are already encoded so we can use a simple
            # string join
            if current_query[-1] == "&":
```

yarl/_url.py
```diff
@@ -1163,7 +1163,7 @@ def _query_var(v: QueryVariable) -> str:
             if TYPE_CHECKING:
                 assert isinstance(v, str)
             return v
-        if issubclass(cls, float):
+        if cls is float or issubclass(cls, float):
```
bdraco, line 1166:
```suggestion
        if issubclass(cls, float):
```

PR #1130, opened 2024-09-08
Improve performance of URL.update_query

bdraco:
## What do these changes do?

- Avoid reparsing the query string into a MultiDict
- Avoid creation of many intermediate MultiDict and MultiDictProxy objects
- Build the final string with iterable helper since there will never be `QueryVariable`

no

### DIFF
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -88,6 +88,7 @@ class _InternalURLCache(TypedDict, total=False):
     explicit_port: Union[int, None]
     raw_path: str
     path: str
+    _parsed_query: List[Tuple[str, str]]
     query: "MultiDictProxy[str]"
     raw_query_string: str
     query_string: str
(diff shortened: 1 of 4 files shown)


yarl/_url.py
```diff
@@ -1250,15 +1261,13 @@ def update_query(self, **kwargs: QueryVariable) -> "URL": ...
...
-            query.update(new_query)
+        if s is None:
+            return URL(self._val._replace(query=""), encoded=True)
 
-        return URL(
-            self._val._replace(query=self._get_str_query(query) or ""), encoded=True
-        )
+        query = MultiDict(self._parsed_query)
+        query.update(parse_qsl(s, keep_blank_values=True))
+        new_str = self._get_str_query_from_iterable(query.items())
```
bdraco, line 1269: We avoid `_get_str_query` here since we already know which one to use

PR #1139, opened 2024-09-09
Accept objects that support __int__ for query vars

### LINKED ISSUE #945: Allow numpy values in query
### Describe the bug

Passing a numpy integer value (numpy.int16, numpy.int32 , numpy.int64 etc.) in dictionary as GET request '_params_' gives an TypeError: Invalid variable type: value should be str, int or float, got 12345 of type <class 'numpy.int16'>

```python
import numpy as np
import aiohttp 

number = np.int16(12345) 
params = {"number": number}

async with aiohttp.ClientSession() as session: 
    async with session.get(url, params=params)  as resp: 
        await resp.text()
```
But it works properly if passing numpy integer values in a dictionary as POST request '_data_'.

### To Reproduce

Place any numpy integer value (numpy.int16, numpy.int32 , numpy.int64 etc.) in  GET request '_params_' 

### Expected behavior

To accept numpy integer type in  GET request '_params_'

### Logs/tracebacks

```python-traceback
TypeError: Invalid variable type: value should be str, int or float, got 12345 of type <class 'numpy.int16'>
```

### Python Version

```console
$ python --version
Version: 3.8.8
```

### aiohttp Version

```console
$ python -m pip show aiohttp
Version: 3.8.5
```

### multidict Version

```console
Version: 6.0.4
```

### yarl Version

```console
$ python -m pip show yarl
Version: 1.9.2
```

### OS

Windows

### Related component

Client

### Additional context

_No response_

### Code of Conduct

bdraco:
## What do these changes do?

Previously we would only accept subclasses of int, we will now accept any object that implements __int__ and has the correct protocol.

`np.int16` will now be accepted

## Related issue number
closes #945

### DIFF
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -12,6 +12,7 @@ from typing import (
     Iterable,
     Iterator,
     List,
+    SupportsInt,
     Tuple,
     TypedDict,
     TypeVar,
@@ -1177,7 +1178,7 @@ class URL:
             if TYPE_CHECKING:
                 assert isinstance(v, str)
             return v
-        if issubclass(cls, float):
+        if cls is float or issubclass(cls, float):
             if TYPE_CHECKING:
                 assert isinstance(v, float)
             if math.isinf(v):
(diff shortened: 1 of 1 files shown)


yarl/_url.py
```diff
@@ -1177,17 +1178,15 @@ def _query_var(v: QueryVariable) -> str:
             if TYPE_CHECKING:
                 assert isinstance(v, str)
             return v
-        if issubclass(cls, float):
+        if cls is float or issubclass(cls, float):
```
bdraco:
```suggestion
        if issubclass(cls, float):
```

bdraco:
This caused a performance regression, but we didn't know it at the time since we were not monitoring performance

Fixed in https://github.com/aio-libs/yarl/pull/1259

PR #1143, opened 2024-09-10
Small performance improvements to checking for chars in paths

bdraco:
## What do these changes do?

Replace startswith with single char checks
startswith: 0.005655332934111357
single char check: 0.0019642498809844255

no


bdraco:
There is also an endswith that could be optimized but its rare to hit it

PR #1153, opened 2024-09-23
Increase release timeout from 7 to 14 minutes

bdraco:
The previous release hit the timeout of 7 minutes so the signatures did not upload.  Double the timeout to 14 minutes so this does not happen again.


bdraco:
I'm going to do another release so we have the latest with signatures.  Sadly, there is nothing I can do about them missing from 1.12.0 though.

PR #1154, opened 2024-09-23
Release 1.12.1

bdraco:
https://github.com/aio-libs/yarl/actions/runs/11002449369

This is a no change release of 1.12.0 since the signatures failure to upload because the release process hit the timeout.  The timeout has now been increased to 14 minutes from 7 minutes to ensure this does not happen again.


webknjaz:
By the way, a post release could also be an option.

PR #1158, opened 2024-09-26
Fix round-trip of IPv6 addresses

bdraco:
## What do these changes do?

The brackets were missing when the URL was convert to a string

fixes #1157

## Are there changes in behavior for the user?
no

Issue #1157, opened 2024-09-26
Parsing an IPv6 address URL with no port number (e.g. `http://[::1]/`) loses the square brackets


c3pmark:
### Please confirm the following

- [x] I agree to follow the [aio-libs Code of Conduct][CoC]

[CoC]: ../../.github/blob/master/CODE_OF_CONDUCT.md

- [X] I have checked the [current issues][issue search] for duplicates.

[issue search]: ../search?q=is%3Aissue&type=issues

- [X] I understand this is open source software provided for free and that I might not receive a timely response.
- [X] I am positive I am **NOT** reporting a (potential) security
vulnerability, to the best of my knowledge. *(These must be shared by
submitting [this report form][vulnerability report form] instead, if
any hesitation exists.)*

[vulnerability report form]: ../security/advisories/new

- [X] I am willing to submit a pull request with reporoducers as xfailing test cases or even entire fix. *(Assign this issue to me.)*

### Describe the bug

When parsing a URL containing an IPv6 address in square brackets with no port number, the brackets vanish. This was introduced in d5c5ab38; before that, the square brackets were preserved.

I'm pretty sure this is incorrect behavior, but I'm not an expert on URL parsing. If it is a bug, I can work on tests and a PR.

### To Reproduce

```
>>> yarl.URL("http://[::1]")
URL('http://::1')
>>> yarl.URL("http://[2607:f8b0:400b:802::200e]")
URL('http://2607:f8b0:400b:802::200e')
```

### Expected behavior

I expected the brackets to be preserved.
```
>>> yarl.URL("http://[::1]")
URL('http://[::1]')
>>> yarl.URL("http://[2607:f8b0:400b:802::200e]")
URL('http://[2607:f8b0:400b:802::200e]')
```

### Logs/tracebacks

```python-traceback
N/A
```

### Python Version

```console
$ python --version
Python 3.12.6
```

### multidict Version

```console
$ python -m pip show multidict
Name: multidict
Version: 6.0.5
Summary: multidict implementation
Home-page: https://github.com/aio-libs/multidict
Author: Andrew Svetlov
Author-email: andrew.svetlov@gmail.com
License: Apache 2
Location: /usr/lib64/python3.12/site-packages
Requires:
Required-by: aiohttp, yarl
```

### yarl Version

```console
Git master as of bf2ced70

$ python -m pip show yarl
Name: yarl
Version: 1.12.2.dev0
Summary: Yet another URL library
[...]

PR #1159, opened 2024-09-26
Add `host_subcomponent` property

bdraco:
Returns the https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2 host subcomponent. The only current practical difference between `raw_host` and `host_subcomponent` is that IPv6 addresses are returned bracketed.

We can also use this in `aiohttp` at https://github.com/aio-libs/aiohttp/blob/d7cd0613472fd4d9940e37f1c55921f6a1515324/aiohttp/client_reqrep.py#L359 and https://github.com/aio-libs/aiohttp/blob/d7cd0613472fd4d9940e37f1c55921f6a1515324/aiohttp/client_reqrep.py#L609

Split from #1158 for #1157

Note that we CANNOT use `self._val.hostname` here because it returns the non-bracketed host

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -206,7 +206,7 @@ There are two kinds of properties: *decoded* and *encoded* (with
       >>> URL('http://[::1]').host_subcomponent
       '[::1]'
 
-   .. versionadded:: 1.13.0
+   .. versionadded:: 1.13
 
 .. attribute:: URL.port
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -303,6 +303,8 @@ There are two kinds of properties: *decoded* and *encoded* (with
    result as :meth:`URL.path`. If the ``%25`` was unquoted, it would be
    impossible to tell the difference between ``%2F`` and ``%252F``.
 
+   .. versionadded:: 1.12
+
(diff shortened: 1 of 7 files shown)


yarl/_url.py
```diff
@@ -395,7 +395,7 @@ def __str__(self) -> str:
                 netloc=self._make_netloc(
                     self.raw_user,
                     self.raw_password,
-                    self.raw_host,
+                    self.host_subcomponent,
```
bdraco, line 398:
```suggestion
                    self.raw_host,
```

docs/api.rst
```diff
@@ -191,7 +191,22 @@ There are two kinds of properties: *decoded* and *encoded* (with
...
+   (:ref:`yarl-api-relative-urls`).
+
+   .. doctest::
+
+      >>> URL('http://хост.домен').host_subcomponent
+      'xn--n1agdj.xn--d1acufc'
+      >>> URL('http://[::1]').host_subcomponent
+      '[::1]'
 
+   .. versionadded:: 1.13.0
```
bdraco, line 209:
```suggestion
   .. versionadded:: 1.13
```

PR #1163, opened 2024-09-27
Improve performance of building URLs with authority

bdraco:
## What do these changes do?

Use the fast `_split_netloc` method so we do not have to reparse the netloc for every property.

Previously we called each `SplitResult` property, and each property call has to reparse the netloc every time. Only do it once

no

## Related issue number
https://github.com/aio-libs/aiohttp/pull/9309#discussion_r1778760440 will call `URL.build` with `authority` much more often in a performance sensitive path.

### DIFF
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -368,10 +368,22 @@ class URL:
             if encoded:
                 netloc = authority
             else:
-                tmp = SplitResult("", authority, "", "", "")
-                port = None if tmp.port == DEFAULT_PORTS.get(scheme) else tmp.port
+                tmp_username, tmp_password, tmp_host, tmp_port = cls._split_netloc(
+                    authority
+                )
+                port = None if tmp_port == DEFAULT_PORTS.get(scheme) else tmp_port
+                tmp_host = (
+                    ""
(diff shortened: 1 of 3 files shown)


yarl/_url.py
```diff
@@ -368,10 +368,15 @@ def build(
...
             else:
-                tmp = SplitResult("", authority, "", "", "")
-                port = None if tmp.port == DEFAULT_PORTS.get(scheme) else tmp.port
+                _user, _password, _host, _port = cls._split_netloc(authority)
+                port = None if _port == DEFAULT_PORTS.get(scheme) else _port
+                _host = (
+                    ""
+                    if _host is None
+                    else cls._encode_host(_host, validate_host=False)
+                )
```
bdraco, line 377: host is not validated against scheme like it is in `__new__` but thats another issue

bdraco:
Verified new tests pass on master as well so we have bug for bug compat with the faster version

bdraco:
I'm going to do a new release of yarl before we publish a new aiohttp with https://github.com/aio-libs/aiohttp/pull/9309#discussion_r1778760440 fixed

PR #1175, opened 2024-10-05
Improve performance of building the origin

### LINKED ISSUE #1174: Calling URL.origin() is expensive when we cannot avoid it in the cookie jar
> Open PRs that probably resolve these performance issues: aio-libs/aiohttp#7784 aio-libs/aiohttp#7777 aio-libs/aiohttp#7790

I see. But they do not eliminate the need to call `URL.origin()`, which is also expensive, even when the jar is empty. Would you think that my suggestion is a good idea? If so, I can open a PR.

_Originally posted by @Rongronggg9 in https://github.com/aio-libs/aiohttp/issues/7583#issuecomment-1806828707_

bdraco:
## What do these changes do?

- Avoid rebuilding the netloc if there is no auth
- Cache the origin so we only build it once as aiohttp will call `.origin()` often on the same URL object when auth is enabled

no

## Related issue number
fixes #1174

### DIFF
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -95,6 +95,7 @@ class _SplitResultDict(TypedDict, total=False):
 
 class _InternalURLCache(TypedDict, total=False):
 
+    _origin: "URL"
     absolute: bool
     scheme: str
     raw_authority: str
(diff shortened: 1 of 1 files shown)


bdraco:
```
from yarl import URL

import timeit

from yarl import _url

url_with_user_pass = URL("http://user:pass@example.com?a+b=c+d")
url_without_user_pass = URL("http://example.com?a+b=c+d")

print(['with user/pass', timeit.timeit('url_with_user_pass.origin()', globals=globals(), number=100000)])
print(['without user/pass', timeit.timeit('url_without_user_pass.origin()', globals=globals(), number=100000)])
```

For testing (note that cache needs to be removed to properly test)

PR #1176, opened 2024-10-05
Improve performance of _encode_host

bdraco:
## What do these changes do?

Improve performance of _encode_host

- If we are not running the potential ip address path, avoid checking for IPv6 zone
- If the parsed address is not IPv6, avoid trying to join the IPv6 zone back to it

This one is very small, I only noticed it because I was looking over `_encode_host`
and wondering why we were looking for an IP zone outside of the ip address block

no

before
__new__before

after
__new__after

### DIFF
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -1048,7 +1048,6 @@ class URL:
             if validate_host:
                 _host_validate(host)
             return host
-
         return _idna_encode(host)
 
     @classmethod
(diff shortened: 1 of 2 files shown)


yarl/_url.py
```diff
@@ -1049,6 +1048,7 @@ def _encode_host(
             if validate_host:
                 _host_validate(host)
             return host
+
```
bdraco, line 1051:
```suggestion
```

bdraco:
While this is wrong, we probably shouldn't change it....

PR #1182, opened 2024-10-08
Add test coverage for handling port 0

bdraco:
## What do these changes do?

This was wrong in previous yarl releases, but fixed by one of the cleanups in 1.14.0. Add coverage to make sure it continues to work as 0 is not a default port

The problem:

`str(URL("http://example.com:0/"))` should be `http://example.com:0/` because `0` is not the default port for `http`

yarl 1.9.5 - 1.13.1 produced it as `http://example.com/`

### DIFF
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -334,14 +334,6 @@ def test_explicit_port_for_implicit_port():
     assert url._port_not_default is None
 
 
-def test_handling_port_zero():
-    url = URL("http://example.com:0")
-    assert url.explicit_port == 0
-    assert url.explicit_port == url._val.port
-    assert url._port_not_default == 0
-    assert str(url) == "http://example.com:0"
-
-
 def test_explicit_port_for_relative_url():
     url = URL("/path/to")
     assert url.explicit_port is None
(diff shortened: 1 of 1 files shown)


webknjaz:
What was the fix, by the way?

bdraco:
> Isn't 0 only used as an ephemeral port that is resolved to something else by the kernel?

I don't expect anyone is using it in production, and this likely only affects downstream tests

bdraco:
> What was the fix, by the way?

I need to go back and figure out which PR fixed it.  Its likely a `None` check that was corrected

webknjaz:
I was under impression that using this port in URLs isn't even legal since it doesn't have defined semantics, I think 🤔

bdraco:
> I was under impression that using this port in URLs isn't even legal since it doesn't have defined semantics, I think 🤔

urllib seems to allow it and it does specifically check the port is between 0 and 65535
```
>>> from urllib.parse import urlparse
>>> urlparse("http://example.com:0")
ParseResult(scheme='http', netloc='example.com:0', path='', params='', query='', fragment='')
>>> urlparse("http://example.com:0").port
0
>>> 
>>> urlparse("http://example.com:70000").port
  File "<stdin>", line 1, in <module>
  File "/opt/homebrew/Cellar/python@3.12/3.12.4/Frameworks/Python.framework/Versions/3.12/lib/python3.12/urllib/parse.py", line 184, in port
    raise ValueError("Port out of range 0-65535")
ValueError: Port out of range 0-65535
```

bdraco:
was fixed in #1170

webknjaz:
Ack. Though, I wonder what the use-case for this is.

bdraco:
I'm not sure there is one. I only noticed because of downstream tests, but none of them actually use 0 in production.

bdraco:
It looks like it originally regressed in https://github.com/aio-libs/yarl/releases/tag/v1.9.5. I'm guessing https://github.com/aio-libs/yarl/pull/1033

webknjaz:
Perhaps it's useful in the context of non-HTTP URLs. Like those used with custom handlers in mobile apps and so on..

bdraco:
I'm sure someone is using it for something somewhere so its good to have coverage we don't regress it again.

PR #1188, opened 2024-10-11
Improve performance of building unencoded URLs

bdraco:
## What do these changes do?

Improve performance of building unencoded URLs

no

### DIFF
--- a/url_benchmark.py
+++ b/url_benchmark.py
@@ -1,0 +1,19 @@
+import timeit
+
+from yarl import URL
+
+print(
+    "Build URL with host and path: {:.3f} sec".format(
+        timeit.timeit(
+            "URL('http://localhost:1234/req')", globals={"URL": URL}, number=100000
+        )
+    )
+)
+
+print(
+    "Build URL with IP Address and path: {:.3f} sec".format(
+        timeit.timeit(
+            "URL('http://127.0.0.1:1234/req')", globals={"URL": URL}, number=100000
+        )
+    )
+)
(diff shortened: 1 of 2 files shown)


yarl/_url.py
```diff
@@ -1108,29 +1129,29 @@ def _split_netloc(
...
-        if not self.absolute and scheme in SCHEME_REQUIRES_HOST:
+        lower_scheme = scheme.lower()
+        if not self.absolute and lower_scheme in SCHEME_REQUIRES_HOST:
             msg = (
                 "scheme replacement is not allowed for "
                 f"relative URLs for the {scheme} scheme"
             )
             raise ValueError(msg)
-        return URL(self._val._replace(scheme=scheme.lower()), encoded=True)
+        return URL(self._val._replace(scheme=lower_scheme), encoded=True)
```
bdraco, line 1154: This is actually a bugfix so it should be moved into a new PR

yarl/_url.py
```diff
@@ -993,22 +987,44 @@ def _normalize_path(cls, path: str) -> str:
...
+        """
+        lower_host = host.lower()
+        is_ascii = host.isascii()
+
+        if host and (host[-1].isdigit() or ":" in host):
+            if "%" in host:
+                return True, lower_host, is_ascii, *host.partition("%")
+            return True, lower_host, is_ascii, host, "", ""
+
+        return False, lower_host, is_ascii, "", "", ""
```
bdraco, line 1013: This could be moved into a separate PR

url_benchmark.py
```diff
@@ -0,0 +1,81 @@
+import timeit
+
+from yarl import URL
+
```
bdraco, line 4: Should have some random ones as well

yarl/_url.py
bdraco: https://github.com/aio-libs/yarl/pull/1190

yarl/_url.py
bdraco: https://github.com/aio-libs/yarl/pull/1189

PR #1192, opened 2024-10-11
Improve performance of unquoter cython implementation

bdraco:
The unquoter would have to jump back into python to pass back to the quoter


bdraco:
we almost never hit this path .. so not worth the change

PR #1194, opened 2024-10-11
Add benchmarks for with_query

bdraco:
baseline

Make URL with query mapping: 0.364 sec
Make URL with query sequence mapping: 2.121 sec


bdraco:
actions is currently broken

PR #1197, opened 2024-10-11
Add benchmarks to converting URLs to strings

bdraco:
SSIA

master
```
Convert URL to string: 0.094 sec
Convert URL with path to string: 0.096 sec
Convert URL with query to string: 0.141 sec
```

1.9.4
```
Convert URL to string: 0.064 sec
Convert URL with path to string: 0.042 sec
Convert URL with query to string: 0.113 sec
```

1.9.5 -- fails -- this release should probably be yanked
```
^CTraceback (most recent call last):
  File "/opt/homebrew/lib/python3.12/site-packages/yarl/_url.py", line 470, in _default_port
    return socket.getservbyname(scheme)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: service/proto not found
```

1.9.6
```
Convert URL to string: 0.168 sec
Convert URL with path to string: 0.160 sec
Convert URL with query to string: 0.216 sec
```

### DIFF
--- a/url_benchmark.py
+++ b/url_benchmark.py
@@ -5,6 +5,8 @@ from yarl import URL
 MANY_HOSTS = [f"www.domain{i}.tld" for i in range(10000)]
 MANY_URLS = [f"https://www.domain{i}.tld" for i in range(10000)]
 BASE_URL = URL("http://www.domain.tld")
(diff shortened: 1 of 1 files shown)


bdraco:
Looks like we need to fix that performance regression

PR #1203, opened 2024-10-11
Drop support for Python 3.8

bdraco:
~~DNM: Should do a patch release to fix the performance regression in https://github.com/aio-libs/yarl/pull/1198 before dropping 3.8~~

## What do these changes do?

Python 3.8 has reached end of life
https://devguide.python.org/versions/

Remove support for Python 3.8


bdraco:
I'll leave this for a few days in case we need to publish a patch to fix any regressions.

bdraco:
master vs 1.15.0 shows ~17% improvement in HA. Probably worth doing one more patch before merging this so py3.8 gets the changes as well

bdraco:
All the rough edges identified by adding the benchmarks have been smoothed out.  1.15.2 looks like it will be the last release to support Python 3.8

bdraco:
Last gate is to wait for https://github.com/home-assistant/core/pull/128309 to pass

PR #1210, opened 2024-10-12
Integrate codspeed benchmarking

bdraco:
Add URL benchmarks to [codspeed](https://codspeed.io/aio-libs/yarl/benchmarks)

A followup PR will integration the quoting benchmarks

The goal is to tell us when a performance regression is introduced early before it makes it to production and hopefully catch any blocking I/O

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -421,6 +421,9 @@ jobs:
       shell: bash
     - name: Self-install
       run: python -Im pip install '${{ steps.wheel-file.outputs.path }}'
+    - name: Fix git owner
+      # https://github.com/CodSpeedHQ/action/issues/104
+      run: git config --global --add safe.directory "*"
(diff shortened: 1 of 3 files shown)


bdraco:
This one is public facing so I skipped the changelog message. Not sure if we should add something or not.

.github/workflows/ci-cd.yml
```diff
@@ -346,6 +346,89 @@ jobs:
...
+      uses: actions/setup-python@v5
+      with:
+        python-version: 3.12
+        cache: pip
+        cache-dependency-path: requirements/*.txt
+    - name: Install dependencies
+      uses: py-actions/py-dependency-install@v4
+      with:
+        path: requirements/test.txt
+    - name: Determine pre-compiled compatible wheel
```
bdraco, line 385: Would be nice to de-duplicate this somehow

PR #1211, opened 2024-10-12
Add quoting benchmarks

bdraco:
Add benchmarks for quoting to codspeed

For now we are only testing the cython versions


bdraco:
Yeah too few iterations

PR #1213, opened 2024-10-12
DNM: Benchmark testing - should show reduced performance

bdraco:
This is a test of Reverting "Improve performance of converting of converting URLs to strings (#1198)" to verify that codspeed reports the performance decrease

This reverts commit 77a2ed99de3544fd7f1765a9e84832fea3248e9a.


bdraco:
Looks like this is working well

PR #1214, opened 2024-10-12
DNM: Compare running benchmarks on Python 3.13

bdraco:
I picked Python 3.12 for the benchmarks because its late in the cycle and unlikely to have significant changes that affect the benchmarks but also new enough to be a good target. Would be nice to see how 3.13 affects things


bdraco:
Looks about right

bdraco:
PurePath has a lot of improvements in 3.13 so would be nice to see if URL subtraction will be magically faster in 3.13

bdraco:
Looks like py3.13 did a some work on optimizing PurePath a bit

bdraco:
Looks like we somehow optimized for Python 3.13 at some point.

Still we should keep running benchmarks on 3.12 until 3.13.1 is released and than move over

PR #1216, opened 2024-10-12
PoC: avoid loosing cached values when calling URL.build

bdraco:
When we make a url from a string we save the raw values, but URL.build throws them away

This version is a bit messy right now since its a PoC, but it can be made more DRY.

### DIFF
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -253,7 +253,6 @@ class URL:
         *,
         encoded: bool = False,
         strict: Union[bool, None] = None,
-        _cache: Union[_InternalURLCache, None] = None,
     ) -> Self:
         if strict is not None:  # pragma: no cover
             warnings.warn("strict parameter is ignored")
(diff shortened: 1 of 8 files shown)


yarl/_url.py
```diff
@@ -398,12 +407,27 @@ def build(
...
         url = cls(
-            SplitResult(scheme, netloc, path, query_string, fragment), encoded=True
+            SplitResult(scheme, netloc, path, query_string, fragment),
+            encoded=True,
+            _cache=cache,
         )
 
         if query:
             return url.with_query(query)
+
```
bdraco, line 430:
```suggestion
```

bdraco:
Looks like its not going to be worth it

We can though

Remove `requote` from `_make_netloc`

Use the faster [] stripping

bdraco:
version without the cache https://github.com/aio-libs/yarl/pull/1220

PR #1218, opened 2024-10-12
Small speed up to making IPv6 URLs

bdraco:
## What do these changes do?

Small speed up to making IPv6 URLs
https://codspeed.io/aio-libs/yarl/branches/strip_brackets it is ~4% but codespeed is configured to only report >= 5% change

no


bdraco:
performance improvement isn't significant enough to get a changelog message

PR #1222, opened 2024-10-12
Improve URL.build performance

bdraco:
## What do these changes do?

Improve URL.build performance

no


CHANGES/1222.misc.rst
```diff
@@ -0,0 +1 @@
+Improved performance of call :py:meth:`~yarl.URL.build` -- by :user:`bdraco`.
```
bdraco, line 1:
```suggestion
Improved performance of calling :py:meth:`~yarl.URL.build` -- by :user:`bdraco`.
```

PR #1224, opened 2024-10-12
Save host_subcomponent when parsing unencoded URLs

bdraco:


bdraco:
seems to not be worth it

PR #1231, opened 2024-10-12
DNM: Run benchmarks on Python 3.8

bdraco:
This is to get a baseline before we remove 3.8 support


bdraco:
Interesting that the cython method calls come up as faster

PR #1234, opened 2024-10-12
Improve performance of converting URL to string

bdraco:
~~Not sure if this is worth it or not, but the string convert method are a bit higher than everything else~~

Seems like 71%-3x speed up is worth it

Calling urllib's urlunsplit was quite a bit more expensive than expected as seen in the benchmarks.

### DIFF
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -417,8 +417,9 @@ class URL:
 
     def __str__(self) -> str:
         val = self._val
+        scheme, netloc, path, query, fragment = val
         if not val.path and self.absolute and (val.query or val.fragment):
-            val = val._replace(path="/")
+            path = "/"
         if (
             explicit_port := self.explicit_port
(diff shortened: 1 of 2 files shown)


yarl/_url.py
```diff
@@ -418,18 +417,31 @@ def __init_subclass__(cls):
...
+        if netloc or (scheme and scheme in USES_AUTHORITY) or url[:2] == "//":
+            if url and url[:1] != "/":
+                url = f"//{netloc or ''}/{url}"
+            else:
+                url = f"//{netloc or ''}{url}"
+        if scheme:
+            url = f"{scheme}:{url}"
+        if query:
+            url = f"{url}?{query}"
+        return f"{url}#{fragment}" if fragment else url
```
bdraco, line 444: Might be a lot faster to make a list and push and insert than join at the end.

Could be done in a future PR so we get a clean profile to prove the theory

PR #1245, opened 2024-10-13
Refactor URL to string conversion to avoid many string operations

bdraco:


bdraco:
was worth a shot, but slower

PR #1246, opened 2024-10-13
Benchmark concatenation instead of f-strings for assembling URL string conversions

bdraco:
I expect this to be almost the same, but happy to be wrong about that


bdraco:
Glad to get validation that f-strings were the way to go, and so much more readable as well

webknjaz:
IIRC when f-strings were being designed in CPython, they made a conscious choice/effort to make them faster than any other way.

bdraco:
That was my expectation, but always good to recheck assumptions in case things have changed.

PR #1248, opened 2024-10-13
Avoid normalizing child paths when there are no dots in the path

bdraco:
Avoid normalizing child paths when there are no dots in the path

This one will need quite a few more tests if benchmarks shows its worth working on

### DIFF
--- a/CHANGES/1248.misc.rst
+++ b/CHANGES/1248.misc.rst
@@ -1,1 +1,1 @@
-Improved performance of :meth:`URL.joinpath() <yarl.URL.joinpath> -- by :user:`bdraco`.
+Improved performance of :py:meth:`~yarl.URL.joinpath` -- by :user:`bdraco`.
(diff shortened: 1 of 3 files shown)


yarl/_url.py
```diff
@@ -957,6 +958,7 @@ def _make_child(self, paths: "Sequence[str]", encoded: bool = False) -> "URL":
                     f"Appending path {path!r} starting from slash is forbidden"
                 )
             path = path if encoded else self._PATH_QUOTER(path)
+            needs_normalize |= "." in path
             segments = list(reversed(path.split("/")))
```
bdraco, line 962: This could call `.reverse()` instead and avoid making a copy

bdraco: There are a few other small cleanups that could be done here. But for another PR

CHANGES/1248.misc.rst
```diff
@@ -0,0 +1 @@
+Improved performance of :meth:`URL.joinpath() <yarl.URL.joinpath> -- by :user:`bdraco`.
```
bdraco, line 1:
```suggestion
Improved performance of :py:meth:`~yarl.URL.joinpath` -- by :user:`bdraco`.
```

PR #1251, opened 2024-10-13
Avoid some string copy in extend_query

bdraco:
Avoid some string copy in extend_query

Performance is unchanged, but it is less code


bdraco:
Looks like we are missing some benchmarks for common extend_query cases

PR #1252, opened 2024-10-13
Check val.netloc instead of self.absolute internally

bdraco:
val.netloc is accessed on a namedtuple, which is much a bit faster than calling the property (even the cython version)

The bechmark only shows ~4% improvement so not worth of a changelog message


bdraco:
We have no benchmarks for `is_default_port` or equality

PR #1255, opened 2024-10-13
Use a dispatch dict to determine how to construct query strings

bdraco:


bdraco:
If benchmark shows its worth doing, will need to add some tests for subclassed `list`, `dict`, `tuple`

bdraco:
Its faster, but not by a lot. I'd say the performance improvement isn't worth all the new code

PR #1256, opened 2024-10-13
Optimize building query strings with MultiDict

bdraco:
Most of the time aiohttp is passing in a MulitDict here.  This will also help a bit for `dict` without Sequence, but as not a common of a case

For historical reasons we have support for sequences inside of MulitDict even though it shouldn't happen and it would be a breaking change to remove it


bdraco:
Might not be worth it, but benchmarks should show either way

PR #1258, opened 2024-10-13
Align Python and C quoter to match type signature of not allowing None

bdraco:
We never pass `None` to the quoters, they should not allow it.

The quoters are not public methods so this is an internal change only.

### DIFF
--- a/yarl/_quoting_c.pyx
+++ b/yarl/_quoting_c.pyx
@@ -203,14 +203,14 @@ cdef class _Quoter:
 
     def __call__(self, val):
         cdef Writer writer
-        if val is None:
-            return None
         if type(val) is not str:
             if isinstance(val, str):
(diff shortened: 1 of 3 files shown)


yarl/_quoting_c.pyx
```diff
@@ -203,14 +203,14 @@ cdef class _Quoter:
...
-        if val is None:
-            return None
         if type(val) is not str:
             if isinstance(val, str):
                 # derived from str
                 val = str(val)
             else:
                 raise TypeError("Argument should be str")
+        if not val:
+            return val
```
bdraco, line 213: py quoter does this, not sure if its worth it for the c quoter, but benchmark will tell us

bdraco:
Looks like its already well optimized and not worth changing

PR #1264, opened 2024-10-15
Improve performance of constructing URLs from unencoded string

bdraco:
## What do these changes do?

Improve performance of constructing URLs with a netloc

This is a ~3% speed up to the common case (note that codspeed reporting threshold if 8%)

no

So much of the request itself is constructing the URL, however I'm running out of thoughts on how to improve the performance here.

Screenshot 2024-10-14 at 10 38 15 PM

new

### DIFF
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -273,7 +273,7 @@ class URL:
             if not netloc:  # netloc
                 host = ""
             else:
-                username, password, host, port = cls._split_netloc(val[1])
+                username, password, host, port = cls._split_netloc(netloc)
                 if host is None:
                     if scheme in SCHEME_REQUIRES_HOST:
                         msg = (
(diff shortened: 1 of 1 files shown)


yarl/_url.py
```diff
@@ -390,7 +391,8 @@ def build(
             if path and netloc:
                 if "." in path:
                     path = cls._normalize_path(path)
-                cls._validate_authority_uri_abs_path(host, path)
+                if path[-1] != "/":
```
bdraco, line 394:
```suggestion
                if path[0] != "/":
```

yarl/_url.py
```diff
@@ -298,7 +298,8 @@ def __new__(
                 if netloc:
                     if "." in path:
                         path = cls._normalize_path(path)
-                    cls._validate_authority_uri_abs_path(host, path)
+                    if path[-1] != "/":
```
bdraco, line 301:
```suggestion
                    if path[0] != "/":
```

yarl/_url.py
```diff
@@ -254,10 +254,10 @@ def __new__(
     ) -> Self:
         if strict is not None:  # pragma: no cover
             warnings.warn("strict parameter is ignored")
-        if type(val) is cls:
-            return val
         if type(val) is str:
             val = urlsplit(val)
+        elif type(val) is cls:
+            return val
         elif type(val) is SplitResult:
```
bdraco, line 261: This is much more rare now that we have `cls._from_val`

bdraco:
not a significant enough speed up to get a changelog message

PR #1265, opened 2024-10-15
Fix `URL.build` not validating that `path` has a leading slash when passing `authority`

bdraco:
## What do these changes do?

Fix `URL.build` not validating that path has a leading slash when passing `authority`
The validation only worked when `host` was passed

Validation of leading `/` in path now works for `authority` instead of just `host` as intended

### DIFF
--- a/CHANGES/1265.bugfix.rst
+++ b/CHANGES/1265.bugfix.rst
@@ -1,0 +1,3 @@
+Fixed :py:meth:`~yarl.URL.build` failing to validate paths must start with a / when passing ``authority`` -- by :user:`bdraco`.
+
+The validation only worked correctly when passing ``host``.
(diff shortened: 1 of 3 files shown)


CHANGES/1265.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+Fixed :py:meth:`~yarl.URL.build` failing to validate paths must start with a / when passing ``authority`` -- by :user:`bdraco`.
```
bdraco, line 1:
```suggestion
Fixed :py:meth:`~yarl.URL.build` failing to validate paths must start with a ``/`` when passing ``authority`` -- by :user:`bdraco`.
```

bdraco:
Could use another set of 👀 to make sure `authority` is supposed to have this validation. Would be very surprised if not since `authority` is in the validation message but maybe I'm missing something....

bdraco:
slept on this one and much more confident the change is correct.

PR #1267, opened 2024-10-15
DNM: Test performance change to encoding URL from str

bdraco:


bdraco:
performance change is within the margin of error

PR #1275, opened 2024-10-15
Improve performance of calling origin() for URLs with credentials

### LINKED ISSUE #1272: Unneeded host re encode in origin
https://github.com/aio-libs/yarl/blob/598cc0c5eda673f97690c1d466d4c9ded8d10b48/yarl/_url.py#L589

this could use the same host sub component solution used elsewhere 

also netloc building is overkill if port is not set 

this one is worth optimizing for sure since it’s used in the cookie loop in aiohttp

bdraco:
Calling `SplitResult.port` and `SplitResult.hostname` has to re-parse the net location.  Use our internal properties instead since its very likely to already be parsed.  The speed up is tiny but it removes another case where we call `_encode_host` which makes the flow a bit easier to manage because every time we call `_encode_host` we need to think about how its going to be mutated.

#1175 did not improve the credentials/auth case as much as it could

closes #1272

related issue #1174

### DIFF
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -584,12 +584,11 @@ class URL:
         user, password, path, query and fragment are removed.
         """
         v = self._val
-        netloc = v.netloc
-        if not netloc:
+        if not v.netloc:
             raise ValueError("URL should be absolute")
         if not v.scheme:
             raise ValueError("URL should have scheme")
-        if "@" not in netloc:
+        if "@" not in v.netloc:
             val = v._replace(path="", query="", fragment="")
         else:
             encoded_host = self.host_subcomponent
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -584,15 +584,16 @@ class URL:
         user, password, path, query and fragment are removed.
         """
         v = self._val
-        if not v.netloc:
+        netloc = v.netloc
+        if not netloc:
             raise ValueError("URL should be absolute")
         if not v.scheme:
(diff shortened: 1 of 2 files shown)


bdraco:
its a really tiny speed up but gets rid of another encode_host call

yarl/_url.py
```diff
@@ -584,15 +584,16 @@ def _origin(self) -> "URL":
         user, password, path, query and fragment are removed.
         """
         v = self._val
-        if not v.netloc:
+        netloc = v.netloc
+        if not netloc:
```
bdraco, line 588:
```suggestion
        if not v.netloc:
```

yarl/_url.py
```diff
@@ -584,15 +584,16 @@ def _origin(self) -> "URL":
...
             raise ValueError("URL should be absolute")
         if not v.scheme:
             raise ValueError("URL should have scheme")
-        if "@" not in v.netloc:
+        if "@" not in netloc:
```
bdraco, line 592:
```suggestion
        if "@" not in v.netloc:
```

yarl/_url.py
```diff
@@ -584,15 +584,16 @@ def _origin(self) -> "URL":
         user, password, path, query and fragment are removed.
         """
         v = self._val
-        if not v.netloc:
+        netloc = v.netloc
```
bdraco, line 587:
```suggestion
```

PR #1276, opened 2024-10-15
DNM: benchmark using __dict__

bdraco:
__dict__ shares keys between instances so it may have a bit of a different profile

downside is set is not blocked.. but we could make another version


bdraco:
good to know the __dict__ shared keys doesn't help us

PR #1279, opened 2024-10-15
Make validate_host a required argument to _encode_host

bdraco:
Internal change only

I made a mistake here previously that thankfully never made it out the door... 

Make passing a value required so we always have to think about what the correct option is.  Additionally most of these calls have been eliminated because the only time we need to validate is on user input, and previously we had cases where we would validate when we already had a internally validated value.

### DIFF
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -1034,7 +1034,7 @@ class URL:
         return False, lower_host, is_ascii, "", "", ""
 
     @classmethod
-    def _encode_host(cls, host: str, validate_host: bool) -> str:
+    def _encode_host(cls, host: str, validate: bool) -> str:
         """Encode host part of URL."""
         looks_like_ip, lower_host, is_ascii, raw_ip, sep, zone = cls._parse_host(host)
         if looks_like_ip:
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -591,7 +591,9 @@ class URL:
         if "@" not in v.netloc:
             val = v._replace(path="", query="", fragment="")
         else:
-            encoded_host = self._encode_host(v.hostname, validate_host=False) if v.hostname else ""
+            encoded_host = (
+                self._encode_host(v.hostname, validate_host=False) if v.hostname else ""
+            )
             netloc = self._make_netloc(None, None, encoded_host, v.port)
             val = v._replace(netloc=netloc, path="", query="", fragment="")
         return self._from_val(val)
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -1226,7 +1226,7 @@ class URL:
             raise ValueError("host replacement is not allowed for relative URLs")
         if not host:
             raise ValueError("host removing is not allowed")
-        encoded_host = self._encode_host(host, True) if host else ""
+        encoded_host = self._encode_host(host, validate_host=True) if host else ""
         port = self.explicit_port
         netloc = self._make_netloc(self.raw_user, self.raw_password, encoded_host, port)
         return self._from_val(val._replace(netloc=netloc))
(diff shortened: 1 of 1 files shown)


yarl/_url.py
```diff
@@ -1034,7 +1034,7 @@ def _parse_host(
         return False, lower_host, is_ascii, "", "", ""
 
     @classmethod
-    def _encode_host(cls, host: str, validate_host: bool = True) -> str:
+    def _encode_host(cls, host: str, validate_host: bool) -> str:
```
bdraco, line 1037:
```suggestion
    def _encode_host(cls, host: str, validate: bool) -> str:
```

yarl/_url.py
```diff
@@ -1034,7 +1034,7 @@ def _parse_host(
         return False, lower_host, is_ascii, "", "", ""
 
     @classmethod
-    def _encode_host(cls, host: str, validate_host: bool = True) -> str:
+    def _encode_host(cls, host: str, validate: bool) -> str:
```
bdraco, line 1037:
```suggestion
    def _encode_host(cls, host: str, validate_host: bool) -> str:
```

yarl/_url.py
bdraco, line 1037: I was trying to figure out how to move `validate_host` out of here but its tied to `isascii` so its hard to untangle

yarl/_url.py
```diff
@@ -389,9 +389,9 @@ def build(
             _host: Union[str, None] = None
             if authority:
                 user, password, _host, port = cls._split_netloc(authority)
-                _host = cls._encode_host(_host, validate_host=False) if _host else ""
+                _host = cls._encode_host(_host, False) if _host else ""
```
bdraco, line 392:
```suggestion
                _host = cls._encode_host(_host, validate_host=False) if _host else ""
```

yarl/_url.py
```diff
@@ -288,7 +288,7 @@ def __new__(
                         raise ValueError(msg)
                     else:
                         host = ""
-                host = cls._encode_host(host, validate_host=False)
+                host = cls._encode_host(host, False)
```
bdraco, line 291:
```suggestion
                host = cls._encode_host(host, validate_host=False)
```

yarl/_url.py
```diff
@@ -389,9 +389,9 @@ def build(
             _host: Union[str, None] = None
             if authority:
                 user, password, _host, port = cls._split_netloc(authority)
-                _host = cls._encode_host(_host, validate_host=False) if _host else ""
+                _host = cls._encode_host(_host, False) if _host else ""
             elif host:
-                _host = cls._encode_host(host)
+                _host = cls._encode_host(host, True)
```
bdraco, line 394:
```suggestion
                _host = cls._encode_host(host, validate_host=True)
```

yarl/_url.py
```diff
@@ -591,7 +591,7 @@ def _origin(self) -> "URL":
         if "@" not in v.netloc:
             val = v._replace(path="", query="", fragment="")
         else:
-            encoded_host = self._encode_host(v.hostname) if v.hostname else ""
+            encoded_host = self._encode_host(v.hostname, False) if v.hostname else ""
```
bdraco, line 594:
```suggestion
            encoded_host = self._encode_host(v.hostname, validate_host=False) if v.hostname else ""
```

yarl/_url.py
```diff
@@ -1226,7 +1226,7 @@ def with_host(self, host: str) -> "URL":
             raise ValueError("host replacement is not allowed for relative URLs")
         if not host:
             raise ValueError("host removing is not allowed")
-        encoded_host = self._encode_host(host) if host else ""
+        encoded_host = self._encode_host(host, True) if host else ""
```
bdraco, line 1229:
```suggestion
        encoded_host = self._encode_host(host, validate_host=True) if host else ""
```

PR #1280, opened 2024-10-15
Rework disabling hypothesis tests in cibuildwheel

### LINKED ISSUE #1262: Remove the need to set ENABLE_HYPOTHESIS_TESTS in tests
https://github.com/aio-libs/yarl/blob/b64f2318df91e295f6fc4eb706794361890ad83d/.github/workflows/ci-cd.yml#L314

> https://cibuildwheel.pypa.io/en/stable/options/#overrides would've solved the problem better, FYI. You could override the pytest command with one including -m 'not hypothesis' (having configured the corresponding marker in the tests).

suggestion by webknjaz

bdraco:
fixes #1262

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -74,7 +74,7 @@ before-test = [
   "PIP_CONSTRAINT= pip install PyYAML",
 ]
 test-requires = "-r requirements/test.txt"
-test-command = "pytest -v --no-cov {project}/tests"
+test-command = "pytest -v -m 'not hypothesis' --no-cov {project}/tests"
 # don't build PyPy wheels, install from source instead
 skip = "pp*"
 
--- a/tests/test_quoting.py
+++ b/tests/test_quoting.py
@@ -1,5 +1,3 @@
-import os
-
 import pytest
 from hypothesis import assume, example, given, note
 from hypothesis import strategies as st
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -74,7 +74,7 @@ before-test = [
   "PIP_CONSTRAINT= pip install PyYAML",
 ]
 test-requires = "-r requirements/test.txt"
-test-command = "pytest -v --no-cov {project}/tests"
+test-command = 'pytest -v -m "not hypothesis" --no-cov {project}/tests'
```
webknjaz, line 77: @bdraco I thought we'd do this granularly, using an override that selects musl..

tests/test_quoting.py
```diff
@@ -473,18 +469,14 @@ def test_unquoter_path_with_plus(unquoter):
...
 def test_fuzz__PyQuoter(safe, protected, qs, requote):
     """Verify that _PyQuoter can be instantiated with any valid arguments."""
     assert _PyQuoter(safe=safe, protected=protected, qs=qs, requote=requote)
 
 
 @given(ignore=st.text(), unsafe=st.text(), qs=st.booleans())
-@pytest.mark.skipif(
-    not ENABLE_HYPOTHESIS_TESTS, reason="hypothesis tests are not enabled"
-)
+@pytest.mark.hypothesis
```
webknjaz, line 479: @bdraco apparently, this marker is already being autoapplied:



*(https://hypothesis.rtfd.io/en/latest/details.html#the-hypothesis-pytest-plugin)*

pyproject.toml
bdraco: Anything that is using qemu nearly an order of magnitude slower, and since the hypothesis are not deterministic their performance profile changes between executions and sometimes its even worse so I figure since we already run them on all the test platforms, running them during the wheel builds as well wasn't strictly needed.

tests/test_quoting.py
bdraco: Nice, that makes this a bit cleaner

pyproject.toml
webknjaz: I think we could still run them on non-emulated  platforms since it extends our environment coverage. But that's something I can look into later. No rush.

PR #1284, opened 2024-10-15
Include port in the fast path for building the network location

bdraco:
Followup to #1271 

We can include port in the fast path as well do avoid doing a full netloc assemble when we know the username and password are None


bdraco:
looks like its not worth it

PR #1285, opened 2024-10-15
DNM: Add noexcept to places where the quoter cannot throw

bdraco:


bdraco:
incorporated into the other quoter PR since it made more sense there

PR #1286, opened 2024-10-15
DNM: Avoid malloc if there is no thing to quote

bdraco:
Probably quite a bit faster if there is nothing to quote

But a bit slower if there is because we have it iterate twice


bdraco:
https://github.com/aio-libs/yarl/pull/1288 is a better choice

PR #1288, opened 2024-10-15
Improve quoter performance when all characters are safe

bdraco:
Avoid allocating the writer when we will end up not altering the string.

For most operations this will be the common case.

This will be a bit of a trade off as it means we have to check to see if we should encode or not but given we usually don't it's going to be faster for the no need to encode case

### DIFF
--- a/yarl/_quoting_c.pyx
+++ b/yarl/_quoting_c.pyx
@@ -202,7 +202,6 @@ cdef class _Quoter:
             set_bit(self._protected_table, ch)
 
     def __call__(self, val):
-        cdef Writer writer
         if val is None:
             return None
         if type(val) is not str:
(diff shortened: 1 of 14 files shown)


bdraco:
The benchmarks are not currently sufficient to justify the change either way.  Will need to do a lot more production testing to consider the trade offs

bdraco:
Pct has a 5% regression still. We can likely use a faster loop

yarl/_quoting_c.pyx
```diff
@@ -211,23 +217,55 @@ cdef class _Quoter:
...
+        cdef Py_ssize_t idx = 0
 
         while idx < length:
-            ch = val[idx]
+            ch = PyUnicode_READ(kind, data, idx)
             idx += 1
             if ch == '%' and self._requote and idx <= length - 2:
-                ch = _restore_ch(val[idx], val[idx + 1])
+                ch = _restore_ch(
+                    PyUnicode_READ(kind, data, idx),
```
bdraco, line 266: Should change unquote to use `PyUnicode_READ` as well

PR #1292, opened 2024-10-16
Use faster Python unicode API for unquoter

bdraco:
Every character that was being iterated during unquote was calling `__Pyx_GetItemInt_Unicode`

Its a bit faster to use `PyUnicode_READ` instead


bdraco:
We don't have any benchmarks for accessing the properties that need to go through the unquoter

bdraco:
could be faster if no change as well

PR #1294, opened 2024-10-16
Fix path quoting benchmarks to use the same quoting config as production

bdraco:
These benchmarks should have used `PATH_QUOTER`, and since they didn't they appear to be much more expensive than a real world use case.


bdraco:
Improvement isn't real, its only fixing the benchmark case to better match the real world

PR #1296, opened 2024-10-16
Improve performance of handling empty query strings

bdraco:
Its rather common to call `URL.query` or `URL.query_string` to see if a key exists when there is no actual query string. Avoid doing the work to construct objects and unquote the query string if there is no query string.

### DIFF
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -820,7 +820,8 @@ class URL:
     @cached_property
     def _parsed_query(self) -> list[tuple[str, str]]:
         """Parse query part of URL."""
-        return parse_qsl(self._val.query, keep_blank_values=True)
+        qs = self._val.query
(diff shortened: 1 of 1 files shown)


yarl/_url.py
```diff
@@ -820,7 +820,8 @@ def path_safe(self) -> str:
     @cached_property
     def _parsed_query(self) -> list[tuple[str, str]]:
         """Parse query part of URL."""
-        return parse_qsl(self._val.query, keep_blank_values=True)
+        qs = self._val.query
+        return parse_qsl(qs, keep_blank_values=True) if qs else []
 
     @cached_property
     def query(self) -> "MultiDictProxy[str]":
```
bdraco, line 827: This should probably return an single empty object for cases where its empty

bdraco:
Doesn't seem worth it

PR #1297, opened 2024-10-16
Improve performance of `URL.build` by avoiding operations

bdraco:
`URL.build` had a few places left where performance could be improved

- Avoid evaluating the run time lambda created at https://github.com/python/cpython/blob/d83fcf8371f2f33c7797bc8f5423a8bca8c46e5c/Lib/collections/__init__.py#L441 
- Avoid creating the `URL` object twice when passing `query` as we already had a method to convert `query` to string, and we shouldn't make two objects to build a URL

### DIFF
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -326,6 +326,10 @@ class URL:
                 or val.query != query
                 or val.fragment != fragment
             ):
+                # Constructing the tuple directly to avoid the overhead of
+                # the lambda and arg processing since NamedTuples are constructed
+                # with a run time built lambda
+                # https://github.com/python/cpython/blob/d83fcf8371f2f33c7797bc8f5423a8bca8c46e5c/Lib/collections/__init__.py#L441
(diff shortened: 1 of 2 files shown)


yarl/_url.py
```diff
@@ -416,9 +418,14 @@ def build(
...
 
-        url = cls._from_val(SplitResult(scheme, netloc, path, query_string, fragment))
         if query:
-            return url.with_query(query)
+            query_string = cls._get_str_query(query) or ""
+
+        url = object.__new__(cls)
+        url._val = tuple.__new__(
+            SplitResult, (scheme, netloc, path, query_string, fragment)
+        )
```
bdraco, line 427: this is non-obvious why its much faster.

Could use a comment to point to https://github.com/python/cpython/blob/d83fcf8371f2f33c7797bc8f5423a8bca8c46e5c/Lib/collections/__init__.py#L441

PR #1300, opened 2024-10-16
Fix query benchmarks to avoid testing property cache hits

bdraco:
Ideally we would look at `__wrapped__` but its not exposed on `under_cached_property,` but this is fine as it will give us a baseline to know when it changes since every call will be a cache miss.

Will give us a better idea on the actual performance impact of #1296


bdraco:
decrease is not real, its a benchmark change

PR #1304, opened 2024-10-17
Small cleanups to making children paths

bdraco:
## What do these changes do?

This function was a bit hard to unpack so I have added some comments about what is going on.

Since we reconstruct nearly everything, create a new tuple instead of replacing


bdraco:
Didn't expect it to speed up that much.  Should do join as well

bdraco:
I couldn't believe `._replace` was so expensive, but it is

bdraco:
`origin`, `join`, and `with_path` are probably good candidates to rework to avoid `_replace`. The others are likely not called enough that they are worth optimizing

PR #1306, opened 2024-10-17
Improve performance of the `URL.origin` method

bdraco:
Improve performance of origin


bdraco:
Looks like origin benchmark isn't good because its testing the cache

bdraco:
The benchmark could still be better because codspeed is re-running it again to avoid jitter which still hits the cache. The actual improvement is better than the 4% its showing.  Will need to think about how to improve the benchmark some more, but no reason to wait for a solution as this is clearly faster for all the other cases as well

PR #1309, opened 2024-10-17
Improve performance of the `URL.update_query` method

bdraco:
Improve performance of the `URL.update_query` method


bdraco:
Well nothing improved because we are missing benchmarks for `update_query`

PR #1310, opened 2024-10-17
Fix URL.origin benchmark to avoid testing the cache

bdraco:
Since origin() is a thin shim around _origin we need to create new URLs for the benchmark to avoid
testing the cache


bdraco:
Performance change is not real, only a chance in how the benchmark works

PR #1311, opened 2024-10-17
Improve query string benchmarking

bdraco:
- Add benchmarks for clearing the query string using with_query
- Add a benchmark for subclassed strings

### DIFF
--- a/tests/test_url_benchmarks.py
+++ b/tests/test_url_benchmarks.py
@@ -226,25 +226,39 @@ def test_url_make_with_ipv6_address_and_path(benchmark: BenchmarkFixture) -> Non
             URL("http://[::1]/req")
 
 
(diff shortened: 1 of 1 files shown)


tests/test_url_benchmarks.py
```diff
@@ -226,27 +230,51 @@ def _run() -> None:
...
+
+def test_with_query_mapping(benchmark: BenchmarkFixture) -> None:
     @benchmark
     def _run() -> None:
         for _ in range(25):
             BASE_URL.with_query(SIMPLE_QUERY)
 
 
-def test_url_make_with_int_query_mapping(benchmark: BenchmarkFixture) -> None:
+def test_with_int_query_mapping(benchmark: BenchmarkFixture) -> None:
```
bdraco, line 250:
```suggestion
def test_with_query_mapping_int_values(benchmark: BenchmarkFixture) -> None:
```

PR #1312, opened 2024-10-17
Small performance improvement for query strings with int values

bdraco:
Moves the `is` check for `int` before the `issubclass` call since its very cheap. It should be a tiny bit slower for subclassed `str` but since `is` is so much faster than `issubclass` it should be de minimis


bdraco:
Its ~3% faster

PR #1317, opened 2024-10-17
Improve performance of with_fragment

bdraco:
Avoid calling `SplitResult._replace` since its much slower


bdraco:
We don't have any tests that actually alter the fragment

PR #1321, opened 2024-10-18
Improve performance of `URL.parent`

bdraco:
Avoid calling `SplitResult._replace` since its much slower, and instead replace with fast `NamedTuple` creation `tuple.__new__(Type, (...)`

### DIFF
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -904,7 +904,6 @@ class URL:
         fragment.
 
         """
-        path = self.raw_path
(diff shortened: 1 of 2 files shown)


bdraco:
We have some missing tests for this.  Need to add them first

yarl/_url.py
```diff
@@ -905,12 +905,14 @@ def parent(self) -> "URL":
 
         """
         path = self.raw_path
```
bdraco, line 907:
```suggestion
```

PR #1336, opened 2024-10-19
Small performance improvement for constructng new URL objects

bdraco:
Avoids the NamedTuple property accesses since we already have the values


bdraco:
oops. typed the title `constructng` -> `constructing`

PR #1340, opened 2024-10-20
Support using the subtraction operator to get the relative path between URLs

### LINKED ISSUE #1183: Add Support for Relative Path Calculation

_No response_

#### Current Behavior

#### Proposed Feature

##### Example Usage
```python
from yarl import URL

```

### Additional context

_No response_

### Code of Conduct

oleksbabieiev:
## What do these changes do?

Support for using the subtraction operator to calculate the difference between the *paths* of two URLs. A new URL with the relative path is returned to the user.

Usage:
```py
from yarl import URL

target = URL("http://example.com/path/index.html")
base = URL("http://example.com/path/")

rel = target - base

print(rel)  # output: "index.html"
```

Now the user can easily calculate the relative path between two URLs!

## Related issue number

Resolves #1183

## Checklist

### DIFF
--- a/CHANGES/1340.feature.rst
+++ b/CHANGES/1340.feature.rst
@@ -1,0 +1,2 @@
+Added support for using the :meth:`subtraction operator <yarl.URL.__sub__>`
+to get the relative path between URLs.
--- a/CHANGES/1340.feature.rst
+++ b/CHANGES/1340.feature.rst
@@ -1,2 +1,2 @@
 Added support for using the :meth:`subtraction operator <yarl.URL.__sub__>`
-to get the relative path between URLs.
+to get the relative path between URLs -- by :user:`oleksbabieiev`.
--- a/CHANGES/1340.feature.rst
+++ b/CHANGES/1340.feature.rst
@@ -1,7 +1,7 @@
 Added support for using the :meth:`subtraction operator <yarl.URL.__sub__>`
 to get the relative path between URLs.
 
-Note that both URLs must have the same scheme and netloc:
+Note that both URLs must have the same scheme, user, password, host and port:
 
 .. code-block:: pycon
 
@@ -10,7 +10,7 @@ Note that both URLs must have the same scheme and netloc:
    >>> target - base
    URL('path/index.html')
 
-URLs can also be relative and have no scheme or netloc at all:
+URLs can also be relative:
 
 .. code-block:: pycon
 
--- a/CHANGES/1340.feature.rst
+++ b/CHANGES/1340.feature.rst
@@ -1,2 +1,22 @@
 Added support for using the :meth:`subtraction operator <yarl.URL.__sub__>`
-to get the relative path between URLs -- by :user:`oleksbabieiev`.
+to get the relative path between URLs.
+
+Note that both URLs must have the same scheme and netloc:
+
+.. code-block:: pycon
+
+   >>> target = URL("http://example.com/path/index.html")
+   >>> base = URL("http://example.com/")
+   >>> target - base
+   URL('path/index.html')
+
+URLs can also be relative and have no scheme or netloc at all:
+
+.. code-block:: pycon
+
+  >>> target = URL("/")
+  >>> base = URL("/path/index.html")
+  >>> target - base
+  URL('..')
+
+-- by :user:`oleksbabieiev`.
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -976,7 +976,7 @@ The path is encoded if needed.
 
 The subtraction (``-``) operator creates a new URL with
 a relative *path* to the target URL from the given base URL.
-*scheme*, *user*, *password*, *host* and *port* are removed.
+*scheme*, *user*, *password*, *host*, *port*, *query* and *fragment* are removed.
 
 .. method:: URL.__sub__(url)
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -974,6 +974,21 @@ The path is encoded if needed.
          >>> base.join(URL('//python.org/page.html'))
          URL('http://python.org/page.html')
 
+The subtraction (``-``) operator creates a new URL with
+a relative *path* to the target URL from the given base URL.
+*scheme*, *user*, *password*, *host* and *port* are removed.
+
+.. method:: URL.__sub__(url)
+
+   Returns a new URL with a relative *path* between two other URL objects.
+
+   .. doctest::
+
+      >>> target = URL('http://example.com/path/index.html')
+      >>> base = URL('http://example.com/')
+      >>> target - base
+      URL('path/index.html')
+
 Human readable representation
 -----------------------------
 
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -70,6 +70,14 @@ def test_str():
         ("http://example.com/path", "http://example.com/path/to/", ".."),
         ("http://example.com/", "http://example.com/", "."),
         ("http://example.com", "http://example.com", "."),
+        ("http://example.com/", "http://example.com", "."),
+        ("http://example.com", "http://example.com/", "."),
+        ("//example.com", "//example.com", "."),
+        ("/path/to", "/spam/", "../path/to"),
+        ("path/to", "spam/", "../path/to"),
+        ("path/to", "spam", "path/to"),
+        ("..", ".", ".."),
+        (".", "..", "."),
     ],
 )
 def test_sub(target: str, base: str, expected: str):
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -53,6 +53,33 @@ def test_str():
     assert str(url) == "http://example.com:8888/path/to?a=1&b=2"
 
 
+@pytest.mark.parametrize(
+    "target,base,expected",
+    [
+        ("http://example.com/path/to", "http://example.com/", "path/to"),
+        ("http://example.com/path/to", "http://example.com/spam", "path/to"),
+        ("http://example.com/path/to", "http://example.com/spam/", "../path/to"),
+        ("http://example.com/path", "http://example.com/path/to/", ".."),
+        ("http://example.com/", "http://example.com/", "."),
+        ("http://example.com", "http://example.com", "."),
+    ],
+)
+def test_sub(target: str, base: str, expected: str):
+    assert URL(target) - URL(base) == URL(expected)
+
+
+def test_sub_with_different_schemes():
+    with pytest.raises(ValueError) as ctx:
+        URL("http://example.com/") - URL("https://example.com/")
+    assert "Both URLs should have the same scheme" == str(ctx.value)
+
+
+def test_sub_with_different_netlocs():
+    with pytest.raises(ValueError) as ctx:
+        URL("https://spam.com/") - URL("https://ham.com/")
+    assert "Both URLs should have the same netloc" == str(ctx.value)
+
+
 def test_repr():
     url = URL("http://example.com")
     assert "URL('http://example.com')" == repr(url)
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -62,7 +62,7 @@ def test_str():
 
 
 @pytest.mark.parametrize(
-    "target,base,expected",
+    ("target", "base", "expected"),
     [
         ("http://example.com/path/to", "http://example.com/", "path/to"),
         ("http://example.com/path/to", "http://example.com/spam", "path/to"),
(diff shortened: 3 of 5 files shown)


bdraco:
Looks like the tests aren't passing in the CI

CHANGES/1340.feature.rst
```diff
@@ -0,0 +1,2 @@
+Added support for using the :meth:`subtraction operator <yarl.URL.__sub__>`
+to get the relative path between URLs.
```
bdraco, line 2:
```suggestion
Added support for using the :meth:`subtraction operator <yarl.URL.__sub__>` to get the relative path between URLs -- by :user:`oleksbabieiev`
```

tests/test_url.py
```diff
@@ -61,6 +61,33 @@ def test_str():
...
+    with pytest.raises(ValueError) as ctx:
+        URL("http://example.com/") - URL("https://example.com/")
+    assert "Both URLs should have the same scheme" == str(ctx.value)
+
+
+def test_sub_with_different_netlocs():
+    with pytest.raises(ValueError) as ctx:
+        URL("https://spam.com/") - URL("https://ham.com/")
+    assert "Both URLs should have the same netloc" == str(ctx.value)
+
```
bdraco, line 89: Will also need some tests for absolute and relative urls, and paths with `..` and `.` for relative urls

yarl/_url.py
```diff
@@ -479,6 +480,34 @@ def __truediv__(self, name: str) -> "URL":
...
+        """A wrapper over os.path.relpath()"""
+
+        if not path:
+            path = "/"
+        if not start:
+            start = "/"
+        if not start.endswith("/"):
+            start = dirname(start)
+
+        return relpath(path, start)
```
bdraco, line 509: The helper can be moved to `_path.py` now

bdraco:
Sorry about the conflicts. I was doing a bit of housekeeping to better organize internals. I think I've worked them out correctly.

oleksbabieiev:
I have also noticed that when using `os.path.relpath()` with absolute and relative paths (e.g. `dir` and `/`), the program first gets the current working directory (essentially doing `os.getcwd()`). Otherwise, it is obviously impossible to calculate the path between the given directories. This behavior is undesirable, so [I have disallowed it](https://github.com/aio-libs/yarl/pull/1340/commits/20060668076cf662fc416030d2d1f35a8bbeea48).

yarl/_path.py
```diff
@@ -31,11 +34,32 @@ def normalize_path_segments(segments: Sequence[str]) -> list[str]:
 def normalize_path(path: str) -> str:
     # Drop '.' and '..' from str path
     prefix = ""
-    if path and path[0] == "/":
+    if path and path[0] == SEPARATOR:
         # preserve the "/" root element of absolute paths, copying it to the
         # normalised output as per sections 5.2.4 and 6.2.2.3 of rfc3986.
-        prefix = "/"
+        prefix = SEPARATOR
```
bdraco, line 40: Please leave the these as is. It's more confusing to have them as a name constant especially since the url spec has multiple different separators.

CHANGES/1340.feature.rst
webknjaz, line 1: Could you add a small demo similar to this entry https://yarl.aio-libs.org/en/latest/changes/#id44 ?

(example: https://github.com/aio-libs/yarl/pull/1082/files#diff-ec5930c2f344dfd8fb4d7e1b1d21cd6ac87144168e5942fb58021b7e08e76100R9-R23)

tests/test_url.py
```diff
@@ -61,6 +61,52 @@ def test_str():
     assert str(url) == "http://example.com:8888/path/to?a=1&b=2"
 
 
+@pytest.mark.parametrize(
+    "target,base,expected",
```
webknjaz, line 65: nit:
```suggestion
    ("target", "base", "expected"),
```

tests/test_url.py
```diff
@@ -61,6 +61,52 @@ def test_str():
...
+def test_sub_with_different_schemes():
+    with pytest.raises(ValueError) as ctx:
+        URL("http://example.com/") - URL("https://example.com/")
+    assert "Both URLs should have the same scheme" == str(ctx.value)
+
+
+def test_sub_with_different_netlocs():
+    with pytest.raises(ValueError) as ctx:
+        URL("https://spam.com/") - URL("https://ham.com/")
+    assert "Both URLs should have the same netloc" == str(ctx.value)
```
webknjaz, line 96: Things like `pytest.raises()` should always specify a regex for the message:
```suggestion
    expected_error_msg = r"^Both URLs should have the same netloc$"
    with pytest.raises(ValueError, match=expected_error_msg):
        URL("https://spam.com/") - URL("https://ham.com/")
```

tests/test_url.py
```diff
@@ -61,6 +61,52 @@ def test_str():
...
+
+
+def test_sub_with_abs_and_rel_paths():
+    with pytest.raises(ValueError) as ctx:
```
webknjaz, line 100: Same here: plz add a `match=`.

yarl/_path.py
```diff
@@ -39,3 +40,24 @@ def normalize_path(path: str) -> str:
 
     segments = path.split("/")
     return prefix + "/".join(normalize_path_segments(segments))
+
+
+def relative_path(path: str, start: str) -> str:
+    """A wrapper over os.path.relpath()"""
```
webknjaz, line 46: Please use a verb to describe what the function does instead of a reference to some unidentified external location. This will help reduce the mental overhead for the readers.

yarl/_path.py
```diff
@@ -39,3 +40,24 @@ def normalize_path(path: str) -> str:
...
+    if not path:
+        path = "/"
+    if not start:
+        start = "/"
+    if not start.endswith("/"):
+        start = dirname(start)
+
+    if (path.startswith("/") and not start.startswith("/")) or (
+        not path.startswith("/") and start.startswith("/")
+    ):
```
webknjaz, line 57: Let's try to explain the semantics of the check by labeling them:
```suggestion
    path_has_leading_slash = path.startswith("/")
    start_has_leading_slash = start.startswith("/")
    both_have_leading_slash = all((path_has_leading_slash, start_has_leading_slash))
    none_have_leading_slash = all((not path_has_leading_slash, not start_has_leading_slash))
    if not both_have_leading_slash and not none_have_leading_slash:
```

yarl/_url.py
```diff
@@ -475,6 +475,21 @@ def __truediv__(self, name: str) -> "URL":
             return NotImplemented
         return self._make_child((str(name),))
 
+    def __sub__(self, other: object) -> "URL":
```
webknjaz, line 478: Could you add a PEP 257-compliant docstring to this method?

yarl/_path.py
```diff
@@ -39,3 +40,24 @@ def normalize_path(path: str) -> str:
 
     segments = path.split("/")
     return prefix + "/".join(normalize_path_segments(segments))
+
+
+def relative_path(path: str, start: str) -> str:
+    """A wrapper over os.path.relpath()"""
+
+    if not path:
```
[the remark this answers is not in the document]
oleksbabieiev: Unfortunately, this will not work with `os.path.relpath()`. Anyway, I've added a check for absolute and relative paths, so I don't think there will be any more problems with this 🤷‍♂️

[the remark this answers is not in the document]
bdraco: You'll need to use https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.relative_to instead

Here are the equivalent functions
https://docs.python.org/3/library/pathlib.html#corresponding-tools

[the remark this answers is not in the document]
oleksbabieiev: This sounds like a good idea, but `os.path.relpath()` and `PurePath.relative_to()` behave differently. For example, `PurePath("/path/to").relative_to("/path/from")` raises a `ValueError`: '/path/to' is not in the subpath of '/path/from'.
And `os.path.relpath("/path/to", "/path/from")` will just return `../to`.
Would you still like to replace `os.path.relpath()` with `PurePath.relative_to()`?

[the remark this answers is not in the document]
bdraco: It looks like `relpath` has a lot of potential calls into [`abspath`](https://github.com/python/cpython/blob/44f841f01af0fb038e142a07f15eda1ecdd5b08a/Lib/posixpath.py#L376) which will call `getcwd`

I think you could build something based on [`relpath`](https://github.com/python/cpython/blob/44f841f01af0fb038e142a07f15eda1ecdd5b08a/Lib/posixpath.py#L497) which would guarantee no I/O happens.

yarl/_path.py
webknjaz: @oleksbabieiev pro tip: it's usually recommended to accept code suggestions via GH UI, since that makes sure there's no accidental partial copy-paste and preserves the credit.

yarl/_path.py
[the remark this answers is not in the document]
oleksbabieiev: Sorry, I was a bit tired, so I didn't notice at first that `PurePath.relative_to()` takes a `walk_up` argument. Now everything works as it should, so I can completely get rid of `os.path` in my code! 🎉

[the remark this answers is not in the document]
oleksbabieiev: I've got a new problem. The `walk_up` parameter was added only in Python 3.12 🤦‍♂️. I will try to do something with it...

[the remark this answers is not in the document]
oleksbabieiev: I made it work with Python 3.9 🎉

bdraco:
The windows wheel build is failing the tests

You probably want to use PosixPurePath instead so windows doesn’t try to use windows path separators

yarl/_path.py
```diff
@@ -39,3 +40,32 @@ def normalize_path(path: str) -> str:
...
+            break
+        elif path.name == "..":
+            raise ValueError(f"'..' segment in {str(base_path)!r} cannot be walked")
+    else:
+        raise ValueError(
+            f"{str(target_path)!r} and {str(base_path)!r} have different anchors"
+        )
+    offset = len(path.parts)
+    parts = [".."] * step + list(target_path.parts)[offset:]
+    return str(PurePosixPath(*parts))
```
bdraco, line 71:
```suggestion
    return str(PurePosixPath(*("..",) * step, *target_path.parts[offset:]))
```

I think you can avoid creating all the lists

tests/test_url.py
webknjaz, line 1: It would be good to have a hypothesis-based property test where subtracting URLs and then joining them with a slash gives the same result..

yarl/_url.py
```diff
@@ -476,6 +476,34 @@ def __truediv__(self, name: str) -> "URL":
...
+        if type(other) is not URL:
+            return NotImplemented
+
+        target = self._val
+        base = other._val
+
+        if target.scheme != base.scheme:
+            raise ValueError("Both URLs should have the same scheme")
+        if target.netloc != base.netloc:
+            raise ValueError("Both URLs should have the same netloc")
```
bdraco, line 502: Its probably faster to do a tuple unpack on both then all the property accesses, but thats not important right now. I'll write a benchmark for this after we merge it an optimize it if needed

bdraco (approved):
Thanks @oleksbabieiev

yarl/_path.py
```diff
@@ -39,3 +40,31 @@ def normalize_path(path: str) -> str:
...
+    target = target or "/"
+    base = base or "/"
+
+    target_path = PurePosixPath(target)
+    base_path = PurePosixPath(base)
+
+    if not base.endswith("/"):
+        base_path = base_path.parent
+
+    for step, path in enumerate([base_path] + list(base_path.parents)):
```
webknjaz, line 60: Can we replace the creation of two additional in-memory lists with `itertools.chain`?

docs/api.rst
```diff
@@ -974,6 +974,21 @@ The path is encoded if needed.
...
+.. method:: URL.__sub__(url)
+
+   Returns a new URL with a relative *path* between two other URL objects.
+
+   .. doctest::
+
+      >>> target = URL('http://example.com/path/index.html')
+      >>> base = URL('http://example.com/')
+      >>> target - base
+      URL('path/index.html')
```
bdraco, line 990: Also missing a `   .. versionadded:: 1.17` here

bdraco:
@oleksbabieiev Would you please address the additional comments in a followup PR?

oleksbabieiev:
@bdraco yes, but I won't be able to do it until tomorrow. Sorry for the inconvenience

bdraco:
no worries. thank you!

yarl/_path.py
bdraco: It looks like this can be a simple unpack. Addressed in https://github.com/aio-libs/yarl/pull/1379

bdraco: chain was a little bit better... performance is still quadratic in most cases though :(

https://github.com/aio-libs/yarl/pull/1382

docs/api.rst
```diff
@@ -974,6 +974,21 @@ The path is encoded if needed.
          >>> base.join(URL('//python.org/page.html'))
          URL('http://python.org/page.html')
 
+The subtraction (``-``) operator creates a new URL with
```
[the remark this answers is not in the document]
oleksbabieiev: Yes, I think it will be better that way

bdraco:
I looks like I jumped the gun and merged this too soon. It needs a little more work before its ready. I opened https://github.com/aio-libs/yarl/pull/1391 to revert it and the work I did on top of it to try to get it into a good state. Sorry I didn't realize it sooner.

Issue #190, opened 2018-05-02
`if type(val) is str` is a bit too strict


hynek:
It makes sense to use a subclass of str for example if I want to put a secret into a class but want to censor the repr: https://github.com/hynek/environ_config/blob/69b1c8ab25fde3c20fa21de1a044e30760c86d36/src/environ/secrets.py#L114-L124

This breaks with the latest versions.

asvetlov:
Hackish.
I can do `str(val)` if the type is not strict `str` but derived class.
You should understand that it implies a little performance penalty.

hynek:
Why not `isinstance(val, str)`? That slow?

hynek:
I only need this outside of aiohttp to parse URLs from secrets so performance is not an issue. Would it be perf-neutral for you if you added `or isinstance()`? It’s lazily evaluated so it shouldn’t be a measurable slow down?

asvetlov:
Done in yarl 1.2.3

hynek:
Amazing, thank you!

PR #1342, opened 2024-10-20
Fix blocking I/O to load python code when the network location contains non-ascii characters

bdraco:
`urlsplit` is not async safe since it can import `unicodedata` at runtime which does blocking I/O to load python code from disk.  This PR replaces it with a new version that is functionality compatible but rewritten a bit to have better performance and avoid run time imports

To improve performance https://github.com/aio-libs/yarl/issues/190#issuecomment-385992055 suggested cythonizing it but I think thats a lot more to maintain, and the performance issues can likely be solved in python.  

Long term this also means we don't have to use `SplitResult` internally anymore and can fix issues where `urlsplit` doesn't work as we expected and have been sitting in the issue queue for years in future PRs


bdraco:
This is a performance improvement and a bugfix

PR #1343, opened 2024-10-20
Add additional test coverage for invalid URLs

bdraco:
We are missing tests for
- ipv6 left bracket missing
- ipv6 right bracket missing
- ip future being bracketed
- ipv4 being backeted
- stripping unsafe url chars
- stripping control chars

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -10,7 +10,7 @@ source =
   */Lib/site-packages/yarl
 
 [report]
-fail_under = 98.95
+fail_under = 98.94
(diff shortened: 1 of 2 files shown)


.coveragerc
```diff
@@ -10,7 +10,7 @@ source =
   */Lib/site-packages/yarl
 
 [report]
-fail_under = 98.95
+fail_under = 98.90
```
bdraco, line 13: This will come back up once we can remove the skipif

PR #1345, opened 2024-10-20
Break out functions that do not need decorated methods

bdraco:
`classmethod`s and `staticmethod`s are slower than normal methods, and these methods never need anything on the URL objects like `_normalize_path_segments`


bdraco:
I'm not sure the static method is actually slower

If not might be nicer to make them static instead as it's less change

bdraco:
For this case, this change makes the most sense since we have some of them calling each other

bdraco:
staticmethod is a tiny bit slower

PR #1349, opened 2024-10-21
Ensure `SplitResult` is only every constructed once to build new URL objects

bdraco:


bdraco:
Still faster for most cases since its in the LRU

PR #1351, opened 2024-10-21
Extract `_make_netloc` from URL object

bdraco:
The `@classmethod` here is a bit slower, and it was only a `@classmethod` to access `self._QUOTER` which is a constant anyways

### DIFF
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -396,7 +396,7 @@ def _make_netloc(
         user = _QUOTER_NO_REQUOTE(user)
     return f"{user}@{ret}" if user else ret
 
-  
+
 def _query_var(v: QueryVariable) -> str:
     cls = type(v)
     if cls is int:  # Fast path for non-subclassed int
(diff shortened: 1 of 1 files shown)


yarl/_url.py
```diff
@@ -72,6 +72,7 @@
     None, str, "Mapping[str, QueryVariable]", "Sequence[Tuple[str, QueryVariable]]"
 ]
 _T = TypeVar("_T")
+_QUOTER_NO_REQUOTE = _Quoter(requote=False)
```
bdraco, line 75:
```suggestion
QUOTER = _Quoter(requote=False)
```

yarl/_url.py
```diff
@@ -458,7 +493,7 @@ class URL:
     # absolute-URI  = scheme ":" hier-part [ "?" query ]
     __slots__ = ("_cache", "_val")
 
-    _QUOTER = _Quoter(requote=False)
+    _QUOTER = _QUOTER_NO_REQUOTE
```
bdraco, line 496:
```suggestion
    _QUOTER = QUOTER
```

PR #1356, opened 2024-10-21
Move query string functions into their own module

bdraco:
Move query string functions into their own module.

Doc strings have been cleaned up as well.

This is an internal change only to better organize code.

At some point we may want to add direct coverage for these functions, however they are already well covered by existing tests.


bdraco:
performance change is wrong because GHA flaked out in https://github.com/aio-libs/yarl/actions/runs/11432993286 

Need to run it again

PR #1365, opened 2024-10-21
Migrate to PEP 585 and PEP 604 style annotations

bdraco:
Now that 3.9 is the minimum supported Python we can use `from __future__ import annotations`

I seem to recall there might have been a reason why we don't want to do this though...cc @Dreamsorcerer

### DIFF
--- a/yarl/_parse.py
+++ b/yarl/_parse.py
@@ -1,9 +1,10 @@
 """URL parsing utilities."""
 
+from __future__ import annotations
+
 import re
 import unicodedata
 from functools import lru_cache
-from typing import Union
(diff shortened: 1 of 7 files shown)


Dreamsorcerer:
> I seem to recall there might have been a reason why we don't want to do this though...cc @Dreamsorcerer

Because future annotations just means not evaluating annotations on import, not that it should use newer types. If someone evaluates these annotations at runtime, it will error exactly the same way as when not using the future import. e.g. If using get_type_hints() etc.

bdraco:
Thanks. We can wait on this one.

PR #1369, opened 2024-10-21
Move `split_url` cache around encoding the URL

bdraco:
Now that we have own url splitting implementation, we can move the cache one level higher so we can cache more of the url encoding.

We can avoid the case where we have to reconstruct the `SplitResult` after the quoter the values since the `SplitResult` is now only generated after all the work is done.

### DIFF
--- a/CHANGES/1369.misc.rst
+++ b/CHANGES/1369.misc.rst
@@ -1,1 +1,1 @@
-Reworked the internal encoding cache to improve performance on cache hit -- by:user:`bdraco`.
+Reworked the internal encoding cache to improve performance on cache hit -- by :user:`bdraco`.
--- a/CHANGES/1369.misc.rst
+++ b/CHANGES/1369.misc.rst
@@ -1,0 +1,1 @@
+Reworked the internal encoding cache to improve performance on cache hit -- by:user:`bdraco`.
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -121,11 +121,8 @@ def rewrite_module(obj: _T) -> _T:
     return obj
 
 
-# encoded is used for the LRU cache key only; Do not remove it
-# as it would mean encoded and non-encoded URLs would share the same cache
-# and result in incorrect encoding for non-encoded URLs
 @lru_cache
-def encode_url(url_str: str, encoded: bool) -> tuple[SplitResult, _InternalURLCache]:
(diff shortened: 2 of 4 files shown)


CHANGES/1369.misc.rst
```diff
@@ -0,0 +1 @@
+Reworked the internal encoding cache to improve performance on cache hit -- by:user:`bdraco`.
```
bdraco, line 1:
```suggestion
Reworked the internal encoding cache to improve performance on cache hit -- by :user:`bdraco`.
```

yarl/_url.py
```diff
@@ -206,91 +267,22 @@ def __new__(
...
-                )
-
-        self = object.__new__(cls)
-        self._val = val
-        self._cache = cache
-        return self
+        url = object.__new__(cls)
+        url._val = split_result
+        url._cache = cache
+        return url
```
bdraco, line 285:
```suggestion
        self = object.__new__(cls)
        self._val = split_result
        self._cache = cache
        return self
```

yarl/_url.py
```diff
@@ -121,6 +121,67 @@ def rewrite_module(obj: _T) -> _T:
     return obj
 
 
+# encoded is used for the LRU cache key only; Do not remove it
+# as it would mean encoded and non-encoded URLs would share the same cache
+# and result in incorrect encoding for non-encoded URLs
+@lru_cache
+def encode_url(url_str: str, encoded: bool) -> tuple[SplitResult, _InternalURLCache]:
```
bdraco, line 128: Maybe we can remove encoded now... need to write tests though

bdraco:
```
========================================================================= 283800 passed, 800 xfailed in 378.65s (0:06:18) =========================================================================
```
 
`pytest --count=200` looks good

yarl/_url.py
```diff
@@ -206,89 +264,20 @@ def __new__(
...
         elif type(val) is cls:
             return val
         elif type(val) is SplitResult:
             if not encoded:
                 raise ValueError("Cannot apply decoding to SplitResult")
+            split_result = val
+            cache = {}
         elif isinstance(val, str):
-            val = split_url(str(val))
+            split_result, cache = encode_url(str(val))
```
bdraco, line 276: I think we also need to handle the case where its already encoded,  it looks like we have no coverage for that

PR #1375, opened 2024-10-22
Add support for host_port_subcomponent

bdraco:
Adds a new property `host_port_subcomponent` with the goal of replacing 

- the implementation in https://github.com/aio-libs/aiohttp/blob/e7d1d6a488d89520f2f897d11004ad168884fe5f/aiohttp/client_reqrep.py#L365
- a few other places that could use it downstream in Home Assistant and other libs


bdraco:
Need to modify aiohttp to use this and run tests against it

bdraco:
all tests pass downstream with this change

```diff
diff --git a/aiohttp/client_reqrep.py b/aiohttp/client_reqrep.py
index 57df4b016..ca40c19e5 100644
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -433,30 +433,10 @@ class ClientRequest:
         self.headers: CIMultiDict[str] = CIMultiDict()
 
         # Build the host header
-        host = self.url.host_subcomponent
-
+        host = self.url.host_port_subcomponent
         # host_subcomponent is None when the URL is a relative URL.
         # but we know we do not have a relative URL here.
         assert host is not None
-
-        if host[-1] == ".":
-            # Remove all trailing dots from the netloc as while
-            # they are valid FQDNs in DNS, TLS validation fails.
-            # See https://github.com/aio-libs/aiohttp/issues/3636.
-            # To avoid string manipulation we only call rstrip if
-            # the last character is a dot.
-            host = host.rstrip(".")
-
-        # If explicit port is not None, it means that the port was
-        # explicitly specified in the URL. In this case we check
-        # if its not the default port for the scheme and add it to
-        # the host header. We check explicit_port first because
-        # yarl caches explicit_port and its likely to already be
-        # in the cache and non-default port URLs are far less common.
-        explicit_port = self.url.explicit_port
-        if explicit_port is not None and not self.url.is_default_port():
-            host = f"{host}:{explicit_port}"
-
         self.headers[hdrs.HOST] = host
 
         if not headers:
```

bdraco:
Turns out to be super nice on performance since it ends up in the cache and subsequent calls to the same url again doesn't have to recalculated it again if it hits the LRU which turns out to be often

PR #1378, opened 2024-10-22
Small performance improvement to URL subtraction

bdraco:
Avoids some `NamedTuple` property accesses by doing a tuple unpack at the start of the function


bdraco:
Doesn't matter so much because all the time is spent in `PurePath`

Screenshot 2024-10-22 at 10 53 13 AM

bdraco:
Still a tiny bit faster and more in-line with how the rest of the functions are written so let do it.

PR #1379, opened 2024-10-22
Small optimizations to calculate_relative_path

bdraco:
- Use slice for single char check
- Avoid extra list creation


bdraco:
Screenshot 2024-10-22 at 11 22 38 AM

Not optimization here is going to make any significant difference because PurePath is so expensive.

bdraco:
Still a nice cleanup but it won't make any performance difference

PR #1381, opened 2024-10-22
Use faster built-in walk_up implementation for calculate_relative_path on Python 3.12+

bdraco:
For Python 3.12+ we can use the built-in walk_up implementation in pathlib.


bdraco:
Its actually slow 😢

PR #1382, opened 2024-10-22
Improve performance of calculate_relative_path

bdraco:
followup to #1340


bdraco:
that didn't help.. because the `__hash__` implementation is really slow

PR #1384, opened 2024-10-23
Small cleanups for the new subtraction operator

### LINKED ISSUE #1377: Small cleanups for the new subtraction operator
Address post merge feedback in https://github.com/aio-libs/yarl/pull/1340

https://github.com/aio-libs/yarl/blob/04d853c1e87df295981b9d2833c09b2b485a1800/yarl/_path.py#L60

oleksbabieiev:
## What do these changes do?

Small cleanups for the new subtraction operator.

N/A

## Related issue number

Resolves #1377 

## Checklist

### DIFF
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -85,25 +85,25 @@ def test_sub(target: str, base: str, expected: str):
 
 
 def test_sub_with_different_schemes():
-    expected_error_msg = "Both URLs should have the same scheme"
+    expected_error_msg = r"^Both URLs should have the same scheme$"
     with pytest.raises(ValueError, match=expected_error_msg):
         URL("http://example.com/") - URL("https://example.com/")
 
 
 def test_sub_with_different_netlocs():
-    expected_error_msg = "Both URLs should have the same netloc"
+    expected_error_msg = r"^Both URLs should have the same netloc$"
     with pytest.raises(ValueError, match=expected_error_msg):
         URL("https://spam.com/") - URL("https://ham.com/")
 
 
 def test_sub_with_different_anchors():
-    expected_error_msg = "'path/to' and '/path' have different anchors"
+    expected_error_msg = r"^'path/to' and '/path' have different anchors$"
     with pytest.raises(ValueError, match=expected_error_msg):
         URL("path/to") - URL("/path/from")
 
 
 def test_sub_with_two_dots_in_base():
-    expected_error_msg = "'..' segment in '/path/..' cannot be walked"
+    expected_error_msg = r"^'..' segment in '/path/..' cannot be walked$"
     with pytest.raises(ValueError, match=expected_error_msg):
         URL("path/to") - URL("/path/../from")
 
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -2,6 +2,8 @@ from enum import Enum
 from urllib.parse import SplitResult, quote, unquote
 
 import pytest
+from hypothesis import example, given
+from hypothesis import strategies as st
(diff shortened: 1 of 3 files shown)


tests/test_url.py
```diff
@@ -85,29 +85,36 @@ def test_sub(target: str, base: str, expected: str):
...
+    expected_error_msg = r"^'..' segment in '/path/..' cannot be walked$"
     with pytest.raises(ValueError, match=expected_error_msg):
         URL("path/to") - URL("/path/../from")
 
 
+def test_sub_combined_with_join():
+    target = URL("http://example.com/path/to")
+    base = URL("http://example.com/")
+    rel = target - base
+    assert target == base.join(rel)
```
webknjaz, line 115:
```suggestion
    assert target == base / rel
```

tests/test_url.py
```diff
@@ -85,29 +85,36 @@ def test_sub(target: str, base: str, expected: str):
...
 
 
 def test_sub_with_two_dots_in_base():
-    expected_error_msg = "'..' segment in '/path/..' cannot be walked"
+    expected_error_msg = r"^'..' segment in '/path/..' cannot be walked$"
     with pytest.raises(ValueError, match=expected_error_msg):
         URL("path/to") - URL("/path/../from")
 
 
+def test_sub_combined_with_join():
```
webknjaz, line 111: Can we make this hypothesis-based?

tests/test_url.py
oleksbabieiev: `URL.join(url)` and `/` are not interchangeable 😅;
`/` is the equivalent of `URL.joinpath(other, encoded=False)`;
So if you want to use `/`, you need to write `base / rel.path`.

oleksbabieiev: But even `base / rel.path` is not the same as `base.join(rel)`

oleksbabieiev:
I understood why some tests fail. I will try to do something about it.

bdraco:
```
FAILED tests/test_url.py::test_sub_combined_with_join - hypothesis.errors.FailedHealthCheck: Data generation is extremely slow: Only produced 9 valid examples in 1.05 seconds (0 invalid ones and 7 exceeded maximum size). Try decreasing size of the data you're generating (with e.g. max_size or max_leaves parameters).
```
It looks like its failing on performance.  The performance is nearly quadratic unfortunately so any long paths are going to be very slow.

bdraco:
I tried to solve the performance issues in https://github.com/aio-libs/yarl/pull/1388, however in the process I became aware that there are other considerations that are missing.

Its clear this isn't ready to go yet and I merged #1340 too soon (apologies to everyone involved) so I've opened up https://github.com/aio-libs/yarl/pull/1391 to revert it so we can start over with a new approach.

I'm not sure if #1388 is the way to go or not but if it is, its probably a good start.

tests/test_url.py
webknjaz: Ah, I see.. This was my attempt to suggest a proper round-trip data conversion so it could be made into a property-based test.

webknjaz:
Side note: it would probably be useful to have own hypothesis strategy under `yarl.testing` that could be utilized both within yarl's own test suite and any downstream/end-user projects. That strategy would need to produce various URL edge cases.
This is not in the scope of this PR, however, I just wanted to record the idea.

oleksbabieiev:
Well, I think it would make sense if I create a new pull request to rework this feature @bdraco

oleksbabieiev:
#1392

PR #1386, opened 2024-10-24
Make URL subtraction a bit lazier to improve performance

bdraco:


yarl/_path.py
```diff
@@ -58,8 +59,13 @@ def calculate_relative_path(target: str, base: str) -> str:
...
         base_path = base_path.parent
 
+    target_path_parents: Union[list[PurePosixPath], None] = None
     for step, path in enumerate(chain((base_path,), base_path.parents)):
-        if path == target_path or path in target_path.parents:
+        if path == target_path:
+            break
+        if target_path_parents is None:
+            target_path_parents = list(target_path.parents)
+        if path in target_path_parents:
```
bdraco, line 68: It might be slower because list() enumates them all

bdraco:
This seems like it’s getting to complex.  PurePath is very slow and the performance is nearly quadratic.  Maybe we need to replace it

bdraco:
Reimplementing parents is turning out to be even more code.  Maybe this is the way to go

PR #1388, opened 2024-10-25
Reduce time complexity of URL subtraction

bdraco:
Time complexity was roughly `O(base_path_segments * target_path_segments)`

Time complexity is now roughly `O(base_path_segments + target_path_segments)`

### DIFF
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -93,7 +93,9 @@ def test_str():
     ],
 )
 def test_sub(target: str, base: str, expected: str):
-    assert URL(target) - URL(base) == URL(expected)
+    expected_result = URL(expected)
+    result = URL(target) - URL(base)
+    assert result == expected_result
 
 
 def test_sub_with_different_schemes():
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -64,7 +64,7 @@ def test_str():
 @pytest.mark.parametrize(
     ("target", "base", "expected"),
     [
-        ("http://example.com/path/to", "http://example.com/", "to"),
+        ("http://example.com/path/to", "http://example.com/", "path/to"),
         ("http://example.com/path/to", "http://example.com/spam", "path/to"),
         ("http://example.com/this/is/a/test", "http://example.com/this/", "is/a/test"),
         (
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -93,9 +93,7 @@ def test_str():
     ],
 )
 def test_sub(target: str, base: str, expected: str):
-    expected_result = URL(expected)
-    result = URL(target) - URL(base)
-    assert result == expected_result
+    assert URL(target) - URL(base) == URL(expected)
 
 
 def test_sub_with_different_schemes():
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -142,9 +142,9 @@ def test_sub_with_different_anchors():
 
 
 def test_sub_with_two_dots_in_base():
-    expected_error_msg = "'..' segment in '/path/..' cannot be walked"
+    expected_error_msg = "'..' segment in 'path/..' cannot be walked"
     with pytest.raises(ValueError, match=expected_error_msg):
-        URL("path/to") - URL("/path/../from")
+        URL("path/to") - URL("path/../from")
 
 
 def test_repr():
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -66,6 +66,18 @@ def test_str():
     [
         ("http://example.com/path/to", "http://example.com/", "path/to"),
         ("http://example.com/path/to", "http://example.com/spam", "path/to"),
+        (
+            "http://example.com/////path/////to",
+            "http://example.com/////spam",
+            "path/to",
+        ),
+        ("http://example.com/this/is/a/test", "http://example.com/this/", "is/a/test"),
+        (
+            "http://example.com/this/./is/a/test",
+            "http://example.com/this/",
+            "is/a/test",
+        ),
+        ("http://example.com/this/is/../a//test", "http://example.com/this/", "a/test"),
         ("http://example.com/path/to", "http://example.com/spam/", "../path/to"),
         ("http://example.com/path", "http://example.com/path/to/", ".."),
         ("http://example.com/", "http://example.com/", "."),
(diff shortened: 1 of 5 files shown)


yarl/_path.py
```diff
@@ -43,6 +44,50 @@ def normalize_path(path: str) -> str:
...
 
 
+class SimplePath:
+    __slots__ = ("_tail", "_root", "_trailer", "normalized")
+
+    def __init__(self, path: str, strip_root: bool = False) -> None:
+        """Initialize a SimplePath object."""
+        self._tail = [x for x in path.split("/") if x and x != "."]
+
+        if strip_root:
```
bdraco, line 54: Strip tail

yarl/_path.py
```diff
@@ -43,6 +44,50 @@ def normalize_path(path: str) -> str:
     return prefix + "/".join(normalize_path_segments(segments))
 
 
+class SimplePath:
```
bdraco, line 47: Maybe URLPath

yarl/_path.py
```diff
@@ -43,6 +44,50 @@ def normalize_path(path: str) -> str:
...
+    def __init__(self, path: str, strip_root: bool = False) -> None:
+        """Initialize a SimplePath object."""
+        self._tail = [x for x in path.split("/") if x and x != "."]
+
+        if strip_root:
+            if path[-1] != "/" and len(self._tail) > 0:
+                self._tail.pop()
+
+        self._root = "/" if path[0] == "/" else ""
+        self._trailer = "." if not path else ""
```
bdraco, line 59: Can we get rid of this?

yarl/_path.py
```diff
@@ -52,20 +97,44 @@ def calculate_relative_path(target: str, base: str) -> str:
...
-            f"{str(target_path)!r} and {str(base_path)!r} have different anchors"
+        msg = (
+            f"{target_path.normalized!r} and {base_path.normalized!r} "
+            "have different anchors"
         )
-    offset = len(path.parts)
+        raise ValueError(msg)
+
+    offset = path.parts_count
     return str(PurePosixPath(*("..",) * step, *target_path.parts[offset:]))
```
bdraco, line 140: Still have one to get rid of

tests/test_url.py
```diff
@@ -93,7 +93,9 @@ def test_str():
     ],
 )
 def test_sub(target: str, base: str, expected: str):
-    assert URL(target) - URL(base) == URL(expected)
+    expected_result = URL(expected)
+    result = URL(target) - URL(base)
+    assert result == expected_result
```
bdraco, line 98:
```suggestion
    assert URL(target) - URL(base) == URL(expected)
```

commonism:
Hi,

seen this and the current implementation does not work with empty path segments due to the use of PurePosixPath
```python
>>> from pathlib import PurePosixPath
>>> PurePosixPath("////")
PurePosixPath('/')
```

current tests do not cover this. There is tests with empty segments, but they match, therefore undetected.

Please take this into consideration for URLPath.

bdraco:
@commonism Would you please add some xfail tests for this case?

commonism:
c.f. https://github.com/aio-libs/yarl/pull/1389#issuecomment-2437342109

bdraco:
https://github.com/aio-libs/yarl/pull/1390

bdraco:
@commonism Would you please check if it looks right now?

yarl/_path.py
```diff
@@ -43,29 +43,85 @@ def normalize_path(path: str) -> str:
...
-            f"{str(target_path)!r} and {str(base_path)!r} have different anchors"
+            f"{target_path_path!r} and {base_path.path!r} have different anchors"
         )
-    offset = len(path.parts)
-    return str(PurePosixPath(*("..",) * step, *target_path.parts[offset:]))
+
+    for step, base_walk in enumerate(chain((base_path,), base_path.parents())):
+        if base_walk.path == target_path_path:
+            break
+        # If the target_path_parent_strs is already built use the quick path
```
bdraco, line 104:
```suggestion
        # If the target_path_parts is already built we can use a fast path
```

tests/test_url.py
```diff
@@ -110,7 +113,7 @@ def test_sub(target: str, base: str, expected: str):
         (
             "http://example.com////path/////to",
             "http://example.com/////spam",
-            "..//path/////to",
+            "../path/////to",
```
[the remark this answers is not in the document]
bdraco: Maybe it is right .. I'm not really sure

```
http://example.com////path/////to
http://example.com/////spam
```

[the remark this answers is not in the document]
bdraco: or all the tests are wrong....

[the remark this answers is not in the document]
commonism: I'd expect it to follow along with the first test case:

```python
…
("/path/to", "/spam/", "../path/to"),
("//path/to", "/spam/", "..//path/to"),
(
    "http://example.com//path/to",
    "http://example.com/spam",
    "..//path/to",
),
(
    "http://example.com////path/////to",
    "http://example.com/////spam",
    "..//path/////to",
)
```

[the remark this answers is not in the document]
commonism: The trailing / in "spam/" has to be removed.

```
("/path/to", "/spam", "../path/to"),
("/path/to", "/spam/", "../../path/to"),
```
And … others to consider
```
("/path/to", "//spam", "../../path/to"),
(
    "http://example.com/path/to",
    "http://example.com//spam",
    "../../path/to",
),
(
    "http://example.com/////path/////to",
    "http://example.com////spam",
    "../../path/////to",
)
```

yarl/_path.py
```diff
@@ -43,29 +43,78 @@ def normalize_path(path: str) -> str:
...
+
+    def __init__(self, path: str, strip_tail: bool = False) -> None:
+        """Initialize a URLPath object."""
+        had_trailing_slash = path[-1] == "/"
+        # Strip trailing slash
+        if path and had_trailing_slash:
+            path = path[:-1]
+        if "." in path:
+            # Strip '.' segments
+            parts = [x for x in path.split("/") if x != "."]
```
commonism, line 59: stripping "." segments is wrong for "/./" which is equal to "//"

commonism: Maybe right here … but I remember there was a special case on this.

bdraco:
I'm think I'm throwing in the towl on this one and going for a revert.  We need to start over

bdraco:
@commonism Thanks for catching the issues before we shipped this.

I opened https://github.com/aio-libs/yarl/pull/1391 to revert so we can start over with a new approach.

PR #1389, opened 2024-10-25
Add more URL subtraction coverage

bdraco:
https://github.com/aio-libs/yarl/pull/1388 will remove `PurePath` and replace it with an internal helper, however there are still some cases that are not correct that will need to be addressed.  To ensure future refactoring does not cause regressions, add more coverage.

### DIFF
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -80,6 +80,7 @@ def test_str():
         ("http://example.com/this/is/../a//test", "http://example.com/this/", "a/test"),
         ("http://example.com/path/to", "http://example.com/spam/", "../path/to"),
         ("http://example.com/path", "http://example.com/path/to/", ".."),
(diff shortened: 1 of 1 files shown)


commonism:
```python
diff --git a/tests/test_url.py b/tests/test_url.py
index e2bcac7..7b86b55 100644
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -69,7 +69,12 @@ def test_str():
         (
             "http://example.com/////path/////to",
             "http://example.com/////spam",
-            "path/to",
+            "path/////to",
+        ),
+        (
+                "http://example.com////path/////to",
+                "http://example.com/////spam",
+                "..//path/////to",
         ),
         ("http://example.com/this/is/a/test", "http://example.com/this/", "is/a/test"),
         (
```

PR #1392, opened 2024-10-25
Rework of the URL subtraction feature

### LINKED ISSUE #1183: Add Support for Relative Path Calculation

oleksbabieiev:
## What do these changes do?

Rework of the feature added in #1340 and removed in #1391 

Being able to calculate the relative path between two URLs using the subtraction syntax:

```py
from yarl import URL

target = URL("http://example.com/path/index.html")
base = URL("http://example.com/path/")

rel = target - base

print(rel)  # output: "index.html"
```

## Related issue number

Resolves #1183 

## Checklist

## Known issues

- [ ] Time complexity is roughly O(base_path_segments * target_path_segments)
- [ ] Empty segments are not handled correctly #1388 
- [ ] Potentially handling special cases with leading/trailing slashes #1388

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -997,6 +997,20 @@ The path is encoded if needed.
          >>> base.join(URL('//python.org/page.html'))
          URL('http://python.org/page.html')
 
+.. method:: URL.__sub__(url)
+
+   Return a new URL with a relative *path* between two other URL objects.
+   *scheme*, *user*, *password*, *host*, *port*, *query* and *fragment* are removed.
+
+   .. doctest::
+
+      >>> target = URL('http://example.com/path/index.html')
+      >>> base = URL('http://example.com/')
+      >>> target - base
+      URL('path/index.html')
+
+   .. versionadded:: 1.17
+
 Human readable representation
 -----------------------------
 
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -61,6 +61,89 @@ def test_str():
     assert str(url) == "http://example.com:8888/path/to?a=1&b=2"
 
 
+@pytest.mark.parametrize(
+    ("target", "base", "expected"),
+    [
+        ("http://example.com/path/to", "http://example.com/", "path/to"),
+        ("http://example.com/path/to", "http://example.com/spam", "path/to"),
+        ("http://example.com/this/is/a/test", "http://example.com/this/", "is/a/test"),
+        (
+            "http://example.com/this/./is/a/test",
+            "http://example.com/this/",
+            "is/a/test",
+        ),
+        ("http://example.com/this/is/../a//test", "http://example.com/this/", "a/test"),
+        ("http://example.com/path/to", "http://example.com/spam/", "../path/to"),
+        ("http://example.com/path", "http://example.com/path/to/", ".."),
+        ("http://example.com/path", "http://example.com/other/../path/to/", ".."),
+        ("http://example.com/", "http://example.com/", "."),
+        ("http://example.com", "http://example.com", "."),
+        ("http://example.com/", "http://example.com", "."),
+        ("http://example.com", "http://example.com/", "."),
+        ("//example.com", "//example.com", "."),
+        ("/path/to", "/spam/", "../path/to"),
+        ("path/to", "spam/", "../path/to"),
+        ("path/../to", "path/", "../to"),
+        ("path/..", ".", "path/.."),
+        ("path/../replace/me", "path/../replace", "replace/me"),
+        ("path/../replace/me", "path/../replace/", "me"),
+        ("path/to", "spam", "path/to"),
+        ("..", ".", ".."),
+        (".", "..", "."),
+    ],
+)
+def test_sub(target: str, base: str, expected: str):
+    expected_url = URL(expected)
+    result_url = URL(target) - URL(base)
+    assert result_url == expected_url
+
+
+@pytest.mark.xfail(reason="Empty segments are not preserved")
+@pytest.mark.parametrize(
+    ("target", "base", "expected"),
+    [
+        (
+            "http://example.com/////path/////to",
+            "http://example.com/////spam",
+            "path/////to",
+        ),
+        (
+            "http://example.com////path/////to",
+            "http://example.com/////spam",
+            "..//path/////to",
+        ),
+    ],
+)
+def test_sub_empty_segments(target: str, base: str, expected: str):
+    expected_url = URL(expected)
+    result_url = URL(target) - URL(base)
+    assert result_url == expected_url
+
+
+def test_sub_with_different_schemes():
+    expected_error_msg = r"^Both URLs should have the same scheme$"
+    with pytest.raises(ValueError, match=expected_error_msg):
+        URL("http://example.com/") - URL("https://example.com/")
+
+
+def test_sub_with_different_netlocs():
+    expected_error_msg = r"^Both URLs should have the same netloc$"
+    with pytest.raises(ValueError, match=expected_error_msg):
+        URL("https://spam.com/") - URL("https://ham.com/")
+
+
+def test_sub_with_different_anchors():
+    expected_error_msg = r"^'path/to' and '/path' have different anchors$"
+    with pytest.raises(ValueError, match=expected_error_msg):
+        URL("path/to") - URL("/path/from")
+
+
+def test_sub_with_two_dots_in_base():
+    expected_error_msg = r"^'..' segment in '/path/..' cannot be walked$"
+    with pytest.raises(ValueError, match=expected_error_msg):
+        URL("path/to") - URL("/path/../from")
+
+
 def test_repr():
     url = URL("http://example.com")
     assert "URL('http://example.com')" == repr(url)
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -12,7 +12,7 @@ from multidict import MultiDict, MultiDictProxy
 from propcache.api import under_cached_property as cached_property
 
 from ._parse import USES_AUTHORITY, make_netloc, split_netloc, split_url, unsplit_result
-from ._path import normalize_path, normalize_path_segments
+from ._path import calculate_relative_path, normalize_path, normalize_path_segments
 from ._query import (
     Query,
     QueryVariable,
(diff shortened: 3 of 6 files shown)


bdraco:
I'm going to do a 1.17.0 release so we can start preparing aiohttp 3.11.x and I want do to some more downstream cleanups.  We can do a 1.18.0 for this once its ready

yarl/_url.py
```diff
@@ -520,6 +520,32 @@ def __truediv__(self, name: str) -> "URL":
...
+        target_scheme, target_netloc, target_path, _, _ = self._val
+        base_scheme, base_netloc, base_path, _, _ = other._val
+
+        if target_scheme != base_scheme:
+            raise ValueError("Both URLs should have the same scheme")
+        if target_netloc != base_netloc:
+            raise ValueError("Both URLs should have the same netloc")
+
+        path = calculate_relative_path(target_path, base_path)
+        return self._from_tup(("", "", path, "", ""))
```
bdraco, line 547:
```suggestion
        return self._from_parts_("", "", path, "", "")
```

bdraco:
Some conflicts happened. I think I've resolved them correctly

docs/api.rst
```diff
@@ -997,6 +997,20 @@ The path is encoded if needed.
...
+.. method:: URL.__sub__(url)
+
+   Return a new URL with a relative *path* between two other URL objects.
+   *scheme*, *user*, *password*, *host*, *port*, *query* and *fragment* are removed.
+
+   .. doctest::
+
+      >>> target = URL('http://example.com/path/index.html')
+      >>> base = URL('http://example.com/')
+      >>> target - base
```
webknjaz, line 1009: Could you also document what happens when you do `base - target`?

tests/test_url.py
```diff
@@ -61,6 +61,89 @@ def test_str():
...
+        (".", "..", "."),
+    ],
+)
+def test_sub(target: str, base: str, expected: str):
+    expected_url = URL(expected)
+    result_url = URL(target) - URL(base)
+    assert result_url == expected_url
+
+
+@pytest.mark.xfail(reason="Empty segments are not preserved")
```
webknjaz, line 101: What does need to be done to fix this? It'd be useful to include this information into the reason so that it's obvious when it can be removed.

tests/test_url.py
```diff
@@ -61,6 +61,89 @@ def test_str():
...
+        ("http://example.com/", "http://example.com/", "."),
+        ("http://example.com", "http://example.com", "."),
+        ("http://example.com/", "http://example.com", "."),
+        ("http://example.com", "http://example.com/", "."),
+        ("//example.com", "//example.com", "."),
+        ("/path/to", "/spam/", "../path/to"),
+        ("path/to", "spam/", "../path/to"),
+        ("path/../to", "path/", "../to"),
+        ("path/..", ".", "path/.."),
+        ("path/../replace/me", "path/../replace", "replace/me"),
```
commonism, line 88: How is this different to the one below?
("path/../replace/me", "path/../replace/", "me"),

tests/test_url.py
```diff
@@ -61,6 +61,89 @@ def test_str():
...
+        ("http://example.com/path/to", "http://example.com/spam/", "../path/to"),
+        ("http://example.com/path", "http://example.com/path/to/", ".."),
+        ("http://example.com/path", "http://example.com/other/../path/to/", ".."),
+        ("http://example.com/", "http://example.com/", "."),
+        ("http://example.com", "http://example.com", "."),
+        ("http://example.com/", "http://example.com", "."),
+        ("http://example.com", "http://example.com/", "."),
+        ("//example.com", "//example.com", "."),
+        ("/path/to", "/spam/", "../path/to"),
+        ("path/to", "spam/", "../path/to"),
```
commonism, line 85: # no path
("/", "/to", ".."),
("/", "/path/to", "../.."),

# normal 
("/path", "/path/to", ".."),

# trailing / - empy segment at the end
("/path", "/path/", ".."),
("/path", "/path/to/", "../.."),

commonism: c.f. https://github.com/aio-libs/yarl/pull/1388#discussion_r1816549886

asvetlov (changes requested):
While I support the intention I propose more conservative approach.
Let's provide a method, name it, instead of `__sub__` operator overloading.
Add tests, doc, etc.

Publish new `yarl` release, wait for feedback.
Tune and polish.
After some time we can return to adding `__sub__`.

oleksbabieiev:
I sincerely apologize for such a long break. I haven't forgotten about this pull request, I've just been really busy lately. Now I need some time to remember everything and get back to work 🫤

tests/test_url.py
oleksbabieiev: If the last segment of the path does not have a `/` at the end, it is simply ignored when calculating the relative path

oleksbabieiev: In fact, we can say that if the path ends with a `/`, then its last segment is treated as a directory, otherwise - as a file. At least that's how it works for now 🤷‍♂️

commonism: pathlib PosixPath can not be used to relativize URL pathes as URL pathes can have empty segments, and PosixPath strips trailing / anyway.

oleksbabieiev: Yeah, I have already realized that...

oleksbabieiev: I think I've managed to solve it

tests/test_url.py
```diff
@@ -61,6 +61,69 @@ def test_str():
     assert str(url) == "http://example.com:8888/path/to?a=1&b=2"
 
 
+@pytest.mark.parametrize(
+    ("target", "base", "expected"),
+    [
+        ("http://example.com/path/to", "http://example.com/", "path/to"),
+        ("http://example.com/path/to", "http://example.com/spam", "path/to"),
```
commonism, line 68: should be `../path/to` ?

commonism:
I disagree with some of the unit tests, as I expect the returned relative path to add up to the complete path when concatenated as a relative path.

```python
def test_relative_to(target: str, base: str, expected: str):
    expected_url = URL(expected)
    target_url = URL(target)
    base_url = URL(base)
    result_url = target_url.relative_to(base_url)
    assert result_url == expected_url
    combined = base_url / expected
    assert target_url == combined
```

Please include this in the unit tests, it'll help to align those with the idea of a relative path and harden the semantics on this operation wrt. to trailing / in base and result.

tests/test_url.py
webknjaz: Good to know, but this information shouldn't be in a difficult-to-find location. It should be in the reason string.

webknjaz (changes requested):
Please, avoid touching anything in the `packaging/` directory.

oleksbabieiev:
> Please, avoid touching anything in the `packaging/` directory.

Ok, but this was an auto-fix from pre-commit hooks. The bot added it after I had already pushed the changes 🤷‍♂️

webknjaz:
Oh, that's weird. I wonder if the pre-commit.ci cache got invalidated randomly and it pulled in some newer transitive deps...

babieiev:
> ```python
>     combined = base_url / expected
>     assert target_url == combined
> ```
> 
> Please include this in the unit tests, it'll help to align those with the idea of a relative path and harden the semantics on this operation wrt. to trailing / in base and result.

@commonism this condition will only be true if we first normalize the URLs before comparing, so I've modified this code a bit:

```python
combined_url = base_url / expected
combined_path = normalized_path(combined_url.path)
target_path = normalized_path(target_url.path)
assert combined_path == target_path
```

If we don't normalize them, the following can happen:

```python
target_url = URL("/path/to")
base_url = URL("/spam")
relative_url = target_url.relative_to(base_url)
print(relative_url) # URL("../path/to")
combined_url = base_url / relative_url
print(combined_url) # URL("/spam/../path/to") => combined_url != target_url
```

But now everything seems to be working properly.

tests/test_url.py
```diff
@@ -75,41 +77,44 @@ def test_str():
...
-        ("http://example.com/", "http://example.com", "/"),
-        ("http://example.com", "http://example.com/", ""),
-        ("//example.com", "//example.com", ""),
+        ("http://example.com/", "http://example.com/", "."),
+        ("http://example.com", "http://example.com", "."),
+        ("http://example.com/", "http://example.com", "."),
+        ("http://example.com", "http://example.com/", "."),
+        ("//example.com", "//example.com", "."),
         ("/path/to", "/spam/", "../path/to"),
         ("path/to", "spam/", "../path/to"),
```
commonism, line 98:
```
+        (
+                "http://example.com/path/to//",
+                "http://example.com/path/to",
+                ".//",
+        ),
+        (
+                "http://example.com/path/to//",
+                "http://example.com/path/to/",
+                ".//",
+        ),
+

File: tests/test_url.py [hunk 8]
```diff
@@ -75,41 +77,44 @@ def test_str():
...
+        (".", "..", "../."),
     ],
 )
 def test_relative_to(target: str, base: str, expected: str):
     expected_url = URL(expected)
-    result_url = URL(target).relative_to(URL(base))
-    assert result_url == expected_url
+    target_url = URL(target)
+    base_url = URL(base)
+    relative_url = target_url.relative_to(base_url)
```
commonism, line 112: something similar to …
```
assert target_url == (x:=(base_url / expected)), (target_url, x)

Comment (babieiev) [reply tests/test_url.py:122]: Sorry, but what exactly do you want me to change here? If I rewrite everything with the walrus operator, I think it will make the code less readable.

Comment (commonism) [reply tests/test_url.py:122]: It's meant to be a new line, making sure the unit tests input data is consistent
```python
assert target_url == base / expected
```
the walrus is just required for error the message, and can be removed when passing.

commonism:
```python
assert normalize_path(target_url.path) == normalize_path((base_url / expected).path)
```

babieiev: > ```py
> assert normalize_path(target_url.path) == normalize_path((base_url / expected).path)
> ```

I think I already have this in my code 🤔
The only difference is that I added a few variables instead of writing a one-liner:

```py
combined_url = base_url / expected
combined_path = normalize_path(combined_url.path)
target_path = normalize_path(target_url.path)
assert combined_path == target_path
```

babieiev: Anyway, I've already accepted your suggestion

babieiev:
Sorry, I messed up a few things, so I'll open a new pull request 😅

PR #1396, opened 2024-10-30
Complete removal of `SplitResult` as an internal

### LINKED ISSUE #172: Replace _val with separate props for url parts
The target is `URL.build` optimization.
The method is crucial for aiohttp web server, avoiding `SplitResult` construction can make the method faster.

`SplitResult` could be restored in every method that need the object.

N.B. Serialization format should support URL classes generated by older library version or the format should not be changed at all (the later looks easier).

bdraco:
partially implements #172

It was suggested in #172 to only upgrade to `SplitResult` when needed.  This will give us more flexibility to change how the internal structure works in the future and we won't be forever bound to using `SplitResult`.

The only place we can't fully get rid of it is for pickle without a breaking change. #172 suggested leaving it for compat, so thats what I did.

In the future we may want to change the split functions to unpack the values into named variables in future. ie `self._scheme, self._netloc, self._path, self._query, self._fragment = val` or store everything in `self._cache` as the single source of truth; However that was too large of a redesign for this PR since it would also require changing all the tuple unpacking and pre cache in `__new__`.

One consideration about getting rid of the `_val` tuple is that `SplitResult` is exposed as an allowed incoming value in new we would have to have code to convert it. With `_val` being a normal tuple, `SplitResult` can be passed in and its still compatible.

It would be a lot nicer if we stored everything in `self._cache` though since it means the `encode_url` and `pre_encoded_url` calls would return a simple dict that was pre-populated, and it also means `propcache` would have more keys pre-populated.

### DIFF
--- a/tests/test_pickle.py
+++ b/tests/test_pickle.py
@@ -18,15 +18,7 @@ def test_pickle():
 def test_default_style_state():
     u = URL("test")
     hash(u)
-    u.__setstate__(
-        (
-            None,
-            {
-                "_val": ("test", "test", "test", "test", "test"),
-                "_strict": False,
-                "_cache": {"hash": 1},
-            },
-        )
-    )
+    val = ("test", "test", "test", "test", "test")
+    u.__setstate__((None, {"_val": val, "_strict": False, "_cache": {"hash": 1}}))
     assert not u._cache
-    assert u._val == ("test", "test", "test", "test", "test")
+    assert u._val == val
--- a/tests/test_pickle.py
+++ b/tests/test_pickle.py
@@ -18,6 +18,15 @@ def test_pickle():
 def test_default_style_state():
     u = URL("test")
     hash(u)
-    u.__setstate__((None, {"_val": "test", "_strict": False, "_cache": {"hash": 1}}))
+    u.__setstate__(
+        (
+            None,
+            {
+                "_val": ("test", "test", "test", "test", "test"),
+                "_strict": False,
+                "_cache": {"hash": 1},
+            },
+        )
+    )
     assert not u._cache
-    assert u._val == "test"
+    assert u._val == ("test", "test", "test", "test", "test")
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -4,14 +4,25 @@ import warnings
 from collections.abc import Mapping, Sequence
 from functools import _CacheInfo, lru_cache
 from ipaddress import ip_address
-from typing import TYPE_CHECKING, Any, Final, TypedDict, TypeVar, Union, overload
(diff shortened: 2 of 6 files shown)


yarl/_url.py
```diff
@@ -1223,9 +1235,9 @@ def with_fragment(self, fragment: Union[str, None]) -> "URL":
             raise TypeError("Invalid fragment type")
         else:
             raw_fragment = FRAGMENT_QUOTER(fragment)
-        if self._val.fragment == raw_fragment:
+        scheme, netloc, path, query, fragment = self._val
+        if fragment == raw_fragment:
             return self
-        scheme, netloc, path, query, _ = self._val
```
bdraco, line 1228: While its a bit cleaner, this might be slower but the benchmark will tell

yarl/_url.py
```diff
@@ -63,6 +74,7 @@
 
 _T = TypeVar("_T")
 
+
```
bdraco, line 77:
```suggestion
```

yarl/_url.py
bdraco: Never mind it’s faster

tests/test_pickle.py
```diff
@@ -18,6 +18,15 @@ def test_pickle():
 def test_default_style_state():
     u = URL("test")
     hash(u)
-    u.__setstate__((None, {"_val": "test", "_strict": False, "_cache": {"hash": 1}}))
+    u.__setstate__(
+        (
+            None,
+            {
+                "_val": ("test", "test", "test", "test", "test"),
```
bdraco, line 25: Would format nicer if it was a temp var

yarl/_url.py
```diff
@@ -383,26 +390,20 @@ def build(
...
-            SplitResult, (scheme, netloc, path, query_string, fragment)
-        )
+        url._val = (scheme, netloc, path, query_string, fragment)
         url._cache = {}
         return url
 
     @classmethod
-    def _from_tup(cls, val: tuple[str, str, str, str, str]) -> "URL":
+    def _from_tup(cls, val: SplitURL) -> "URL":
         """Create a new URL from a tuple.
```
bdraco, line 400: Stale docstring

bdraco:
For the future we could store everything in `self._cache` as the single source of truth.

That would get rid of the tuple.. but thats a major change, and since SplitResult is
exposed as an allowed incoming value in __new__ we would have to have code to convert it

It would be a lot nicer though since it means the encode_url and unencoded_url calls
would return a simple dict that was pre-populated, and it also means propcache
would have more keys pre-populated

bdraco:
https://github.com/aio-libs/yarl/pull/1397 Aims to fully eliminate Val and only construct it lazily

PR #1397, opened 2024-10-30
Replace `self._val` with separate properties 

### LINKED ISSUE #172: Replace _val with separate props for url parts

bdraco:
`self._val` has been eliminated and replaced with `self._scheme, self._netloc, self._path, self._query, self._fragment`

Second part of https://github.com/aio-libs/yarl/pull/1396

Closes #172


yarl/_url.py
```diff
@@ -80,9 +87,19 @@ class CacheInfo(TypedDict):
 
 
 class _InternalURLCache(TypedDict, total=False):
+
+    # Required, these were the original
+    # values for a SplitResult
+    scheme: str
+    raw_netloc: str
+    _path: str
```
bdraco, line 95: Maybe raw_real_path instead ?

bdraco:
This seems lower and less readable

bdraco:
Going to try one more thing before I throw in the towel on this one

yarl/_url.py
```diff
@@ -432,25 +438,28 @@ def __eq__(self, other: object) -> bool:
...
+            path1,
+            self._query,
+            self._fragment,
+        ) == (
+            other._scheme,
+            other._netloc,
+            path2,
+            other._query,
+            other._fragment,
+        )
```
merwok, line 455: Could it be faster to use chained comparisons here (scheme == scheme and etc == etc) rather than build tuples?

bdraco: Let's try that

bdraco: Faster 👍

bdraco:
still 8% slower making the URL
https://codspeed.io/aio-libs/yarl/branches/elim_val

If we move the URL inside the `pre_encoded_url` and `encode_url`

bdraco:
that works but need to make sure we do not cache on default value since unpickling mutates the object and everything else expected `URL` is immutable.

tests/test_pickle.py
```diff
@@ -16,9 +16,45 @@ def test_pickle():
 
 
 def test_default_style_state():
-    u = URL("test")
-    hash(u)
-    val = ("test", "test", "test", "test", "test")
+    u = object.__new__(URL)
+    val = ("set_state", "set_state", "set_state", "set_state", "set_state")
     u.__setstate__((None, {"_val": val, "_strict": False, "_cache": {"hash": 1}}))
-    assert not u._cache
```
bdraco, line 23: `self._val` is a cached_property so cache will get populated now

yarl/_url.py
```diff
@@ -80,6 +77,15 @@
     Self = Any
 
 
+class UndefinedType(Enum):
+    """Singleton type for use with not set sentinel values."""
+
+    _singleton = 0
+
+
+UNDEFINED = UndefinedType._singleton
```
merwok, line 86: Surprised to see the whole enum machinery here rather than `UNDEFINED = object()` but I suppose the costs are only once at import time

yarl/_url.py
```diff
@@ -1000,19 +1040,20 @@ def with_user(self, user: Union[str, None]) -> "URL":
 
         """
         # N.B. doesn't cleanup query/fragment
```
merwok, line 1042: note still relevant?

yarl/_url.py
bdraco: The typing gets messy otherwise as any object would be allowed

yarl/_url.py
bdraco: Yes

PR #1400, opened 2024-10-30
Release 1.17.1

bdraco:
Screenshot 2024-10-30 at 2 29 49 PM

This one was hard to quantify as almost everything is much faster.


bdraco:
Had a few false starts, but finally happy with the diff between v1.17.0.

bdraco:
Doing one more final retest on production Home Assistant before proceeding to release

PR #1405, opened 2024-10-30
Small cleanups to `encode_url`

bdraco:
- Avoid useless assignments
- Combine nested if statements
- Set `raw_path` in the cache since we already have it along with coverage to make sure its set correctly

### DIFF
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -182,12 +182,13 @@ def encode_url(url_str: str) -> "URL":
 
     if path:
         path = PATH_REQUOTER(path)
-        if netloc:
-            if "." in path:
-                path = normalize_path(path)
+        if netloc and "." in path:
+            path = normalize_path(path)
+    if query:
+        query = QUERY_REQUOTER(query)
+    if fragment:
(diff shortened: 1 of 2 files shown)


bdraco:
1.32s call     tests/test_url_benchmarks.py::test_human_repr

Unrelated but `human_repr` is a bit slow compared to everything else since all the quoting is pure python... not sure how frequently its used in production, but at least on Home Assistant testing its very rare

yarl/_url.py
```diff
@@ -182,15 +182,19 @@ def encode_url(url_str: str) -> "URL":
...
     if path:
         path = PATH_REQUOTER(path)
-        if netloc:
-            if "." in path:
-                path = normalize_path(path)
-
-    query = QUERY_REQUOTER(query) if query else query
-    fragment = FRAGMENT_REQUOTER(fragment) if fragment else fragment
+        if netloc and "." in path:
+            path = normalize_path(path)
```
bdraco, line 186: path can get normalized from `.` to ``

We have no coverage for that

bdraco: added coverage

PR #1407, opened 2024-11-01
Improve performance of the quoters

bdraco:
same change as https://github.com/aio-libs/aiohttp/pull/9603

### DIFF
--- a/yarl/_quoting_c.pyx
+++ b/yarl/_quoting_c.pyx
@@ -102,7 +102,7 @@ cdef inline void _release_writer(Writer* writer):
         PyMem_Free(writer.buf)
 
 
-cdef inline int _write_char(Writer* writer, Py_UCS4 ch, bint changed):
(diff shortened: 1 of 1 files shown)


bdraco:
Looks like its already better optimized here

asvetlov (commented):
Could you please postpone the merge until tomorrow. I'd like to dig in deeper a little

bdraco:
This one didn't show enough performance change to make it worth it.

https://github.com/aio-libs/aiohttp/pull/9603 does though. Already left a note there that it needs a "second set of 👀"

asvetlov:
Well, I've looked on Cython generated code. 
Fast path (when a error is not raised) is equally fast for both versions.
Error path shouldn't be optimized at all, a chance to get out-of-memory error is negligible. It should be processed, sure; but spending time to optimize it doesn't make sense.

bdraco:
I agree, that's why I closed this one

Issue #1411, opened 2024-11-09
Tests failing in a new environment with an exception (`ValueError: chr() arg not in range(0x110000)`)


markgreene74:
### Please confirm the following

- [X] I agree to follow the [aio-libs Code of Conduct][CoC]

[CoC]: ../../.github/blob/master/CODE_OF_CONDUCT.md

- [X] I have checked the [current issues][issue search] for duplicates.

[issue search]: ../search?q=is%3Aissue&type=issues

- [X] I understand this is open source software provided for free and that I might not receive a timely response.
- [X] I am positive I am **NOT** reporting a (potential) security
vulnerability, to the best of my knowledge. *(These must be shared by
submitting [this report form][vulnerability report form] instead, if
any hesitation exists.)*

[vulnerability report form]: ../security/advisories/new

- [X] I am willing to submit a pull request with reporoducers as xfailing test cases or even entire fix. *(Assign this issue to me.)*

### Describe the bug

Working on setting up a dev environment for the `Sprint @ Man Group London Hackathon 2024, Nov 9` with @ajsanchezsanz and trying to run `pytest`, we got the following error:

```
>   return <Py_UCS4>-1
E   ValueError: chr() arg not in range(0x110000)


yarl/_quoting_c.pyx:55: ValueError
```

### To Reproduce

Please note: `pyenv` is used in this example, but the same problem can be observed as well in a brand new GitHub codespace.

- Clone the repo
```
git clone git@github.com:aio-libs/yarl.git --recurse-submodules
```
- Create and activate a new virtual environment, for example using `pyenv`
```
pyenv virtualenv 3.12.4 yarl-env
pyenv activate yarl-env
```
- Install the minimum requirements to run the tests
```
cd yarl
pip install -r requirements/test.txt
pip install -e .
```
- Run the tests
```
pytest ./tests ./yarl -v
```

### Expected behavior

Tests to pass and no exception raised

### Logs/tracebacks

```python-traceback
Excerpt of the traceback:

zsh ❯ pytest ./tests ./yarl -v
=================================== test session starts ===================================
platform darwin -- Python 3.12.4, pytest-8.3.3, pluggy-1.5.0 -- /Users/gc/.pyenv/versions/3.12.4/envs/yarl-env/bin/python
codspeed: 3.0.0 (disabled, mode: walltime, timer_resolution: 41.7ns)
cachedir: .pytest_cache
[...]

PR #1412, opened 2024-11-09
📦 Cap Cython under Python 3.12+ below v3.1

ajsanchezsanz:
## What do these changes do?

Use latest Cython stable version, avoiding using pre-release versions.
Why? There is some issue with Cython 3.1.0a1 that is making the tests to fail.

No

## Related issue number

Fixes https://github.com/aio-libs/yarl/issues/1411

## Checklist

- [ ] Unit tests for the changes exist - N/A
- [ ] Documentation reflects the changes -N/A

### DIFF
--- a/CHANGES/1412.bugfix.rst
+++ b/CHANGES/1412.bugfix.rst
@@ -1,0 +1,3 @@
+Stopped implicitly allowing the use of Cython pre-release versions when
+building the distribution package -- by :user:`ajsanchezsanz` and
+:user:`markgreene74`.
--- a/packaging/pep517_backend/_backend.py
+++ b/packaging/pep517_backend/_backend.py
@@ -11,7 +11,6 @@ from pathlib import Path
 from shutil import copytree
 from sys import implementation as _system_implementation
 from sys import stderr as _standard_error_stream
-from sys import version_info as _python_version_tuple
 from tempfile import TemporaryDirectory
 from warnings import warn as _warn_that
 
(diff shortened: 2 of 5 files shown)


CHANGES/1412.bugfix.rst
webknjaz, line 1: Let's change or symlink this to a `packaging` type.

CHANGES/1412.bugfix.rst
```diff
@@ -0,0 +1 @@
+Removed usage of Cython pre-release versions -- by :user:`ajsanchezsanz`, :user:`markgreene74`.
```
webknjaz, line 1:
```suggestion
Stopped implicitly allowing the use of Cython pre-release versions when
building the distribution package -- by :user:`ajsanchezsanz` and
:user:`markgreene74`.
```

packaging/pep517_backend/_backend.py
```diff
@@ -378,8 +374,7 @@ def get_requires_for_build_wheel(
         )
 
     c_ext_build_deps = [] if is_pure_python_build else [
-        'Cython >= 3.0.0b3' if IS_PY3_12_PLUS  # Only Cython 3+ is compatible
-        else 'Cython',
+        'Cython >= 3.0.0, < 3.1.0',
```
webknjaz, line 377: Let's not put the lower limit for older CPython versions.
```suggestion
        'Cython ~= 3.0.0; python_version >= "3.12"',
        'Cython; python_version < "3.12"',
```

CHANGES/1412.bugfix.rst
asvetlov: Sorry, I don't follow

asvetlov: Ok.
Honestly I don't really care.
Let's rename a file extension without making a symlink

webknjaz:
The bot works correctly. You deleted the file, and so there's no change note in the PR. The proper fix is to actually add one.

asvetlov:
Aah, my oversight. Sorry.

PR #1415, opened 2024-11-14
Add aiohttp tests to CI

### LINKED ISSUE #534: Test aiohttp (and maybe other dependent libraries) before new releases
Many complexities arose from `1.6.0` release (https://github.com/aio-libs/yarl/pull/503 breaking API), specially `aiohttp` breaking in production services, and fixing the `yarl` as `aiohttp` dependency and in production configuration becoming a big challenge.

To prevent something like this happening again, maybe `yarl` should have CI tests, at least as a warning, to ensure `aiohttp` passes tests with new changes, so big issues like that can be caught before a release is made?

If not a CI check, at least some scripts to be run manually.

What do you think?

----

Related convo:
- https://github.com/aio-libs/aiohttp/issues/4987#issuecomment-708104973

Cycloctane:
## What do these changes do?

Test aiohttp with new changes in yarl to prevent unexpected breaking changes.

## Related issue number

fixes #534 

## Checklist

### DIFF
--- a/.github/workflows/aiohttp.yml
+++ b/.github/workflows/aiohttp.yml
@@ -1,0 +1,64 @@
+---
+    
+name: Aiohttp
+
+
+on:
+  merge_group:
+  push:
+    branches:
+    - master
+    tags:
+    - v*
+  pull_request:
+    branches:
+    - master
+
+
+env:
+  COLOR: yes
+
+
+jobs:
+
+  test-aiohttp:
+    name: Aiohttp tests
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout aiohttp
+      uses: actions/checkout@v4
+      with:
+        repository: aio-libs/aiohttp
+        submodules: true
+    - name: Checkout yarl
+      uses: actions/checkout@v4
+      with:
+        path: vendor/yarl
+    - name: Setup Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: 3.x
+        cache: pip
+        cache-dependency-path: requirements/*.txt
+    - name: Update pip, wheel, setuptools, build, twine
+      run: |
+        python -m pip install -U pip wheel setuptools build twine
+    - name: Provision the dev env
+      run: make .develop
+    - name: Cythonize yarl
+      working-directory: vendor/yarl
+      run: make cythonize
+    - name: Install yarl
+      working-directory: vendor/yarl
+      run: >-
+        python -m pip install -e .
+    - name: Ensure version
+      run: python -m pip list
+    - name: Run tests
+      run: python -m pytest
+      shell: bash
+    - name: Run dev_mode tests
+      run: python -X dev -m pytest -m dev_mode
+      shell: bash
+
+...
--- a/.github/workflows/aiohttp.yml
+++ b/.github/workflows/aiohttp.yml
@@ -1,5 +1,5 @@
 ---
-    
+
 name: Aiohttp
 
 
@@ -16,7 +16,7 @@ on:
 
 
 env:
-  COLOR: yes
+  COLOR: "yes"
   FORCE_COLOR: 1
   PY_COLORS: 1
 
@@ -26,6 +26,7 @@ jobs:
   test-aiohttp:
     name: Aiohttp tests
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     steps:
     - name: Checkout aiohttp
       uses: actions/checkout@v4
--- a/.github/workflows/aiohttp.yml
+++ b/.github/workflows/aiohttp.yml
@@ -52,8 +52,6 @@ jobs:
       working-directory: vendor/yarl
       run: >-
         python -m pip install -e .
-    - name: Ensure version
-      run: python -m pip list
     - name: Run tests
       run: python -m pytest
       shell: bash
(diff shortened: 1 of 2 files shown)


asvetlov (approved):
Awesome, thanks!

Would you consider testing against maintained aiohttp releases as well?

I see two options:
1. Install aiohttp from PyPI tarball. It has tests, maybe tarball could be combined with wheels to avoid compilation steps.
2. Install from GitHub maintained branch.

For any approach editing workflow file for updating a list of supported aiohttp versions is totally fine.

I don't know which is better. Maybe fast aiohttp installation time could be a priority.

It could be implemented in the following PR, sure.

What is your opinion?

Cycloctane:
> Awesome, thanks!
> 
> Would you consider testing against maintained aiohttp releases as well?
> 
> I see two options:
> 
>     1. Install aiohttp from PyPI tarball. It has tests, maybe tarball could be combined with wheels to avoid compilation steps.
> 
>     2. Install from GitHub maintained branch.
> 
> 
> For any approach editing workflow file for updating a list of supported aiohttp versions is totally fine.
> 
> I don't know which is better. Maybe fast aiohttp installation time could be a priority.
> 
> It could be implemented in the following PR, sure.
> 
> What is your opinion?

I think both options are acceptable if they achieve the same result. I'd prefer downloading tarball and whl from pypi since it can avoid compiling and building steps.

CHANGES/1415.contrib.rst
```diff
@@ -0,0 +1 @@
+Add aiohttp tests to CI. -- by :user:`Cycloctane`.
```
webknjaz, line 1:
```suggestion
Started running downstream :mod:`aiohttp` tests in CI -- by :user:`Cycloctane`.
```

.github/workflows/aiohttp.yml
```diff
@@ -0,0 +1,67 @@
...
+        path: vendor/yarl
+    - name: Setup Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: 3.x
+        cache: pip
+        cache-dependency-path: requirements/*.txt
+    - name: Update pip, wheel, setuptools, build, twine
+      run: |
+        python -m pip install -U pip wheel setuptools build twine
```
webknjaz, line 48: Nothing short of `build` seems to be necessary here.

.github/workflows/aiohttp.yml
webknjaz, line 1: If we want to make this check required in pull requests, it must be integrated into the main CI/CD workflow.

.github/workflows/aiohttp.yml
```diff
@@ -0,0 +1,67 @@
...
+      run: >-
+        python -m pip install -e .
+    - name: Ensure version
+      run: python -m pip list
+    - name: Run tests
+      run: python -m pytest
+      shell: bash
+    - name: Run dev_mode tests
+      run: python -X dev -m pytest -m dev_mode
+      shell: bash
```
bdraco, line 65: We probably don't need to run the `dev_mode` tests here

.github/workflows/aiohttp.yml
Dreamsorcerer: I don't mind too much, but running a different project's test suite kind of feels to me like it's cleaner to me to be in a separate file. We can still add this as a separate required status in the repo settings.

.github/workflows/aiohttp.yml
Dreamsorcerer: Yeah, there's no harm in running it, but probably not doing anything useful for yarl. I assume that came from the llhttp workflow which definitely does need this to test strict/lax parsing differences.

Dreamsorcerer:
> Downloading tarball and whl from pypi may be better since it can avoid compiling and building steps.

So, one reason you probably want to stick to checking out the aiohttp code, is when yarl decides to break something in aiohttp on-purpose, you won't be able to get the tests passing on a released version of aiohttp. By checking out the code from the repo, we can adjust the tests on aiohttp to allow the new behaviour and then rerun the tests here against the newest commit.

.github/workflows/aiohttp.yml
Cycloctane: > Nothing short of `build` seems to be necessary here.

I just realized that we can remove this entire step, since the neccesary upgrades are already ensured by the makefile command.

bdraco:
/home/docs/checkouts/readthedocs.org/user_builds/yarl/checkouts/1415/docs/changes.rst:19: WARNING: py:mod reference target not found: aiohttp [ref.mod]

Changelog message isn't formatting

bdraco:
> /home/docs/checkouts/readthedocs.org/user_builds/yarl/checkouts/1415/docs/changes.rst:19: WARNING: py:mod reference target not found: aiohttp [ref.mod]
> 
> Changelog message isn't formatting

I'd go with ``aiohttp`` instead since we don't want to add an downstream project to the `intersphinx_mapping` as its unlikely to ever be used.

CHANGES/1415.contrib.rst
```diff
@@ -0,0 +1 @@
+Started running downstream :mod:`aiohttp` tests in CI -- by :user:`Cycloctane`.
```
bdraco, line 1:
```suggestion
Started running downstream ``aiohttp`` tests in CI -- by :user:`Cycloctane`.
```

.github/workflows/aiohttp.yml
bdraco: I'm going to do a 1.18.0 release, and now that this is passing, its super helpful to know that aiohttp isn't going to break from releases .  I think this is good to merge as-is, and we can move it later if we want.

bdraco:
Thanks @Cycloctane

PR #1421, opened 2024-11-18
Allow replacing path without replacing query or fragment

### LINKED ISSUE #111: No way to replace path without replacing query

paul-nameless:
## What do these changes do?

Add keep_query and keep_fragment arguments to with_path function to allow replacing path without  replacing query or fragment

Yes, users can now replace the path in a URL without unintentionally modifying the query string or fragment by using the new arguments. Existing behavior remains unchanged unless the new arguments are explicitly used.

## Related issue number

closes #111 

## Checklist


CHANGES/111.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed an issue where replacing the path would also replace the query and fragment.
```
asvetlov, line 1: Please update the text. Something like "add `keep_query` and `keep_fragment` flags to <enumerate affected methods> for optional keeping query string and fragment in the resulting URL". The exact text is up to you but please keep in mind: it would be nice if a user could get a brief idea of the change without digging into PR content.

paul-nameless: Sure, it makes more sense.
Done

yarl/_url.py
```diff
@@ -1119,7 +1119,14 @@ def with_port(self, port: Union[int, None]) -> "URL":
...
         )
 
-    def with_path(self, path: str, *, encoded: bool = False) -> "URL":
+    def with_path(
+        self,
+        path: str,
+        *,
+        encoded: bool = False,
+        with_query: bool = False,
+        with_fragment: bool = False,
```
bdraco, line 1128:
```suggestion
        keep_query: bool = False,
        keep_fragment: bool = False,
```

yarl/_url.py
```diff
@@ -1128,7 +1135,9 @@ def with_path(self, path: str, *, encoded: bool = False) -> "URL":
                 path = normalize_path(path) if "." in path else path
         if path and path[0] != "/":
             path = f"/{path}"
-        return self._from_parts(self._scheme, netloc, path, "", "")
+        query = self._query if with_query else ""
+        fragment = self._fragment if with_fragment else ""
```
bdraco, line 1139:
```suggestion
        query = self._query if keep_query else ""
        fragment = self._fragment if keep_fragment else ""
```

tests/test_url.py
```diff
@@ -1235,11 +1235,29 @@ def test_with_path_query():
     assert str(url.with_path("/test")) == "http://example.com/test"
 
 
+def test_with_path_keep_query():
+    url = URL("http://example.com?a=b")
+    url2 = url.with_path("/test", with_query=True)
```
bdraco, line 1240:
```suggestion
    url2 = url.with_path("/test", keep_query=True)
```

tests/test_url.py
```diff
@@ -1235,11 +1235,29 @@ def test_with_path_query():
...
 
+def test_with_path_keep_fragment():
+    url = URL("http://example.com#frag")
+    url2 = url.with_path("/test", with_fragment=True)
+    assert str(url2) == "http://example.com/test#frag"
+
+
+def test_with_path_keep_fragment_and_query():
+    url = URL("http://example.com?a=b#frag")
+    url2 = url.with_path("/test", with_query=True, with_fragment=True)
```
bdraco, line 1257:
```suggestion
    url2 = url.with_path("/test", keep_query=True, keep_fragment=True)
```

tests/test_url.py
```diff
@@ -1235,11 +1235,29 @@ def test_with_path_query():
...
+    assert str(url2) == "http://example.com/test?a=b#frag"
+
```
bdraco, line 1259: Please add tests for explicitly passing `keep_fragment=False` and `keep_query=False`

CHANGES/111.bugfix.rst
```diff
@@ -0,0 +1 @@
+Added ``keep_query`` and ``keep_fragment`` flags in the :py:meth:`yarl.URL.with_path`, allowing users to optionally retain the query string and fragment in the resulting URL when replacing the path.
```
bdraco, line 1:
```suggestion
Added ``keep_query`` and ``keep_fragment`` flags in the :py:meth:`yarl.URL.with_path`, allowing users to optionally retain the query string and fragment in the resulting URL when replacing the path -- by :user:`paul-nameless`.
```

bdraco (approved):

bdraco:
reframed this as a new feature since the current behavior doesn't change and its by design. All the functionality here is new.

bdraco:
Screenshot 2024-11-18 at 12 36 02 PM

bdraco:
One last thing, we also need to update the docs for `with_path` with the new params

https://yarl--1421.org.readthedocs.build/en/1421/api/#yarl.URL.with_path

asvetlov:
We also have `.with_name()` and `.with_suffix()`.
Maybe these methods need proposed skip options as well?

paul-nameless:
Updated the documentation

bdraco (approved):

CHANGES/111.feature.rst
```diff
@@ -0,0 +1 @@
+1421.feature.rst
```
asvetlov, line 1: Sorry, I've missed. How is this file rendered finally? What is desired CHANGELOG record?

docs/api.rst
```diff
@@ -652,10 +652,14 @@ section generates a new :class:`URL` instance.
...
 
-.. method:: URL.with_path(path)
+.. method:: URL.with_path(path, keep_query=False, keep_fragment=False)
 
    Return a new URL with *path* replaced, encode *path* if needed.
 
+   .. versionadded:: 1.18
+
+   If ``keep_query=True`` or ``keep_fragment=True`` it retains the existing query or fragment in the URL.
+
```
asvetlov, line 662:
```suggestion
   If ``keep_query=True`` or ``keep_fragment=True`` it retains the existing query or fragment in the URL.

   .. versionchanged:: 1.18

      Added *keep_query* and *keep_fragment* parameters.
```

CHANGES/111.feature.rst
bdraco: Screenshot 2024-11-18 at 9 28 03 PM

https://yarl--1421.org.readthedocs.build/en/1421/changes/#features

bdraco: Usually I use the PR# for the changelog messages. If the issue has a lot in it, symlink to the PR as well so both show up

docs/api.rst
```diff
@@ -652,10 +652,15 @@ section generates a new :class:`URL` instance.
...
+.. method:: URL.with_path(path, keep_query=False, keep_fragment=False)
 
    Return a new URL with *path* replaced, encode *path* if needed.
 
+   If ``keep_query=True`` or ``keep_fragment=True`` it retains the existing query or fragment in the URL.
+
+   .. versionchanged:: 1.18
+
+      Added *keep_query* and *keep_fragment* parameters.
    .. doctest::
```
bdraco, line 664:
```suggestion

   .. doctest::
```

bdraco (approved):

bdraco:
We will also need to bump the version to 1.18.0.dev0 after merging this PR

CHANGES/111.feature.rst
asvetlov: Got you

asvetlov:
> We will also need to bump the version to 1.18.0.dev0 after merging this PR

Technically, we can do it right now.

tests/test_url.py
```diff
@@ -1235,11 +1235,35 @@ def test_with_path_query():
...
+def test_with_path_keep_fragment_and_query():
+    url = URL("http://example.com?a=b#frag")
+    url2 = url.with_path("/test", keep_query=True, keep_fragment=True)
+    assert str(url2) == "http://example.com/test?a=b#frag"
+
+
+def test_with_path_keep_fragment_and_query_false():
+    url = URL("http://example.com?a=b#frag")
+    url2 = url.with_path("/test", keep_query=False, keep_fragment=False)
+    assert str(url2) == "http://example.com/test"
```
webknjaz, line 1264: Looks like these could be collapsed via `@pytest.mark.parametrize()`. Plus, there's the case of `keep_query=False` with `keep_fragment=False` that seems to be missing, but could be added as another param.

tests/test_url.py
```diff
@@ -1240,6 +1240,23 @@ def test_with_path_fragment():
...
 
 
+@pytest.mark.parametrize(
+    "original_url, keep_query, keep_fragment, expected_url",
+    [
+        ("http://example.com?a=b#frag", True, False, "http://example.com/test?a=b"),
+        ("http://example.com?a=b#frag", False, True, "http://example.com/test#frag"),
+        ("http://example.com?a=b#frag", True, True, "http://example.com/test?a=b#frag"),
+        ("http://example.com?a=b#frag", False, False, "http://example.com/test"),
+    ],
```
webknjaz, line 1250: Let's add context by naming each param:
```suggestion
    ("original_url", "keep_query", "keep_fragment", "expected_url"),
    (
        pytest.param("http://example.com?a=b#frag", True, False, "http://example.com/test?a=b", id="query-only"),
        pytest.param("http://example.com?a=b#frag", False, True, "http://example.com/test#frag", id="fragment-only"),
        pytest.param("http://example.com?a=b#frag", True, True, "http://example.com/test?a=b#frag", id="all"),
        pytest.param("http://example.com?a=b#frag", False, False, "http://example.com/test", id="none"),
    ),
```

tests/test_url.py
```diff
@@ -10,7 +10,7 @@
     "\x11\x12\x13\x14\x15\x16\x17\x18\x19\x1a\x1b\x1c\x1d\x1e\x1f "
 )
 _VERTICAL_COLON = "\ufe13"  # normalizes to ":"
-_FULL_WITH_NUMBER_SIGN = "\uFF03"  # normalizes to "#"
+_FULL_WITH_NUMBER_SIGN = "\uff03"  # normalizes to "#"
 _ACCOUNT_OF = "\u2100"  # normalizes to "a/c"
```
webknjaz, line 16: Seeing that the same params are being used in several tests, let's just save the marker into a variable:
```suggestion

with_query_frag_preservation_matrix = pytest.mark.parametrize(
    ("original_url", "keep_query", "keep_fragment", "expected_url"),
    (
        pytest.param("http://example.com?a=b#frag", True, False, "http://example.com/test?a=b", id="query-only"),
        pytest.param("http://example.com?a=b#frag", False, True, "http://example.com/test#frag", id="fragment-only"),
        pytest.param("http://example.com?a=b#frag", True, True, "http://example.com/test?a=b#frag", id="all"),
        pytest.param("http://example.com?a=b#frag", False, False, "http://example.com/test", id="none"),
    ),
)


```

And then, use it on the tests down below as follows:
```python
@with_query_frag_preservation_matrix
def test_...(
    original_url: str,
    keep_query: bool,
    keep_fragment: bool,
    expected_url: str,
): -> None:
    ...
```

asvetlov: I'm -0 for this proposal.
It reduces tests *redundancy* but on the other hand it reduces the tests *readability* as well; I should look at the variable content to figure out that parameters are used.
As a maintainer, I read the code more often than write it.
I like a moderate portion of redundancy in tests, it simplifies reading and debugging at least in *my personal* workflow. Other people could have an opposite opinion, sure.

tests/test_url.py
asvetlov: Agree, the proposed change makes tests much better readable.

yarl/_url.py
```diff
@@ -1276,7 +1285,12 @@ def with_fragment(self, fragment: Union[str, None]) -> "URL":
             self._scheme, self._netloc, self._path, self._query, raw_fragment
         )
 
-    def with_name(self, name: str) -> "URL":
+    def with_name(
+        self,
+        name: str,
+        keep_query: bool = False,
+        keep_fragment: bool = False,
+    ) -> "URL":
```
asvetlov, line 1293:
```suggestion
    def with_name(
        self,
        name: str,
        *,
        keep_query: bool = False,
        keep_fragment: bool = False,
    ) -> "URL":
```

yarl/_url.py
```diff
@@ -1303,9 +1317,17 @@ def with_name(self, name: str) -> "URL":
...
+        query = self._query if keep_query else ""
+        fragment = self._fragment if keep_fragment else ""
+        return self._from_parts(self._scheme, netloc, "/".join(parts), query, fragment)
+
+    def with_suffix(
+        self,
+        suffix: str,
+        keep_query: bool = False,
+        keep_fragment: bool = False,
+    ) -> "URL":
```
asvetlov, line 1330:
```suggestion
    def with_suffix(
        self,
        suffix: str,
        *,
        keep_query: bool = False,
        keep_fragment: bool = False,
    ) -> "URL":
```

asvetlov (changes requested):
Please update affected method signatures in docs to reflect added parameters.

docs/api.rst
```diff
@@ -652,10 +652,16 @@ section generates a new :class:`URL` instance.
       >>> URL('http://example.com:8888').with_port(None)
       URL('http://example.com')
 
-.. method:: URL.with_path(path)
+.. method:: URL.with_path(path, keep_query=False, keep_fragment=False)
```
asvetlov, line 655:
```suggestion
.. method:: URL.with_path(path, *, keep_query=False, keep_fragment=False)
```

docs/api.rst
```diff
@@ -857,27 +863,39 @@ section generates a new :class:`URL` instance.
       >>> URL('http://example.com/path#frag').with_fragment(None)
       URL('http://example.com/path')
 
-.. method:: URL.with_name(name)
+.. method:: URL.with_name(name, keep_query=False, keep_fragment=False)
```
asvetlov, line 866:
```suggestion
.. method:: URL.with_name(name, *, keep_query=False, keep_fragment=False)
```

docs/api.rst
```diff
@@ -857,27 +863,39 @@ section generates a new :class:`URL` instance.
...
+
    .. doctest::
 
       >>> URL('http://example.com/path/to?arg#frag').with_name('new')
       URL('http://example.com/path/new')
       >>> URL('http://example.com/path/to').with_name("ім'я")
       URL('http://example.com/path/%D1%96%D0%BC%27%D1%8F')
 
-.. method:: URL.with_suffix(suffix)
+.. method:: URL.with_suffix(suffix, keep_query=False, keep_fragment=False)
```
asvetlov, line 886:
```suggestion
.. method:: URL.with_suffix(suffix, *, keep_query=False, keep_fragment=False)
```

asvetlov (approved):
LGTM, thanks!

PR #1423, opened 2024-11-19
Bump to 1.18.0.dev0

asvetlov:


webknjaz:
Why?

bdraco:
Its for https://github.com/aio-libs/yarl/pull/1421

webknjaz:
Ah, makes sense. I was lacking that context. This causes me to think that maybe such bumps would belong in related PRs, though.

PR #1424, opened 2024-11-20
Remove internal `_default_port` property

bdraco:
## What do these changes do?

Now that we have scheme available as `self._scheme` is should be a bit faster to avoid the extra property lookup, and a few less lines of code.

Screenshot 2024-11-19 at 10 01 55 PM
Screenshot 2024-11-19 at 10 02 01 PM

no

### DIFF
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -102,7 +102,6 @@ class _InternalURLCache(TypedDict, total=False):
     absolute: bool
     scheme: str
     raw_authority: str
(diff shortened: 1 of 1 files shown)


bdraco:
*sigh* codecov-action release broke ci
https://github.com/codecov/codecov-action/releases

bdraco:
https://github.com/codecov/codecov-action/issues/1683

PR #1425, opened 2024-11-20
Try downgrading codecov-action to 5.0.3

bdraco:
see https://github.com/codecov/codecov-action/issues/1683


bdraco:
We should remove the code-cov pin once https://github.com/codecov/codecov-action/issues/1683 is solved

PR #1432, opened 2024-11-29
Add a cache for building encoded URLs

bdraco:
Since `aiohttp` `web_request` tends to see the same `URL`s over and over, its advantageous to cache building the `URL` objects. We currently have a cache for parsing URLs, but we did not have one for building URLs.  

We didn't have a cache on `URL.build`  because `URL.build` accepts a dict for the query which is not hashable. This is solved by constructing the query string before the cache is used. Note, that it may be better to avoid caching URLs with a query string in future but it didn't seem make enough difference in testing that it was worth the additional complexity.

Since `aiohttp` rarely `URL.build` with `encoded=False` we do not cache unencoded builds as the hit rate was < 50% in production testing.  The hit rate for https://github.com/aio-libs/aiohttp/blob/1fa237ffc9e7aa70cbabb68ab64d6fe03255cbc0/aiohttp/web_request.py#L429 is nearly perfect

Additionally, and similar to https://github.com/aio-libs/yarl/pull/1434 every time the new `URL` object was created it would have a new `self._cache`, but caching the build, `self._cache` will already be initialized.

### DIFF
--- a/CHANGES/1432.misc.rst
+++ b/CHANGES/1432.misc.rst
@@ -1,0 +1,1 @@
+Improved cache performance when :class:`~yarl.URL` objects are constructed from :py:meth:`~yarl.URL.build` with ``encoded=True`` -- by :user:`bdraco`.
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -222,7 +222,7 @@ def build_pre_encoded_url(
     host: str,
     port: Union[int, None],
     path: str,
-    query_string: str,
+    qs: str,
     fragment: str,
     quote_query: bool,
 ) -> "URL":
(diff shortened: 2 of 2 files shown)


bdraco:
The hit rate for pre-encoded in production is great. Not so much for unencoded ... might not make sense to cache building unencoded

bdraco:
We still end up losing all the cache if they call 

```
    @reify
    def url(self) -> URL:
        """The full URL of the request."""
        # authority is used here because it may include the port number
        # and we want yarl to parse it correctly
        return URL.build(scheme=self.scheme, authority=self.host).join(self._rel_url)
```

Because `.join` isn't cached....

bdraco:
2024-11-29 09:05:09.163 CRITICAL (SyncWorker_9) [homeassistant.components.profiler] Cache stats for lru_cache <function build_pre_encoded_url at 0x7f827e1e1260> at /usr/local/lib/python3.13/site-packages/yarl/_url.py: CacheInfo(hits=187, misses=15, maxsize=128, currsize=15)

PR #1433, opened 2024-11-29
DNM: test hit rate of a join cache

bdraco:
https://github.com/aio-libs/yarl/pull/1432#issuecomment-2508133741

PR #1434, opened 2024-11-29
Make `from_parts` a LRU to increase the chance we can preserve the internal cache

bdraco:
We end up doing the same modification to `URL` objects quite frequently. Because `URL` objects are immutable, when this happens the `self._cache` is lost and has to be reconstructed for the new `URL` object that is produced. Additionally, we had ended up with a few places where the `object.__new__(URL)` was inlined because it was faster. We can move this all behind an `lru_cache` which ensures `self._cache` is preserved if the parts are in the cache which is the common case for most applications. With this change we can centralize most of the `__new__` construction and make the code a bit more DRY.

### DIFF
--- a/CHANGES/1434.misc.rst
+++ b/CHANGES/1434.misc.rst
@@ -1,1 +1,1 @@
-Improved cache performance for operations that modify the :class:`~yarl.URL` object -- by :user:`bdraco`.
+Improved cache performance for operations that produce a new :class:`~yarl.URL` object -- by :user:`bdraco`.
--- a/CHANGES/1434.misc.rst
+++ b/CHANGES/1434.misc.rst
@@ -1,0 +1,1 @@
+Improved cache performance for operations that modify the :class:`~yarl.URL` object -- by :user:`bdraco`.
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -213,6 +213,19 @@ def pre_encoded_url(url_str: str) -> "URL":
     return self
 
 
+@lru_cache
+def from_parts(scheme: str, netloc: str, path: str, query: str, fragment: str) -> "URL":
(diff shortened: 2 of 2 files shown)


bdraco:
Looking at the distribution of calls. It makes sense to close  https://github.com/aio-libs/yarl/pull/1433 and merge a combination of this and https://github.com/aio-libs/yarl/pull/1432

bdraco:
Hit rate looks great

`2024-11-29 11:58:55.494 CRITICAL (SyncWorker_0) [homeassistant.components.profiler] Cache stats for lru_cache <function from_parts at 0x7ff1e12c5300> at /usr/local/lib/python3.13/site-packages/yarl/_url.py: CacheInfo(hits=2404, misses=502, maxsize=128, currsize=128)`

Also we end up with a massive improvement in the `self._cache` hit rate because we don't have to rebuild it every time we modify the url if its already in the cache

CHANGES/1434.misc.rst
```diff
@@ -0,0 +1 @@
+Improved cache performance for operations that modify the :class:`~yarl.URL` object -- by :user:`bdraco`.
```
bdraco, line 1:
```suggestion
Improved cache performance for operations that produce a new :class:`~yarl.URL` object -- by :user:`bdraco`.
```

bdraco:
2024-11-29 19:42:14.766 CRITICAL (SyncWorker_5) [homeassistant.components.profiler] Cache stats for lru_cache <function from_parts_uncached at 0x7f1095e91440> at /usr/local/lib/python3.13/site-packages/yarl/_url.py: CacheInfo(hits=13122, misses=228, maxsize=128, currsize=128)

bdraco:
Thats after startup ~100 minutes

PR #1435, opened 2024-11-29
Release 1.18.1

bdraco:
Screenshot 2024-11-29 at 2 11 34 PM

benchmark change: https://codspeed.io/aio-libs/yarl/runs/compare/673f3e7a6b39e9df22ee032d..674a20e33e94b7d243f7a97a


bdraco:
I took another look at the miss rate on `from_parts` and it looks like its always query string so it probably makes sense not to cache that

bdraco:
Will redo this after https://github.com/aio-libs/yarl/pull/1436

PR #1436, opened 2024-11-29
Avoid caching the constructed URL when altering the query string

bdraco:
When a query string is passed, the hit rate on the cache goes way down. Its best not to cache the query calls because the data is far more likely to change for query values.

This will show a reduction on the benchmark but its no different than it was before #1434


CHANGES/1436.misc.rst
```diff
@@ -0,0 +1 @@
+1432.misc.rst
```
bdraco, line 1:
```suggestion
1434.misc.rst
```

bdraco:
`CacheInfo(hits=771, misses=132, maxsize=128, currsize=128)`


Once its prime the miss rate barely increases now.

PR #1437, opened 2024-11-29
Increment version to 1.18.2.dev0

bdraco:


bdraco:
I'm going to bump to .2 and do another release
since PEP440 recommends doing that
> Note

> Creating post-releases of pre-releases is strongly discouraged, as it makes the version identifier difficult to parse for human readers. In general, it is substantially clearer to simply create a new pre-release by incrementing the numeric component.

PR #1443, opened 2024-12-01
Improve performance of path functions on cache miss

bdraco:
There were a few places we called the quoter when we know its never going to do anything.

On cache miss for `.query_string`, this will help a bit.  ~~However its not possible to benchmark right now because of https://github.com/aio-libs/yarl/pull/1443#issuecomment-2509515742~~

We can avoid some of the property accesses internally as well which should speed these up a bit.  Since the `aiohttp` url dispatcher calls `path_safe` there is a much larger chance we get a cache miss so it makes sense to optimize this one a bit.

### DIFF
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -797,6 +797,7 @@ class URL:
             # To avoid string manipulation we only call rstrip if
             # the last character is a dot.
             raw = raw.rstrip(".")
+        port = self.explicit_port
         if (port := self.explicit_port) is None or port == DEFAULT_PORTS.get(
             self._scheme
         ):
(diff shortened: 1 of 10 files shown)


yarl/_url.py
```diff
@@ -790,15 +790,17 @@ def host_port_subcomponent(self) -> Union[str, None]:
...
             # they are valid FQDNs in DNS, TLS validation fails.
             # See https://github.com/aio-libs/aiohttp/issues/3636.
             # To avoid string manipulation we only call rstrip if
             # the last character is a dot.
             raw = raw.rstrip(".")
-        if port is None or port == DEFAULT_PORTS.get(self._scheme):
+        port = self.explicit_port
+        if (port := self.explicit_port) is None or port == DEFAULT_PORTS.get(
+            self._scheme
+        ):
```
bdraco, line 803:
```suggestion
        if port is None or port == DEFAULT_PORTS.get(self._scheme):
```

bdraco:
Its not going to show any difference because the cached_property is already cached and it only does it once

`type(url).query_string.wrapped(url)` would work if `wrapped` was `public` in `propcache`

bdraco:
Probably should do a new release of `propcache` with `wrapped` marked as `public` but it will likely fail because of https://github.com/pypa/gh-action-pypi-publish/issues/307

bdraco:
Its actually unlikely to make much difference since with the caching improvements (unreleased due to the publish issue) its so much more likely to be cached anyways.

bdraco:
Well I guess it is worth it

PR #1447, opened 2024-12-01
Increment version to 1.18.4.dev0

bdraco:


bdraco:
Perf regression is not real.  Released versions ship without without cython tracing but CI versions use cython tracing shows its showing the difference

PR #1452, opened 2024-12-07
proposal: use meson-python as build backend

trim21:
## What do these changes do?

This is proposal and POC for using meson-python as build system.

Currently we need to maintain a build backend to support our use-case for pure-python/binary module, dynamic cython flags, which is supported by meson.

### uv

uv related change are just for speedup, which can be removed.

### coverage hit line changes

coverage dropped ~200 lines from removed packaging directory.

### benchmark 

benchmark of this PR is misleading becuase it use different compile flags, it should not be compared with master branch.

### Cython upgrade:

cython check `Py_UCS4` value range in trace mode, `<Py_UCS4>-1` now will throw a `ValueError`, so a extra return value is need to indicate the `_restore_ch` return.

The upgrade of cython can be avoided by patch generated c file: https://github.com/aio-libs/yarl/pull/1452/commits/0c4e11cb36fa9675969bb6313d778c65aa521308#diff-feabd8e812a476b3324d4cc9144ef3e2ff439b8afda577322f4ba99f6185d63b

No, only affect developers.

## Checklist


webknjaz (changes requested):
Thanks for the proposal but I'm not seeking to migrate away from setuptools anytime soon. This adds unnecessary maintenance burden on many levels, including replicating the configuration across many similarly packaged project within and outside aio-libs, keeping all of them in sync, learning several new tools and supporting this long-term (I don't suppose you're offering to support packaging of all aio-libs projects for years to come instead of me, are you?)

Other things I see being problematic:
* another interference change for the downstream packagers
* forcing contributors to install even more tooling
* a bunch of formatting changes all over
* some test matrix changes
* requiring cython to build sdists
* blocking the upgrade path to eventually using setuptools-scm
* losing the reproducibility infrastructure
* unrelated hacks making their way into a packaging PR
* the patch being big to the point of discouraging the reviewers from scheduling looking into it
* no prior discussion / agreement of the need or any acceptance really

So I'm strongly 👎 on this.

trim21:
Thanks for your response.

> I don't suppose you're offering to support packaging of all aio-libs projects for years to come instead of me, are you?

You have convinced me here 😆

I wouldn't call it mergeable eigher, it present as a proposal and POC for another option without  maintaining a build backend for further discussion.

It's indeed a dead end that does not lead to `setuptools-scm` and add maintenance complexity.

I'll close it.

Some misunderstanding:

It doesn't not require cpython to build sdist (ninja extra is optional and can be removed), it require cpython only when you compile binary wheel from sdist. on pypy it's will only build a pure python wheel.

ths hack on yarl/_quoting_c.pyx is caused by upgrading the cython to 3.1.0a1 to support pyx coverage, and will be fixed in 3.1, which is also not necessary. the formatter change in yarl/_quoting_c.pyi is also not expected not generated by ruff-format, both I'd very much to avoid but necessary to make ci pass at this time.

webknjaz:
> It doesn't not require cpython to build sdist (ninja extra is optional and can be removed), it require cpython only when you compile binary wheel from sdist.

I made a typo and meant Cython. What I meant is that the deps listed in `[build-system].requires` are unconditional and Cython would be unnecessarily pulled in for building the sdist (even if unused).

Among other things that are lost is the reference conversion in readme+changelog.

By the way, there's no Ruff in this project, so those changes are coming from your editor being misconfigured.

trim21:
OH, sorry. It's not ruff-format but black. It will infer minial target python version from pyproject and cause this change.

PR #1456, opened 2024-12-20
Support the free-threaded build of CPython 3.13

### LINKED ISSUE #1455: Support the free-threaded build of CPython 3.13

_No response_

Hey all! 👋

I took the liberty of looking into adding support for the 3.13 free-threaded build of CPython after `multidict` (I also plan to look at other libraries under the `aio-libs` organization). This aims to be a tracking issue for all work necessary to support it, for which I'm going to open a PR shortly.

- [ ] Set up CI for free-threading
- [ ] Audit the Cython extension module for thread-safety issues
  - [ ] The only issues seems to be the usage of a statically allocated buffer for the `Writer` struct. That's fairly straightforward to solve.
- [ ] Audit the Python code for thread-safety issues
  - [ ] One fairly complicated thing to figure out was that there's some thread safety issues in the Python code as well. This stems from havily using non-thread-local caches when creating `URL` instances. Instantiating a `URL` with a string url for example calls `encode_url` which uses an `lru_cache`. This means that multiple threads might end up sharing the same object, even if the `URL` object appears to have been created locally. There's a couple possible solution to this:
    - Disable `lru_cache` under the free-threaded build so that object sharing between threads is more straightfoward.
    - Instead of using `lru_cache` to cache the `URL` object itself, it could instead cache the immutable parts that make up a `URL`. 
- [ ] Mark the Cython extension module as thread-safe with setting `freethreading_compatible` in the cython directives (can be easily done through `pyproject.toml`).
- [ ] Build wheels for the free-threaded build.

Hope this all sounds good! Let me know in case there's something that doesn't seem right.

- We could say that people testing on the free-threaded build should only install the pure-Python wheels, but that doesn't seem like the best say forward, since the required changes are not that intrusive.

### Additional context

_No response_

### Code of Conduct

lysnikolaou:
## What do these changes do?

- Add a matrix config for free-threaded Python to run tests with it.
- Change configuration to correctly build wheels under the free-threaded build.
- Fix thread-safety issues in the Cython extension module.

No.

## Related issue number

Resolves #1455 

## Checklist


tests/test_quoting_benchmarks.py
webknjaz, line 1: A more elegant solution for applying the same marker to all the tests would be setting the module-level value:
```python
pytestmark = pytest.mark.skipif(CODSPEED_MISSING, reason="pytest-codspeed needs to be installed")
```

Also, `pytest` has an [`importorskip()`](https://docs.pytest.org/en/stable/reference/reference.html#pytest-importorskip) helper for this use-case which might be even better fit.

docs/spelling_wordlist.txt
```diff
@@ -1,5 +1,6 @@
 Bluesky
 Bugfixes
+CPython
 Changelog
 Codecov
```
webknjaz, line 5:
```suggestion
Changelog
Codecov
CPython
```

.github/workflows/reusable-linters.yml
```diff
@@ -21,7 +21,7 @@ env:
   PY_COLORS: 1  # Recognized by the `py` package, dependency of `pytest`
   PYTHONIOENCODING: utf-8
   PYTHONUTF8: 1
-  PYTHON_LATEST: 3.12
+  PYTHON_LATEST: 3.13
```
webknjaz, line 24: Could you move changes that aren't directly related into separate PRs? This would make the diff smaller and easier to follow in general.

packaging/pep517_backend/_backend.py
```diff
@@ -371,10 +372,12 @@ def get_requires_for_build_wheel(
...
-    c_ext_build_deps = [] if is_pure_python_build else [
-        'Cython ~= 3.0.0; python_version >= "3.12"',
-        'Cython; python_version < "3.12"',
-    ]
+    if is_pure_python_build:
+        c_ext_build_deps = []
+    elif sysconfig.get_config_var('Py_GIL_DISABLED'):
+        c_ext_build_deps = ['Cython ~= 3.1.0a1']
+    else:
+        c_ext_build_deps = ['Cython ~= 3.0.11']
```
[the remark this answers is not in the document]
lysnikolaou: I don't have a strong opinion on this. You know best what the availability of the team will be in case something breaks because of a Cython release. Maybe loosening the constraint but have a more strictly pinned build requirement might be a better idea?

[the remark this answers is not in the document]
webknjaz: Yeah, I usually try pinning things. But last time I worked towards that goal, cibuildwheel had a few limitations in this area. They should be fixed by now so I should look into it again…

requirements/cython.txt
```diff
@@ -1 +1 @@
-cython==3.0.11
+cython~=3.0.11
```
webknjaz, line 1: Because this is a constraint file integrated in `cibuildwheel` via `pyproject.toml`, I'd expect it to be pinned exactly.
```suggestion
cython==3.0.11
```

Thought I think I'd just go ahead and merge #1469 in to bump it already and this file wouldn't be updated in the PR.

webknjaz:
>     * Use Cython 3.0.x for all version, except for free-threaded where we're using 3.1.0a1

It's probably okay for now. Although, see my comment about `cython.txt` being a constraint file.

>     * Split dependencies so that `codspeed` is not a required test dependency and can be omitted in `cibuildwheel` (that's because `codspeed` depends on `cffi`, which does not have free-threaded wheels available yet - though there's ongoing work on that).

Sounds reasonable.

> If you this is all okay, I'd be happy to change the rest of my PRs (`frozenlist`, `propcache`) to follow this exact approach too.

Yep, with the constraint file quirk.

------

Out of curiosity, have you checked the [`importorskip()`](https://docs.pytest.org/en/stable/reference/reference.html#pytest-importorskip) idea? Does it not work?

webknjaz:
FTR the CI is failing due to https://github.com/actions/setup-python/issues/1034 and the following fix will need to be implemented here: https://github.com/aio-libs/frozenlist/pull/622 + https://github.com/aio-libs/frozenlist/pull/633.

lysnikolaou:
> Out of curiosity, have you checked the [`importorskip()`](https://docs.pytest.org/en/stable/reference/reference.html#pytest-importorskip) idea? Does it not work?

Yes. `mypy` fails if we use `importorskip`, because it cannot find the `BenchmarkFixture` attribute. Not sure why that is.

requirements/cython.txt
lysnikolaou: Done! I pinned it to 3.0.11. Once #1469 is in, I can rebase so that we get 3.0.12.

Dreamsorcerer:
> > Out of curiosity, have you checked the [`importorskip()`](https://docs.pytest.org/en/stable/reference/reference.html#pytest-importorskip) idea? Does it not work?
> 
> Yes. `mypy` fails if we use `importorskip`, because it cannot find the `BenchmarkFixture` attribute. Not sure why that is.

The old code referenced the result of importorskip() instead of importing the module, no type checker is going to know what that is. If you wanted to use both, you could use importorskip for the runtime behaviour, then use `if TYPE_CHECKING` to import the module for type checkers.

Dreamsorcerer:
i.e.
```
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from pytest_codspeed import BenchmarkFixture

def test...(benchmark: "BenchmarkFixture"):
```

Dreamsorcerer:
> > If you wanted to use both, you could use importorskip for the runtime behaviour, then use `if TYPE_CHECKING` to import the module for type checkers.
> 
> Thanks for the pointer. Do you feel that this is a better approach than the one currently implemented? The current one does both in a single try/except, so it feels a bit clener to me.

I don't mind either way. @webknjaz?

lysnikolaou:
> FTR the CI is failing due to [actions/setup-python#1034](https://github.com/actions/setup-python/issues/1034) and the following fix will need to be implemented here: [aio-libs/frozenlist#622](https://github.com/aio-libs/frozenlist/pull/622) + [aio-libs/frozenlist#633](https://github.com/aio-libs/frozenlist/pull/633).

One alternative to this would be to use `setup-uv` instead of `setup-python`, which could also speed up dependency installation etc. If you feel okay about this, I can try it out in a separate PR. Otherwise, I can also move the custom actions to this repo as well.

webknjaz:
> > FTR the CI is failing due to [actions/setup-python#1034](https://github.com/actions/setup-python/issues/1034) and the following fix will need to be implemented here: [aio-libs/frozenlist#622](https://github.com/aio-libs/frozenlist/pull/622) + [aio-libs/frozenlist#633](https://github.com/aio-libs/frozenlist/pull/633).
> 
> One alternative to this would be to use `setup-uv` instead of `setup-python`, which could also speed up dependency installation etc. If you feel okay about this, I can try it out in a separate PR. Otherwise, I can also move the custom actions to this repo as well.

I'm still looking into the uv integration, I'm not sure it's ready for prime time just yet. I'll be looking into that separately. But feel free to send a PR with the caching fix separately.

bdraco:
The performance regression in the writer looks real.  I can't come up with a better solution though that is thread-safe.

packaging/pep517_backend/_backend.py
bdraco, line 380:
```suggestion
        c_ext_build_deps = ['Cython ~= 3.0.12']
```

requirements/codspeed.txt
```diff
@@ -0,0 +1,2 @@
+-r test.txt
+pytest-codspeed==3.1.2
```
bdraco, line 2:
```suggestion
pytest-codspeed==3.2.0
```

lysnikolaou:
> The performance regression in the writer looks real. I can't come up with a better solution though that is thread-safe.

We could try to make it so that the global buffer is used in the non-free-threaded build, but not in the free-threaded one. Cython, however, makes that particularly hard. There's two possible solutions:

- Rewrite the buffer handling in C, where we can use `#ifdef Py_GIL_DISABLED`.
- Use the Cython compile-time `DEF` and `IF` features, that have been deprecated and have no good alternatives.

Both of the solutions are sub-optimal. Thoughts?

bdraco:
The `C` solution sounds better because it will work long term vs the `DEF` and `IF` are guaranteed to break sooner

Dreamsorcerer:
Are there any plans to expose Py_GIL_DISABLED in Cython? If so, might make more sense to come back and fix it when that happens.

If we do want to write it without Cython, should we consider Rust instead of C?

yarl/_quoting_c.pyx
```diff
@@ -91,7 +92,15 @@ cdef struct Writer:
 
 
 cdef inline void _init_writer(Writer* writer):
-    writer.buf = &BUFFER[0]
+    cdef char *buf
+    if sysconfig.get_config_var("Py_GIL_DISABLED"):
```
Dreamsorcerer, line 96: The benchmark seems to suggest this call takes about half the time of the entire benchmark, maybe we can just store it as a global (the value should surely never change)?

Also, codecov seems to think that this code is never True. Not sure if that's an issue with coverage collection on free-threaded?

bdraco: we should be able to write it as a `cdef bint`

lysnikolaou: I'm not sure how codecov handles different Python versions / builds. It could be that it doesn't recognize it, cause it'll only ever be true under the free-threaded build. Though, I'm not sure whether it knows enough about `get_config_var` and build flags to know that.

Just to make sure, I verified that it is indeed true under the free-threaded build, so we can just ignore this, I think.

yarl/_quoting_c.pyx
```diff
@@ -91,7 +93,16 @@ cdef struct Writer:
 
 
 cdef inline void _init_writer(Writer* writer):
-    writer.buf = &BUFFER[0]
+    cdef char *buf
+    if IS_GIL_DISABLED:
+        print("Calling malloc...")
```
bdraco, line 98:
```suggestion
```

stale print

lysnikolaou: Whoops, sorry.

yarl/_quoting_c.pyx
bdraco: https://app.codecov.io/github/aio-libs/yarl/commit/8b896726b8bd5b740bd4563a70961ba026168361/blob/yarl/_quoting_c.pyx?dropdown=coverage

It looks like we have a tag conflict :

Multiple flags detected. Please ensure one flag per upload.

lysnikolaou: That doesn't seem to be related to this PR. The action indeed uploads multiple flags, but that [seems to be the intended behavior](https://github.com/aio-libs/yarl/blob/426b7ac44e216015687adec4a16218005ecd2606/.github/workflows/ci-cd.yml#L369). Also, you can see [the same error on other commits](https://app.codecov.io/github/aio-libs/yarl/commit/426b7ac44e216015687adec4a16218005ecd2606).

lysnikolaou: I'm not sure. Is this a blocker for this PR? Otherwise, the performance regressions seems to have been resolved with the latest commit.

Dreamsorcerer: I'm not sure, maybe something that @webknjaz might know more about.

I can see 2 uploads tagged for py-3.13.2 (on Linux), these are the regular test runs. In there one of them has coverage for the .pyx file (the other would be the no-extensions run). Under py-3.13.1 tag, there are also 2 uploads, these are the free-threaded runs. Both have full coverage of _quoting_py.py and no coverage of _quoting_c.pyx. This suggests to me that maybe the extensions are not getting executed in the free-threaded build in the CI for some reason...

The file links expire after 30 seconds, but you can search and access them from the sidebar (and just search for the filename in the upload): https://app.codecov.io/github/aio-libs/yarl/commit/8b896726b8bd5b740bd4563a70961ba026168361

Dreamsorcerer: So the CI mentions a Cython coverage plugin:
https://github.com/aio-libs/yarl/blob/8b896726b8bd5b740bd4563a70961ba026168361/.github/workflows/ci-cd.yml#L303

So, maybe it just doesn't work with the alpha release or free-threading yet.

I'd suggest we add that test just to be sure that it is running, and then we can merge, assuming that coverage will likely be fixed by a later release.

lysnikolaou: Opened #1475 for this.

> So, maybe it just doesn't work with the alpha release or free-threading yet.

Cython 3.0.12 is installed during the test run, so that might be what's wrong as well.

bdraco (approved):

webknjaz (approved):

tests/test_quoting_benchmarks.py
```diff
@@ -1,6 +1,11 @@
 """codspeed benchmark for yarl._quoting module."""
 
-from pytest_codspeed import BenchmarkFixture
+import pytest
+
+try:
+    from pytest_codspeed import BenchmarkFixture
+except ImportError:
```
webknjaz, line 7:
```suggestion
except ImportError:  # pragma: no branch  # only hit in cibuildwheel
```

tests/test_url_benchmarks.py
```diff
@@ -1,6 +1,11 @@
 """codspeed benchmarks for yarl.URL."""
 
-from pytest_codspeed import BenchmarkFixture
+import pytest
+
+try:
+    from pytest_codspeed import BenchmarkFixture
+except ImportError:
```
webknjaz, line 7:
```suggestion
except ImportError:  # pragma: no branch  # only hit in cibuildwheel
```

yarl/_quoting_c.pyx
```diff
@@ -91,7 +93,15 @@ cdef struct Writer:
...
 
 cdef inline void _init_writer(Writer* writer):
-    writer.buf = &BUFFER[0]
+    cdef char *buf
+    if IS_GIL_DISABLED:
+        buf = <char *> PyMem_Malloc(BUF_SIZE)
+        if buf == NULL:
+            PyErr_NoMemory()
+            return
+        writer.buf = buf
```
webknjaz, line 102: Why is this branch marked as not covered? https://app.codecov.io/gh/aio-libs/yarl/commit/dcb3ae1102d49a77b4bd824839c9a72c80cc229c?flags%5B0%5D=pytest&dropdown=coverage#fec01998dabbadc38df09608935ec4b7-R98

webknjaz: @lysnikolaou this might be because the test job installs older cython: https://github.com/aio-libs/yarl/actions/runs/13526871108/job/37800294520?pr=1456#step:8:3. Could you try updating it there to see if it helps?

lysnikolaou: This is probably happening because the `quansight-labs/setup-python` action returns the version number without the trailing `t` so one set of coverage results might be overriding the other. I've fixed that.

Also opend #1480 to get rif of the hypothesis warnings.

yarl/_quoting_c.pyx
```diff
@@ -91,7 +93,15 @@ cdef struct Writer:
 
 
 cdef inline void _init_writer(Writer* writer):
-    writer.buf = &BUFFER[0]
+    cdef char *buf
+    if IS_GIL_DISABLED:
+        buf = <char *> PyMem_Malloc(BUF_SIZE)
```
asvetlov, line 98: Could the code always use a thread-local buffer?
In this case, the amortized time will be O(1), like with the static buffer.

lysnikolaou: Implemented in [6be69fc](https://github.com/aio-libs/yarl/pull/1456/commits/6be69fc0797cc9fecc948b9d930d24418212d0de). Could you please have a look and let me know whether this approach seems okay?

.github/workflows/ci-cd.yml
```diff
@@ -240,7 +241,7 @@ jobs:
 
     - name: Setup Python ${{ matrix.pyver }}
       id: python-install
-      uses: actions/setup-python@v5
+      uses: quansight-labs/setup-python@v5
```
Dreamsorcerer, line 244: I think this can be reverted now, I saw a mention that they made a new release this week with support for the t version.

packaging/pep517_backend/_backend.py
webknjaz: It doesn't really matter because of how the compat operator works..

webknjaz (approved):

yarl/_quoting_c.pyx
bdraco: I may have missed the intent here, but I was expecting something like what was asked for in https://github.com/cython/cython/issues/6745

lysnikolaou: Because the buffer is only ever needed within the lifetime of `_do_quote_or_skip`, making it a local to that function effectively makes it a thread-local as well. So we're getting the same benefit without the need for C++'s `thread_local`.

bdraco (approved):

bdraco:
Thanks @lysnikolaou & thanks for your patience with this one

PR #1475, opened 2025-02-25
Add test that C extension is correctly loaded

lysnikolaou:
## What do these changes do?

- Adds a test that the C extensions are loaded okay.

No.


bdraco:
Thanks @lysnikolaou

PR #1480, opened 2025-03-11
Add hidden folders to pytest's norecursedirs

lysnikolaou:
## What do these changes do?

Since we're setting (and not extending) `norecursedirs`, `.hypothesis` should also be there, otherwise we might be getting spurious warnings like the ones in [this test run](https://github.com/aio-libs/yarl/actions/runs/13787597245/job/38559648422?pr=1456#step:16:136).

No.

### DIFF
--- a/CHANGES/1480.contrib.rst
+++ b/CHANGES/1480.contrib.rst
@@ -1,0 +1,2 @@
+Added all hidden folders to pytest's ``norecursedirs`` to prevent it
+from trying to collect tests there -- by :user:`lysnikolaou`.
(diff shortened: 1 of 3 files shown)


webknjaz (commented):
Could you do https://github.com/ansible/awx-plugins/commit/e9bcd0ccd58a49de5b6d4f5f7cf6fb7bfb70438e instead?

lysnikolaou:
Done!

webknjaz (approved):
Thanks! I think this deserves a contrib change note..

lysnikolaou:
Added a contrib change note. Thanks for the suggestion @webknjaz!

bdraco (approved):

PR #1493, opened 2025-04-03
Improve performance of query string parsing

bdraco:
## What do these changes do?

When looking at the performance regression in #1492, I noticed that much of the time was spent in `parse_qsl`, however we do not need all the complexity of `parse_sql` since we use a very limited subset of the functionality.

We can write a faster version using the unquoter built-in to yarl with very little code.

no

## Related issue number
n/a

## Checklist

### DIFF
--- a/CHANGES/1493.misc.rst
+++ b/CHANGES/1493.misc.rst
@@ -1,0 +1,1 @@
+Improved performance of parsing query strings -- by :user:`bdraco`.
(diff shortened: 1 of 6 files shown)


asvetlov:
Good idea!
I suspect all functions from `urllib.parse` could be replaced with faster ones.
The only big question is the compatibility; URL parsing is not a trivial thing.

bdraco:
> Good idea!
> I suspect all functions from `urllib.parse` could be replaced with faster ones.
> The only big question is the compatibility; URL parsing is not a trivial thing.

We are down to very few urllib.parse calls left. The one in https://github.com/aio-libs/yarl/blob/f73f47ec58f082a1aead5342b22cfd3818598ff1/yarl/_quoters.py#L32 could probably be replaced as well

PR #1500, opened 2025-04-06
Increment version to 1.19.1.dev0

bdraco:
waits for https://github.com/aio-libs/yarl/actions/runs/14287805338


bdraco:
performance benchmarks are a side effect of Cython tracing being turned off for the release and than this PR ran right after

Issue #1485, opened 2025-03-18
Unhandled exception (IndexError) in URL parsing


lenix123:
### Please confirm the following

- [x] I agree to follow the [aio-libs Code of Conduct][CoC]

[CoC]: ../../.github/blob/master/CODE_OF_CONDUCT.md

- [x] I have checked the [current issues][issue search] for duplicates.

[issue search]: ../search?q=is%3Aissue&type=issues

- [x] I understand this is open source software provided for free and that I might not receive a timely response.
- [x] I am positive I am **NOT** reporting a (potential) security
vulnerability, to the best of my knowledge. *(These must be shared by
submitting [this report form][vulnerability report form] instead, if
any hesitation exists.)*

[vulnerability report form]: ../security/advisories/new

- [ ] I am willing to submit a pull request with reporoducers as xfailing test cases or even entire fix. *(Assign this issue to me.)*

### Describe the bug

We were fuzzing `URL()` parser and have found IndexError.

### To Reproduce

1. install yarl 1.18.3:
```
pip install --force-reinstall -v "yarl==1.18.3"
```
2. reproduce IndexError:
```
from yarl import URL
URL("//]'z[")
```

### Expected behavior

Parse string without an error

### Logs/tracebacks

```python-traceback
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.10/dist-packages/yarl/_url.py", line 355, in __new__
    return pre_encoded_url(val) if encoded else encode_url(val)
  File "/usr/local/lib/python3.10/dist-packages/yarl/_url.py", line 147, in encode_url
    scheme, netloc, path, query, fragment = split_url(url_str)
  File "/usr/local/lib/python3.10/dist-packages/yarl/_parse.py", line 72, in split_url
    if bracketed_host[0] == "v":
IndexError: string index out of range
```

### Python Version

```console
$ python --version
Python 3.10.12
```

### multidict Version

```console
$ python -m pip show multidict
Version: 6.2.0
```

### propcache Version

```console
$ python -m pip show propcache
Version: 0.3.0
```

### yarl Version

```console
$ python -m pip show yarl
Version: 1.18.3
```

### OS

Ubuntu:22.04

webknjaz:
> ### Expected behavior
> 
> Parse string without an error

[...]

PR #1512, opened 2025-05-22
:bug: more meaningful exception if reversed brackets in ipv6 URL

MaelPic:
## What do these changes do?

For IPv6 URL, if brackets are set in reversed order (closing bracket before open bracket), the raised exception is now `ValueError("Invalid IPv6 URL")`, which is more convenient that the reported one (`IndexError: string index out of range`).

If url contains brackets in opposite order:
Previous behavior: exception `IndexError` is raised
New behavior: exception `ValueError` is raised

## Related issue number

Fixes https://github.com/aio-libs/yarl/issues/1485

## Checklist


tests/test_url.py
```diff
@@ -338,9 +338,18 @@ def test_ipv6_missing_right_bracket() -> None:
         URL("http://[1dec:0:0:0::1/")
 
 
-def test_ipv4_brackets_not_allowed() -> None:
-    with pytest.raises(ValueError, match="An IPv4 address cannot be in brackets"):
-        URL("http://[127.0.0.1]/")
+@pytest.mark.parametrize(
+    ("url"),
```
webknjaz, line 342:
```suggestion
    "url",
```

webknjaz:
note: we agreed with Mael that he'll add a change note later on, while traveling

CHANGES/1512.bugfix.rst
```diff
@@ -0,0 +1,7 @@
+Aligned the type of exception raised for improper IPv6 URL values,
+which should be `ValueError`.
+
+It fixes the issue where exception `IndexError` was raised in the
+following conditions: empty IPv6 URL, brackets in reverse order.
+
+-- by :user:` MaelPic`.
```
webknjaz, line 7:
```suggestion
-- by :user:`MaelPic`
```

CHANGES/1512.bugfix.rst
```diff
@@ -0,0 +1,7 @@
+Aligned the type of exception raised for improper IPv6 URL values,
+which should be `ValueError`.
+
+It fixes the issue where exception `IndexError` was raised in the
+following conditions: empty IPv6 URL, brackets in reverse order.
```
webknjaz, line 5:
```suggestion
These fixes the issue where exception :exc:`IndexError` was
leaking from the internal code because of not being handled and
transformed into a user-facing error. The problem was happening
under the following conditions: empty IPv6 URL, brackets in
reverse order.
```

CHANGES/1512.bugfix.rst
```diff
@@ -0,0 +1,7 @@
+Aligned the type of exception raised for improper IPv6 URL values,
+which should be `ValueError`.
```
webknjaz, line 2:
```suggestion
Started raising a :exc:`ValueError` exception raised for corrupted
IPv6 URL values.
```

tests/test_url.py
```diff
@@ -338,9 +338,20 @@ def test_ipv6_missing_right_bracket() -> None:
...
 
 
-def test_ipv4_brackets_not_allowed() -> None:
-    with pytest.raises(ValueError, match="An IPv4 address cannot be in brackets"):
-        URL("http://[127.0.0.1]/")
+@pytest.mark.parametrize(
+    "url",
+    [
+        "http://[]/",  # Empty IPv6 URL
+        "http://[1]/",  # No semicolon
```
webknjaz, line 345: *(you probably meant “no colon” in there)*

tests/test_url.py
```diff
@@ -338,9 +338,20 @@ def test_ipv6_missing_right_bracket() -> None:
...
+        "http://[127.0.0.1]/",  # IPv4 inside brackets
+        "http://]1dec:0:0:0::1[/",  # Brackets in reversed order
+    ],
```
webknjaz, line 348: Instead of having code comments, it's best to set explicit param ids. They will show up in the test report, and will make it easier to both reason about the testing semantics, plus can be used to select and run one of the test params with `pytest`:
```suggestion
    (
        "http://[]/",
        "http://[1]/",
        "http://[127.0.0.1]/",
        "http://]1dec:0:0:0::1[/",
    ),
    ids=(
        "empty-IPv6-like-URL",
        "no-colons-in-IPv6",
        "IPv4-inside-brackets",
        "brackets-in-reversed-order",
    ),
```

webknjaz (commented):
@MaelPic thanks for your contribution during the PyCon US Sprints last week!

Would you me able to make the following cosmetic changes? Or do you need us to pick up the PR and complete it instead?

yarl/_parse.py
```diff
@@ -69,11 +69,11 @@ def split_url(url: str) -> SplitURLType:
             # Valid bracketed hosts are defined in
             # https://www.rfc-editor.org/rfc/rfc3986#page-49
             # https://url.spec.whatwg.org/
-            if bracketed_host[0] == "v":
+            if bracketed_host.startswith("v"):
```
bdraco, line 72:
```suggestion
            if bracketed_host and bracketed_host[0] == "v":
```

This is faster since there is no function call

Dreamsorcerer (approved):

bdraco:
Thanks @MaelPic

PR #1514, opened 2025-06-03
Use Cython 3.1 universally

lysnikolaou:
## What do these changes do?

- Use Cython 3.1 universally
- Delete unnecessary requirement files

No.

## Checklist


webknjaz (approved):
LGTM functionally, but it'd be good to figure out the benchmark thing. I'll let @bdraco decide whether this is mergeable.

bdraco:
At least for some of them it's now including the Cython tracing overhead and the performance regression is not real.  It will take me a few hours to investigate the rest of them.  I'm at the OHF summit in Dublin this week so I'm quite time limited and trying to only focus on blockers and bugs this week

bdraco:
I will do a deep dive next week when I'm back in Texas

packaging/pep517_backend/_backend.py
```diff
@@ -376,10 +375,8 @@ def get_requires_for_build_wheel(
 
     if is_pure_python_build:
         c_ext_build_deps = []
-    elif sysconfig.get_config_var('Py_GIL_DISABLED'):
-        c_ext_build_deps = ['Cython ~= 3.1.0a1']
     else:
-        c_ext_build_deps = ['Cython >= 3.0.12']
+        c_ext_build_deps = ['Cython >= 3.1.1']
```
bdraco, line 379:
```suggestion
        c_ext_build_deps = ['Cython >= 3.1.2']
```

bdraco:
https://github.com/aio-libs/yarl/pull/1517

Hit the same way. Let me do a PR with line tracing off and see what we get

bdraco (approved):

lysnikolaou:
> @lysnikolaou shouldn't we also use the [`freethreading_compatible`](https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#compiler-directives) setting, by the way?

We're doing that already [in the `.pyx` file itself](https://github.com/aio-libs/yarl/blob/3871d2c46b05d63fe0d91a29468dc3f049e27ebe/yarl/_quoting_c.pyx#L1).

webknjaz:
> > @lysnikolaou shouldn't we also use the [`freethreading_compatible`](https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#compiler-directives) setting, by the way?
> 
> We're doing that already [in the `.pyx` file itself](https://github.com/aio-libs/yarl/blob/3871d2c46b05d63fe0d91a29468dc3f049e27ebe/yarl/_quoting_c.pyx#L1).

@lysnikolaou oh, I see… I wonder if it'd be a good idea to centralize those settings through https://github.com/aio-libs/yarl/blob/3871d2c/pyproject.toml#L22-L37 so that individual pyx files wouldn't need a copy of that comment.

lysnikolaou:
> I wonder if it'd be a good idea to centralize those settings through [`3871d2c`/pyproject.toml#L22-L37](https://github.com/aio-libs/yarl/blob/3871d2c/pyproject.toml?rgh-link-date=2025-06-10T14%3A56%3A39Z#L22-L37) so that individual pyx files wouldn't need a copy of that comment.

Yeah, that's a good point. This wasn't possible up until now, because we were using different versions of Cython that did not have support for the `freethreading_compatible` directive, but it's now okay since we upgrade to 3.1. Opened #1528.

PR #1519, opened 2025-06-09
DNM: Test PR with line tracing disabled

bdraco:


bdraco:
> @bdraco are you going to download the wheels from GHA and test them locally?

yes

bdraco:
There is still something going wrong here.

When I install yarl 1.20.0 from PyPI I get 0.036 on quoting ascii benchmark.

When install the wheels with tracing disabled here I get 0.073....

bdraco:
Now I downloaded the latest build and its 0.020 sec ....

PR #1520, opened 2025-06-09
Build benchmark wheels without Cython tracing

bdraco:
## Summary

This PR modifies the benchmark workflow to build wheels without Cython tracing enabled, ensuring benchmarks always measure production performance.

## Motivation

Currently, the benchmark workflow uses pre-built wheels from the CI pipeline. These wheels have Cython line tracing enabled for development builds (non-tagged commits) to support code coverage collection. However, this creates two issues:

1. **Inconsistent performance measurements**: Development builds run massively slower due to tracing overhead, making it impossible to accurately compare performance changes over time
2. **Not representative of user experience**: Users install production wheels without tracing, so benchmarks should measure the same configuration

## Changes

- Modified the `benchmark` job to build yarl locally with `--config-settings=with-cython-tracing=false`
- Removed dependency on pre-built wheel artifacts
- Simplified the workflow by eliminating the complex wheel selection logic

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -381,28 +381,12 @@ jobs:
   benchmark:
     name: Benchmark
     needs:
-    - build-pure-python-dists  # transitive, for accessing settings
-    - build-wheels-for-tested-arches
     - pre-setup  # transitive, for accessing settings
     runs-on: ubuntu-latest
     timeout-minutes: 5
     steps:
     - name: Checkout project
       uses: actions/checkout@v4
-    - name: Retrieve the project source from an sdist inside the GHA artifact
-      uses: re-actors/checkout-python-sdist@release/v2
-      with:
-        source-tarball-name: >-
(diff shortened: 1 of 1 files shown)


bdraco:
I think we still want python debug symbols though

.github/workflows/ci-cd.yml
```diff
@@ -422,47 +406,8 @@ jobs:
...
-        --dry-run
-        --report=-
-        --quiet
-        | jq --raw-output .install[].download_info.url
-        | tee -a "${GITHUB_OUTPUT}"
-      shell: bash
-    - name: Self-install
-      run: python -Im pip install '${{ steps.wheel-file.outputs.path }}'
+    - name: Build and install yarl without Cython tracing
+      run: python -Im pip install . --config-settings=with-cython-tracing=true
```
bdraco, line 410:
```suggestion
      run: python -Im pip install . --config-settings=with-cython-tracing=false
```

PR #1521, opened 2025-06-09
Only enable line tracing when building with Cython tracing

bdraco:
## What do these changes do?

Only enable line tracing when building with Cython tracing

## Checklist

### DIFF
--- a/packaging/pep517_backend/_cython_configuration.py
+++ b/packaging/pep517_backend/_cython_configuration.py
@@ -18,7 +18,7 @@ from ._transformers import get_cli_kwargs_from_config, get_enabled_cli_flags_fro
 class Config(TypedDict):
     env: dict[str, str]
     flags: dict[str, bool]
-    kwargs: dict[str, str]
+    kwargs: dict[str, str | dict[str, str]]
     src: list[str]
 
 
@@ -79,10 +79,18 @@ def get_local_cython_config() -> Config:
     return config_mapping['tool']['local']['cythonize']  # type: ignore[no-any-return]
 
 
-def make_cythonize_cli_args_from_config(config: Config) -> list[str]:
+def make_cythonize_cli_args_from_config(config: Config, cython_line_tracing_requested: bool = False) -> list[str]:
     py_ver_arg = f'-{_python_version_tuple.major!s}'
 
     cli_flags = get_enabled_cli_flags_from_config(config['flags'])
+
+    # If line tracing is requested, add it to the directives
+    if cython_line_tracing_requested:
+        directives = config['kwargs'].setdefault('directive', {})
+        assert isinstance(directives, dict)  # Type narrowing for mypy
+        directives['linetrace'] = 'True'
+        directives['profile'] = 'True'
+
     cli_kwargs = get_cli_kwargs_from_config(config['kwargs'])
 
     return cli_flags + [py_ver_arg] + cli_kwargs + ['--'] + config['src']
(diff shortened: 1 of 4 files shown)


packaging/pep517_backend/_transformers.py
```diff
@@ -17,7 +17,9 @@ def _emit_opt_pairs(opt_pair: tuple[str, Union[dict[str, str], str]]) -> Iterato
     yield from ("=".join(map(str, (flag_opt,) + pair)) for pair in sub_pairs)
 
 
-def get_cli_kwargs_from_config(kwargs_map: dict[str, str]) -> list[str]:
+def get_cli_kwargs_from_config(
+    kwargs_map: dict[str, Union[str, dict[str, str]]],
```
webknjaz, line 21: Why is this nested?

packaging/pep517_backend/_cython_configuration.py
```diff
@@ -79,10 +79,18 @@ def get_local_cython_config() -> Config:
...
 
-def make_cythonize_cli_args_from_config(config: Config) -> list[str]:
+def make_cythonize_cli_args_from_config(config: Config, cython_line_tracing_requested: bool = False) -> list[str]:
     py_ver_arg = f'-{_python_version_tuple.major!s}'
 
     cli_flags = get_enabled_cli_flags_from_config(config['flags'])
+
+    # If line tracing is requested, add it to the directives
+    if cython_line_tracing_requested:
+        directives = config['kwargs'].setdefault('directive', {})
```
webknjaz, line 89: Can we move `config['kwargs']` into a dedicated var, as it's also used in the return statement?

packaging/pep517_backend/_transformers.py
bdraco: that doesn't make sense though...

bdraco: Yeah it is correct.  we want `--directive=linetrace=True` so it needs to be inside another dict

bdraco: it was hard to capture so I tested with
```diff
diff --git a/packaging/pep517_backend/_cython_configuration.py b/packaging/pep517_backend/_cython_configuration.py
index da7a4f8..a2a0f4c 100644
--- a/packaging/pep517_backend/_cython_configuration.py
+++ b/packaging/pep517_backend/_cython_configuration.py
@@ -8,6 +8,7 @@ from contextlib import contextmanager
 from pathlib import Path
 from sys import version_info as _python_version_tuple
 from typing import TypedDict
+from sys import stderr as _standard_error_stream
 
 from expandvars import expandvars
 
@@ -93,6 +94,9 @@ def make_cythonize_cli_args_from_config(config: Config, cython_line_tracing_requ
 
     cli_kwargs = get_cli_kwargs_from_config(config['kwargs'])
 
+    print(['cli_kwargs', cli_kwargs], file=_standard_error_stream)
+
+
     return cli_flags + [py_ver_arg] + cli_kwargs + ['--'] + config['src']
 
 
```

packaging/pep517_backend/_cython_configuration.py
bdraco: It mutating the `dict` so it will be captured in the return statement

webknjaz: Yes, that's what I was thinking.

packaging/pep517_backend/_cython_configuration.py
```diff
@@ -79,10 +79,18 @@ def get_local_cython_config() -> Config:
...
+        assert isinstance(directives, dict)  # Type narrowing for mypy
+        directives['linetrace'] = 'True'
+        directives['profile'] = 'True'
```
webknjaz, line 92: Looking into this more, I'd probably even move this entire block into a helper function mutating `config_kwargs`.

PR #1529, opened 2025-06-10
Unquoting improvements

Vizonex:
## What do these changes do?

How do I keep finding things to optimize??? I got rid of the list in the unquoter and I replaced  it with a custom unicode writer I wrote. I'm gonna allow anybody to use it as they wish since it's got a feature where it's able to dynamically handle unicode characters but now we can get rid of needing to join a bunch of python objects all at one time. 

There's a new technique in the Unquoter that allocates the same amount of memory as the string it's taking in to be examined and decoded. This should be an optimization and not a performance bottleneck now.

## Checklist


Vizonex:
how is that a regression do I stand corrected?

PR #1538, opened 2025-06-26
Add the URL subtraction feature

### LINKED ISSUE #1183: Add Support for Relative Path Calculation

babieiev:
## What do these changes do?

Rework of the URL subtraction feature (added in #1340; removed in #1391)

Being able to calculate the relative path between two URLs:

```py
from yarl import URL

target_url = URL("http://example.com/path/index.html")
base_url = URL("http://example.com/path/")

relative_url = target_url.relative_to(base_url)

print(relative_url)  # output: "index.html"
```

As suggested in https://github.com/aio-libs/yarl/pull/1392#pullrequestreview-2441220999, the idea is (temporarily?) implemented as a regular method, not a subtraction operator.

## Related issue number

Resolves #1183 

## Checklist

## Known issues

- [x] Time complexity is roughly O(base_path_segments * target_path_segments)
- [x] Empty segments are not handled correctly #1388 
- [x] Potentially handling special cases with leading/trailing slashes #1388

### DIFF
--- a/CHANGES/1538.feature.rst
+++ b/CHANGES/1538.feature.rst
@@ -1,0 +1,15 @@
+Added a new method :py:meth:`~yarl.URL.relative_to`
+to get the relative path between URLs.
+
+Note that both URLs must have the same scheme, user, password, host and port:
+
+.. code-block:: pycon
+
+    >>> target = URL("http://example.com/path/to")
+    >>> base = URL("http://example.com/")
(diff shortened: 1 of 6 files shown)


commonism:
+1 for making it look trivial

PR #1586, opened 2025-10-26
📦 Make release builds compile very optimized

webknjaz:
This patch improves the in-tree build backend to allow the end-users append `CFLAGS` and `LDFLAGS` by setting respecitve environment variables.

It additionally sets up default compiler flags to perform build with maximum optimization in release mode. And when line tracing is requested, the compiler and linker flags are configured to include as much information as possible for debugging and coverage tracking.

As a result, the C-extensions published to PyPI are going to be small while their development/testing environment variants will be big.

yarl's footprint is smaller, and it should run faster.

## Checklist

### DIFF
--- a/CHANGES/1586.contrib.rst
+++ b/CHANGES/1586.contrib.rst
@@ -1,0 +1,1 @@
+1586.packaging.rst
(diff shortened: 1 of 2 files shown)


rossburton:
To ensure it's here too: this breaks ~all Linux distributions where we want to build moderately optimised binaries with debug symbols, and split the symbols out ourselves. I filed #1592 with further details.

PR #1607, opened 2025-12-12
Add pydantic support

asvetlov:
## What do these changes do?

`yarl.URL` could be used in `pydantic` models without a boilrplate.

The change doesn't break any existing code but helps with URL serialization/deserialization of `pydantic` models if the library was installed by user.

## Related issue number

None

## Checklist

- [x ] I think the code is well written

### DIFF
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -5,4 +5,4 @@
 #
 # After implementing PEP 780 '"free-threading" not in sys_abi_features' could
 # be used instead of extraction of test-cibuildwheel.txt file.
-pydantic>=2.0
+pydantic>=2.0; sys_platform == 'linux'


tests/test_pydantic.py
```diff
@@ -0,0 +1,40 @@
...
+    assert m == TstModel(url=url)
+    assert isinstance(m.url, URL)
+
+
+def test_validate_invalid() -> None:
+    dct = {"url": 123}
+    with pytest.raises(ValidationError, match="url"):
+        TstModel.model_validate(dct)
+
+
```
bdraco: nit: Would be nice to add a test for json round trip here

bdraco:
LGTM.

I'll look at fixing the CI later today

asvetlov:
Let's run pydantic tests for Linux only on CI, it should be enough

PR #1617, opened 2025-12-16
Bump version to the next dev

asvetlov:


bdraco:
Looks like we already did this

PR #1630, opened 2026-02-13
Mask password in URL.__repr__ for safer debugging output #1629

### LINKED ISSUE #1629: Mask password in `URL.__repr__` for safer debugging output

Currently, `__repr__` of `yarl.URL` (from [yarl](https://github.com/aio-libs/yarl)) includes the raw password when a URL contains user credentials:

```python
>>> from yarl import URL
>>> URL("https://user:secret@example.com")
URL('https://user:secret@example.com')
```

This behavior can unintentionally expose sensitive credentials in logs, tracebacks, debugging sessions, or error reporting systems where `repr()` is implicitly used.

`__repr__` should mask the password component, similar to how many HTTP clients and URL tooling libraries handle credential rendering. For example:

```python
>>> from yarl import URL
>>> URL('https://user:secret@example.com')
URL('https://user:********@example.com')
```

or another clearly redacted placeholder.

### Rationale

* `repr()` is frequently surfaced in logs and diagnostics.
* Accidental credential leakage via logging is a common operational risk.
* Masking credentials in representations aligns with the principle of least exposure.
* This change makes the ecosystem incrementally ("epsilon") safer with minimal behavioral disruption.

Importantly, this proposal affects only the string representation — not the internal storage or functional behavior of the URL object.

### Backward Compatibility Considerations

* The change alters the exact `repr()` output, which may affect strict string-based tests.
* However, this impact is limited and security-positive.
* If necessary, the full URL (including password) remains accessible via explicit API accessors or by `str()`.

Creating a subclass from the `yarl.URL` to implement masking logic. However, inheriting from the `URL` class is forbidden:
```python
>>> class MyUrl(yarl.URL):
...     def __repr__(self):
...         return "SECRET"
... 
  File "<input>", line 1, in <module>
  File "********/.venv/Lib/site-packages/yarl/_url.py", line 482, in __init_subclass__
    raise TypeError(f"Inheriting a class {cls!r} from URL is forbidden")
TypeError: Inheriting a class <class '__main__.MyUrl'> from URL is forbidden
 
```

### Additional context

We can make a PR request for this specific change. Below is a patch of the required changes:

```patch
Index: yarl/_parse.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/yarl/_parse.py b/yarl/_parse.py
--- a/yarl/_parse.py	(revision 7829b01ec6217349dfce3a1dd3190d29ddbb14a9)
+++ b/yarl/_parse.py	(date 1770988382994)
@@ -161,6 +161,7 @@
     host: str | None,
     port: int | None,
     encode: bool = False,
+    mask_password: bool = False,
 ) -> str:
     """Make netloc from parts.
 
@@ -176,6 +177,8 @@
     if user is None and password is None:
         return ret
     if password is not None:
+        if mask_password:
+            password = "********"
         if not user:
             user = ""
         elif encode:
Index: tests/test_url.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/tests/test_url.py b/tests/test_url.py
--- a/tests/test_url.py	(revision 7829b01ec6217349dfce3a1dd3190d29ddbb14a9)
+++ b/tests/test_url.py	(date 1770988686742)
@@ -57,13 +57,13 @@
 
 
 def test_str() -> None:
-    url = URL("http://example.com:8888/path/to?a=1&b=2")
-    assert str(url) == "http://example.com:8888/path/to?a=1&b=2"
+    url = URL("http://user:password@example.com:8888/path/to?a=1&b=2")
+    assert str(url) == "http://user:password@example.com:8888/path/to?a=1&b=2"
 
 
 def test_repr() -> None:
-    url = URL("http://example.com")
-    assert "URL('http://example.com')" == repr(url)
+    url = URL("http://user:password@example.com")
+    assert "URL('http://user:********@example.com')" == repr(url)
 
 
 def test_origin() -> None:
Index: yarl/_url.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/yarl/_url.py b/yarl/_url.py
--- a/yarl/_url.py	(revision 7829b01ec6217349dfce3a1dd3190d29ddbb14a9)
+++ b/yarl/_url.py	(date 1770990305586)
@@ -491,24 +491,31 @@
     def __init_subclass__(cls) -> NoReturn:
         raise TypeError(f"Inheriting a class {cls!r} from URL is forbidden")
 
-    def __str__(self) -> str:
+    def _make_string(self, mask_password: bool = False) -> str:
         if not self._path and self._netloc and (self._query or self._fragment):
             path = "/"
         else:
             path = self._path
-        if (port := self.explicit_port) is not None and port == DEFAULT_PORTS.get(
-            self._scheme
+        if (
+            (port := self.explicit_port) is not None
+            and port == DEFAULT_PORTS.get(self._scheme)
+            or mask_password
         ):
             # port normalization - using None for default ports to remove from rendering
             # https://datatracker.ietf.org/doc/html/rfc3986.html#section-6.2.3
             host = self.host_subcomponent
-            netloc = make_netloc(self.raw_user, self.raw_password, host, None)
+            netloc = make_netloc(
+                self.raw_user, self.raw_password, host, None, mask_password=True
+            )
         else:
             netloc = self._netloc
         return unsplit_result(self._scheme, netloc, path, self._query, self._fragment)
 
+    def __str__(self) -> str:
+        return self._make_string(mask_password=False)
+
     def __repr__(self) -> str:
-        return f"{self.__class__.__name__}('{str(self)}')"
+        return f"{self.__class__.__name__}('{self._make_string(mask_password=True)}')"
 
     def __bytes__(self) -> bytes:
```

This patch was kindly supported by [Sopra Steria](https://www.soprasteria.nl/) Pythoneers

### Code of Conduct

- [x] I agree to follow the aio-libs Code of Conduct

mbaas038:
## What do these changes do?

- `yarl.URL.__repr__` exposed passwords in URLs, risking credential leakage in logs, tracebacks, or debugging output.
- The password component is masked in `__repr__` while leaving other URL parts unchanged.
- Added conditional masking logic to `__repr__` that replaces the password with a placeholder (`********`) if present. 
- Improves security by preventing accidental exposure. 
- Does **not** affect URL functionality or internal storage.  

- Slight change in `repr()` output — may affect tests relying on exact string output.

## Related issue number

- Fixes #1629 

## Checklist

This PR was kindly supported by [Sopra Steria](https://www.soprasteria.nl/) Pythoneers


yarl/_url.py
```diff
@@ -491,24 +491,31 @@ def build(
...
+    def _make_string(self, mask_password: bool = False) -> str:
         if not self._path and self._netloc and (self._query or self._fragment):
             path = "/"
         else:
             path = self._path
-        if (port := self.explicit_port) is not None and port == DEFAULT_PORTS.get(
-            self._scheme
+        if (
+            (port := self.explicit_port) is not None
+            and port == DEFAULT_PORTS.get(self._scheme)
```
jhbuhrman: Small remark from the co-author of this PR: perhaps it is better to surround the `and`-expression with parentheses. The functionality won't change, it is purely for reasons of readability and maintainability.

```suggestion
            ((port := self.explicit_port) is not None
             and port == DEFAULT_PORTS.get(self._scheme))
```
(trusting on `ruff` for the formatting)

yarl/_url.py
```diff
@@ -491,24 +491,30 @@ def build(
...
+        ) or mask_password:
             # port normalization - using None for default ports to remove from rendering
             # https://datatracker.ietf.org/doc/html/rfc3986.html#section-6.2.3
             host = self.host_subcomponent
-            netloc = make_netloc(self.raw_user, self.raw_password, host, None)
+            netloc = make_netloc(
+                self.raw_user, self.raw_password, host, None, mask_password=True
```
jhbuhrman:
```suggestion
                self.raw_user, self.raw_password, host, None, mask_password=mask_password
```

bdraco:
@aiolibsbot review

yarl/_url.py
```diff
@@ -491,24 +491,34 @@ def build(
...
         else:
             netloc = self._netloc
         return unsplit_result(self._scheme, netloc, path, self._query, self._fragment)
 
+    def __str__(self) -> str:
+        return self._make_string(mask_password=False)
+
     def __repr__(self) -> str:
-        return f"{self.__class__.__name__}('{str(self)}')"
+        return f"{self.__class__.__name__}('{self._make_string(mask_password=True)}')"
```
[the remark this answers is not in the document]
aiolibsbot: Worth raising, but I would push back on changing the repr shape in this PR for two reasons.

**Backward compatibility.** `URL.__repr__` has returned `URL('...')` for the entire life of the library; switching to `<URL ... (masked)>` for any URL with a password is a breaking change for every downstream test, log scraper, and snapshot file that pattern-matches the current shape. The whole point of this PR is to land a security improvement without breaking users, and a conditional repr format change actively works against that — code that used to get one shape now gets a different shape depending on whether userinfo is present, which is harder to reason about than the current uniform format.

**Python convention is a SHOULD, not a MUST.** The docs say "If at all possible" the repr should be eval-able and otherwise "a string of the form `<...>` should be returned." There is well-established precedent for objects whose repr is not round-trippable but still uses the `Class(...)` form — `datetime.datetime(..., tzinfo=...)` with a custom tzinfo, `pathlib.Path` on Windows with a `WindowsPath` repr that does not work on POSIX, `re.compile`'s repr, etc. `URL('http://user:********@host')` falling into that category is defensible.

If you do want eval-correctness preserved when there is no password, the cleanest path is to make the mask a no-op when `raw_password is None` (the current code already does this implicitly, since `make_netloc` only touches `password` when it is not `None`) and leave the `Class('...')` shape alone in both cases. That is what the PR is doing today; the lossy case is bounded to URLs that actually have a password, which are the only URLs whose repr should not be eval'd anyway. Worth a maintainer call, but I would not block this PR on it — and there is a higher-priority correctness bug on the port-drop case to fix first.

PR #1643, opened 2026-03-14
fix: handle uuid.UUID conversion in query params

### LINKED ISSUE #1458: Convert an `uuid.UUID` variable into `int` when pass it in params

- [X] I agree to follow the [aio-libs Code of Conduct][CoC]

- [X] I have checked the [current issues][issue search] for duplicates.

- [X] I understand this is open source software provided for free and that I might not receive a timely response.
- [X] I am positive I am **NOT** reporting a (potential) security
any hesitation exists.)*

- [X] I am willing to submit a pull request with reporoducers as xfailing test cases or even entire fix. *(Assign this issue to me.)*

### Describe the bug

I've defined a pydantic model (let's call it `User`) as the input model of my view function, which has a field annotated as  uuid.UUID named `user_id`.

When I pass user.user_id as a part of params in a request(use aiohttp), yarl will convert the type of uuid.UUID into int before cast it into string.

For example: uuid.UUID('3199712f-1b78-4420-852b-a73ee09e6a8f') -> '65928888857327045292976149998723820175'

by this line: https://github.com/aio-libs/yarl/blob/master/yarl/_query.py#L35

So i've change my code to check every param in params if it's instance of uuid.UUID.

I came here to ask if this conversion behavior is as expected? Is there a better compatibility implementation?

### To Reproduce

as description above

### Expected behavior

as description above

### Logs/tracebacks

```python-traceback
no need
```

### Python Version

```console
$ python --version
3.11
```

### multidict Version

```console
6.1.0
```

### propcache Version

```console
0.2.1
```

### yarl Version

```console
$ python -m pip show yarl
1.18.3
```

### OS

linux or macos

### Additional context

_No response_

toroleapinc:
This PR fixes the issue where uuid.UUID objects were being converted to integers instead of their string representation when used as query parameter values.

## Problem
When passing a uuid.UUID object as a query parameter, yarl was converting it to an integer because UUID implements __int__(), making it match the SupportsInt protocol. This resulted in URLs like:

Expected: http://example.com?user_id=3199712f-1b78-4420-852b-a73ee09e6a8f  
Actual:   http://example.com?user_id=65928888857327045292976149998723820175

## Solution
Added explicit handling for uuid.UUID objects in the query_var function to prioritize string conversion over integer conversion. The fix:

1. Checks for isinstance(v, uuid.UUID) before isinstance(v, SupportsInt)
2. Uses str(v) for UUID objects, which returns the canonical UUID string format
3. Maintains backward compatibility for all other types

## Changes
- yarl/_query.py: Added uuid import and UUID handling in query_var()
- tests/test_update_query.py: Added comprehensive UUID tests covering single values, multiple values, and mixed types
- CHANGES/1458.bugfix.rst: Added changelog entry

## Testing
Added test cases covering:
- Single UUID in query parameters
- Multiple UUID values for same parameter  
- UUID mixed with other data types
- Both with_query() and update_query() methods

Fixes #1458


toroleapinc:
bumping this one — the only failing check is `codecov/project/typing` but `codecov/patch` is green and the new code (the `isinstance(v, uuid.UUID)` branch and the `uuid.UUID` addition to `SimpleQuery`) is fully type-annotated, so the project-wide threshold drop doesn't seem to be from this diff. happy to look further if you spot something though.

bdraco:
uuid as int is very likely the wrong behavior. I think everything wants a string but this is technically a breaking change so the change note needs to be .breaking.rst

bdraco:
@aiolibsbot review

PR #1646, opened 2026-03-26
fix: mask password in __repr__ (#1630), UUID as str in query (#1643), split_url perf

naarob:
## Summary

Three improvements to `yarl`:

---

### Fix #1630 — Password exposed in `URL.__repr__`

`URL.__repr__` returned the full URL string including plaintext passwords, risking credential leakage in logs, tracebacks, and debug output.

**Fix:** If `self.password is not None`, use `with_password('********')` to build the repr string. `str()`, `==`, and all other operations remain unaffected.

```python
# Before
repr(URL('http://user:s3cr3t@example.com'))  # URL('http://user:s3cr3t@example.com')

# After
repr(URL('http://user:s3cr3t@example.com'))  # URL('http://user:********@example.com')
```

---

### Fix #1643 — `uuid.UUID` converted to `int` in query params

`uuid.UUID` implements `__int__()`, so it matched the `SupportsInt` protocol and was silently converted to a 128-bit integer instead of its dash-separated string representation.

**Fix:** In `query_var()`, check `type(v).__str__ is not object.__str__` before falling back to `int()`. This is a general solution: any type that defines a meaningful `__str__` (not inherited from `object`) will use its string representation.

```python
# Before
URL('http://x.com').with_query({'id': uuid4()})  # ?id=24197857...

# After
URL('http://x.com').with_query({'id': uuid4()})  # ?id=12345678-1234-...
```

---

### Perf — `split_url()` delimiter search

Replaced the `for c in delim_chars` loop with 3 conditional `find()` calls using the already-computed `has_hash` / `has_question_mark` flags. Eliminates Python loop overhead and avoids redundant `find()` calls for URLs without query strings or fragments.

**Tests:** 850 passed — no regressions.

### DIFF
--- a/yarl/_parse.py
+++ b/yarl/_parse.py
@@ -51,10 +51,19 @@ def split_url(url: str) -> SplitURLType:
             delim_chars = "/#"
         else:
             delim_chars = "/"
-        for c in delim_chars:  # look for delimiters; the order is NOT important
-            wdelim = url.find(c, 2)  # find first of this delim
-            if wdelim >= 0 and wdelim < delim:  # if found
-                delim = wdelim  # use earliest delim position
+        # Perf: find each delimiter independently and take the minimum.
+        # Avoids repeated character iteration and Python loop overhead.
+        slash = url.find("/", 2)
+        if slash >= 0:
+            delim = slash
+        if has_question_mark:
+            q = url.find("?", 2)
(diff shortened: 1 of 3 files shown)


bdraco:
Please split this into 3 separate PRs

PR #1650, opened 2026-04-07
fix code-coverage in tests specifically the test_quoting module

Vizonex:
## What do these changes do?

This is a fix based on something I tried to help someone review rather recently #1642. I came to the conclusion that this module was the culprit for some the brokenness being experienced so I attempted to come up with a tiny solution to try and solve a large sum of it that requires a bit less code. Know that I didn't find a way to retain some ids but I'm sure this change is rather negligible. The percentage of code covered goes up by 1% but it should be enough to help #1642 pass.

This change should only affect how contributions are handled as well as relieve us of some problems. [There is a similar problem with multidict](https://github.com/aio-libs/multidict/pull/1314) that I attempted to also get my hands into rather recently. While waiting this is where I decided to go to next so now I am here, seems yarl contains similar issues where coverage is almost completely broken in some areas.

## Related issue number

#1642 

## Checklist


bdraco:
@aiolibsbot review

PR #1653, opened 2026-04-12
Fix IPv6 Zone ID decoding to correctly handle RFC 6874 %25 separator

rodrigobnogueira:
## What do these changes do?

Fixes incorrect decoding of IPv6 Zone IDs in URLs that use the
:rfc:`6874` ``%25`` zone separator.

Two call sites in ``yarl/_url.py`` are updated:

* ``_encode_host()``: when ``%25`` is present in the host string,
  partition on ``%25`` (the RFC 6874 separator) instead of bare
  ``%``. The raw form (``raw_host`` / ``host_subcomponent`` /
  ``str(url)``) is preserved verbatim.
* ``.host`` property: when an IP-literal host contains ``%25``,
  decode it back to ``%`` so callers receive the human-readable
  zone identifier (e.g. ``fe80::1%1`` rather than ``fe80::1%251``).
  The same decoding flows through ``.authority`` and
  ``.human_repr()``.

The pre-fix behaviour, where ``_encode_host`` split on bare ``%``,
is still accepted as a fallback so that hosts constructed
programmatically (e.g. ``with_host("fe80::1%1")``) keep working.

| URL | ``.raw_host`` | ``.host`` before | ``.host`` after |
| --- | --- | --- | --- |
| ``http://[fe80::1%251]/`` | ``fe80::1%251`` | ``fe80::1%251`` | ``fe80::1%1`` |
| ``http://[fe80::1%25eth0]/`` | ``fe80::1%25eth0`` | ``fe80::1%25eth0`` | ``fe80::1%eth0`` |

Yes. ``.host``, ``.authority``, and ``.human_repr()`` now return
the decoded zone identifier (``%`` instead of ``%25``) for IPv6
URLs that carry an RFC 6874 zone separator. ``raw_host``,
``host_subcomponent``, and ``str(url)`` are unchanged, so the
on-the-wire representation is preserved.

No. The change is localised to two small spots in ``yarl/_url.py``;
the dual-path encoder fallback preserves backward compatibility for
programmatically constructed hosts. While not a practical SSRF
vector (zone IDs are link-local scope only), the previous decoding
disagreed with other RFC 6874-aware parsers.

## Related issue number

No linked issue; surfaced while reviewing :rfc:`6874` compliance
of the host parser.

## Checklist

- [ ] If you provide code modification, please add yourself to ``CONTRIBUTORS.txt`` (N/A; no ``CONTRIBUTORS.txt`` in this repo)
- [x] Add a new news fragment into the ``CHANGES/`` folder
  * name: ``1653.bugfix.rst``
  * category: ``bugfix``


rodrigobnogueira:
The codecov/project/typing check is failing at 97.63% against a 100% target, but this is a pre-existing issue unrelated to this PR.

bdraco:
@aiolibsbot review

yarl/_url.py
```diff
@@ -1557,7 +1560,10 @@ def _encode_host(host: str, validate_host: bool) -> str:
     # If the host ends with a digit or contains a colon, its likely
     # an IP address.
     if host and (host[-1].isdigit() or ":" in host):
-        raw_ip, sep, zone = host.partition("%")
+        if "%25" in host:
+            raw_ip, sep, zone = host.partition("%25")
+        else:
+            raw_ip, sep, zone = host.partition("%")
```
bdraco:
```suggestion
        part = "%25" if "%25" in host else "%"
        raw_ip, sep, zone = host.partition(part)
```

bdraco: But we need to do an analysis to see if its safe to decode here

bdraco:
@aiolibsbot review

yarl/_url.py
rodrigobnogueira: Safe: ``raw_host`` / ``host_subcomponent`` / ``str(url)`` keep the encoded form, so wire format is unchanged. The decode only fires on ``.host`` (and through it ``.authority`` / ``.human_repr``), is restricted to ``%25`` (``%2F``, ``%3A`` etc. are not touched, so no smuggling of ``/`` ``:`` ``@`` ``?`` ``#``), and only ever applies to bytes after the zone separator; the IP literal is parsed through ``ipaddress.ip_address`` before that. Zone IDs are link-local per RFC 6874 §3 so they never leave the host. The decoded form (``fe80::1%eth0``) is also what stdlib ``socket`` / ``ipaddress`` accept, which is the usual reason callers reach for ``.host``.

bdraco (approved):

bdraco:
Thanks @rodrigobnogueira

PR #1655, opened 2026-04-12
Fix host validation: IPv6 zone ID characters and NFKC percent bypass

rodrigobnogueira:
## What do these changes do?

Fixes two incomplete validation cases in host parsing.

---

### Finding 1 — IPv6 zone ID character validation

**The bug:** `URL.build()` with `validate_host=True` (the default) did not validate the zone ID portion of IPv6 addresses. Any character — including CR, LF, null bytes — could be embedded in `url.host` via `URL.build(host='::1%<bad>')`. This creates an asymmetry:

```python
# Correctly rejected:
URL.build(scheme='http', host='example.com\r\nX-Injected: evil')
# ValueError: Host '...' cannot contain '\r' ...

# Was NOT rejected (bug):
URL.build(scheme='http', host='::1%\r\nX-Injected: evil', path='/')
# url.host = '::1%\r\nX-Injected: evil'
```

**Root cause:** `_encode_host()` validates the IP address portion via `ip_address()` but the zone string was concatenated back verbatim without any character inspection.

**Fix:** Added `_ZONE_ID_UNSAFE_RE` to reject ASCII control characters (CTL, `\x00–\x1f` and `\x7f`) in the zone ID. Per [RFC 4007 §11.2](https://datatracker.ietf.org/doc/html/rfc4007#section-11.2), zone IDs are OS-specific text strings with no defined format. [RFC 9844 §6.3](https://datatracker.ietf.org/doc/html/rfc9844#section-6-3) recommends rejecting characters inappropriate for the environment; for yarl we reject ASCII control characters. All other characters — spaces, Unicode, parentheses, etc. — are accepted.

---

### Finding 2 — NFKC fullwidth/small percent sign bypass

**The bug:** `_check_netloc()` normalizes the netloc via NFKC and checks for URL-reserved characters, but `%` was missing from the checked set. U+FF05 (FULLWIDTH PERCENT SIGN `％`) and U+FE6A (SMALL PERCENT SIGN `﹪`) both normalize to `%` under NFKC, so they passed validation and ultimately produced a literal `%` in `url.host` via the standard library IDNA fallback in `_idna_encode()`:

```python
URL('http://evil.com\uff052e.internal/').host
# 'evil.com%2e.internal'   ← literal % injected
```

**Root cause:** The `idna` library correctly rejects `%` as invalid in a hostname label, but the `except UnicodeError` fallback to standard library `host.encode('idna')` does its own NFKC normalization and silently accepts the character.

**Fix:** Add `%` to the character set checked after NFKC normalization in `_check_netloc()`.

---

### Changes

| File | Change |
|---|---|
| `yarl/_url.py` | Add `_ZONE_ID_UNSAFE_RE` regex; validate zone ID in `_encode_host()` by rejecting CTL characters |
| `yarl/_parse.py` | Add `%` to NFKC character check in `_check_netloc()` |
| `tests/test_url_build.py` | Tests for zone ID CTL rejection (CRLF, null byte) and acceptance of non-CTL (spaces, Unicode, parens) |
| `tests/test_url.py` | Tests for U+FF05 and U+FE6A being rejected |


tests/test_url_build.py
```diff
@@ -35,6 +35,21 @@ def test_url_ipv4_in_ipv6() -> None:
...
+@pytest.mark.parametrize(
+    ("zone", "desc"),
+    (
+        ("\r\nX-Injected: evil", "crlf-injection"),
+        ("\x00evil", "null-byte"),
+        ("zone with spaces", "spaces"),
+    ),
+    ids=("crlf-injection", "null-byte", "spaces"),
+)
+def test_url_build_ipv6_zone_id_invalid_chars(zone: str, desc: str) -> None:
```
Dreamsorcerer: desc is unused? There's already a description via the ids.

yarl/_url.py
```diff
@@ -89,6 +89,10 @@
     re.VERBOSE,
 )
 
+# RFC 6874 ZoneID = 1*( unreserved / pct-encoded )
```
Dreamsorcerer: Can we first link to the relevant part of an RFC (the new ones allow linking to any paragraph), and second use the latest RFC. This one was obsoleted by RFC 9844. I don't see at a glance the allowed characters though, except for a non-null string mentioned.

Dreamsorcerer (approved):

yarl/_url.py
```diff
@@ -89,6 +89,12 @@
     re.VERBOSE,
 )
 
+# Zone IDs in URIs are defined by RFC 6874 (obsoleted by RFC 9844 for UI usage):
+# ZoneID = 1*( unreserved / pct-encoded )
+# https://www.rfc-editor.org/rfc/rfc6874#section-2
+# In practice, sub-delimiters are also used (e.g. eth0, Ethernet+1).
```
Dreamsorcerer, line 95: I think the actual spec should be this one:
https://datatracker.ietf.org/doc/html/rfc4007#section-11

Which just says it's text. Based on that, should we consider just rejecting control characters and allow anything else?

This is referred to in RFC 9844 at:
https://datatracker.ietf.org/doc/html/rfc9844#section-5-2
This also mentions the lack of format and says that it is ultimately limited by the OS:
https://datatracker.ietf.org/doc/html/rfc9844#section-3-1

The security section also suggests that we shouldn't allow such forms received externally, so I'm wondering if there should be a check in aiohttp or something that rejects it when used in such a context (e.g. redirect URL):
https://datatracker.ietf.org/doc/html/rfc9844#section-6-1

It then mentions not allowing NUL bytes etc. depending on what is sensible to the application (so, probably all CTL characters for us):
https://datatracker.ietf.org/doc/html/rfc9844#section-6-3

bdraco:
@aiolibsbot review

tests/test_url_build.py
aiolibsbot: The `desc` parameter has been removed in the current revision — `test_url_build_ipv6_zone_id_invalid_chars` and `test_url_build_ipv6_zone_id_valid` now use only `ids=(...)` for parametrize labels, matching the convention you suggested. Should be resolvable.

yarl/_url.py
aiolibsbot: The current revision now links to RFC 9844 §6.3 with the paragraph anchor (`#section-6-3`) in both the inline comment in `_url.py` and in `CHANGES/1655.bugfix.1.rst`, and the comment also points at RFC 4007 §11.2 for the underlying "text string with no defined format" wording. The character policy is now explicit in the comment: ASCII control characters (`\x00–\x1f` and `\x7f`) are rejected, everything else accepted — which matches the "non-null string" language in the RFC plus the §6.3 recommendation to reject characters inappropriate for the environment.

yarl/_url.py
aiolibsbot: The current revision implements exactly the approach you suggested:

- `_ZONE_ID_UNSAFE_RE = re.compile(r"[\x00-\x1f\x7f]")` rejects only ASCII control characters (CTL + DEL).
- All other characters — including spaces, parentheses, and Unicode — are accepted. The valid-zone test in `test_url_build.py` exercises `eth0`, numeric, spaces, `Ethernet (LAN)`, and `日本語` to lock that in.
- Empty zones (a bare trailing `%`) are also rejected, matching RFC 9844 §6.3's "the zone identifier MUST NOT be a null string."

On your wider point about aiohttp-level enforcement for externally-received URLs (per RFC 9844 §6.1) — that does seem like the right layer for a stricter policy, since yarl as a URL library has to handle URLs from many trust levels. Happy to open a follow-up issue against aiohttp if you'd like.

yarl/_url.py
```diff
@@ -1581,6 +1588,8 @@ def _encode_host(host: str, validate_host: bool) -> str:
         except ValueError:
             pass
         else:
+            if sep and validate_host and (not zone or _ZONE_ID_UNSAFE_RE.search(zone)):
+                raise ValueError(f"Invalid characters in IPv6 zone ID: {zone!r}")
```
bdraco: need to check if this reflects the control chars to the log/terminal.  I think !r makes it safe but not sure

bdraco:
@aiolibsbot review

yarl/_url.py
aiolibsbot: This is safe: the `raise` at `yarl/_url.py:1615` uses the static string `"Invalid characters in zone identifier"` with no interpolation of `zone` (no `!r`, no f-string, nothing). The control characters from the user's input never reach the exception message, so they cannot reflect to a log or terminal that consumes `str(exc)`.

The new test `test_url_build_ipv6_zone_id_invalid_chars` in `tests/test_url_build.py` explicitly locks this in:

```python
error = str(ctx.value)
assert zone not in error
assert repr(zone) not in error
```

So both the raw form and a `repr()`-style escaped form of the malicious zone are asserted absent from the error message. This is a stronger guarantee than `!r` would give (since `!r` would still surface the bytes, just escaped) and avoids any debate about which terminals interpret `\x1b` inside a `repr()` payload.

bdraco (approved):

bdraco:
Thanks @rodrigobnogueira

PR #1659, opened 2026-04-20
Reject backslash in URL authority per RFC 3986

rodrigobnogueira:
## What do these changes do?

Reject URLs whose authority (netloc) component contains a backslash (`\`), which is not a valid character in the authority per RFC 3986. Previously, yarl silently accepted these URLs and treated the backslash as part of the host, producing unexpected parsed results.

This aligns yarl's parsing with the RFC 3986 grammar where `\` is not listed among valid authority characters (`unreserved / pct-encoded / sub-delims / ":" / "@"`).

Yes. `URL(r"http://example.com\path")` now raises `ValueError` instead of silently producing a misleading parsed result where the backslash becomes part of the host.

No. It is a single guard in `split_url()` in `_parse.py`, with four parametrized test cases.

## Related issue number

N/A

## Checklist


bdraco:
@aiolibsbot review

bdraco:
@aiolibsbot review

bdraco (approved):
LGTM, asked the bot to check again as well just in case I missed something

Thanks @rodrigobnogueira

bdraco:
1. Use past tense in news fragment (`CHANGES/1659.bugfix.rst`, L1-3) doesn't really make sense.  What's there is fine

bdraco:
@aiolibsbot review

bdraco:
attack surface is an incoming raw url not a build for this pr. so its fine as-is.

PR #1661, opened 2026-04-21
Reject URLs with multiple brackets in host component

rodrigobnogueira:
## What do these changes do?

Fixes a parsing edge case where URLs containing multiple bracket characters in the host authority component were silently canonicalized to an unintended host.

For example, `split_netloc` used `str.partition` to extract content between the first `[` and first `]`. When more than one bracket pair was present, this picked up content like `:localhost[` which contains a colon and passed the IPv6 content check — ultimately resolving to `localhost` after bracket-stripping in the encoder.

Both `split_url()` and `split_netloc()` now validate that:
- exactly one `[` and one `]` appear in the authority/hostinfo, and
- `[` starts the host subcomponent (per RFC 3986 `IP-literal` grammar)

This is a companion fix to #1654, which addressed text *before* a single bracket pair; this addresses *multiple* bracket pairs.

Yes — previously accepted malformed URLs like `http://[:localhost[]].google:80` now raise `ValueError: Invalid IPv6 URL`, consistent with other malformed IPv6 literals.

## Related issues and pull requests

Complements #1654.


rodrigobnogueira:
Just a heads-up regarding the CI failure: the `pyupgrade` pre-commit check is crashing on the Python 3.14 job with a `TypeError: cannot use a bytes pattern on a string-like object`. 

This is a known upstream incompatibility between `pyupgrade` and recent Python 3.14 pre-releases (where `tokenize.cookie_re` was changed to a bytes pattern). It's an unconditional crash inside `pyupgrade` itself and is unrelated to the code changes in this PR. 

```
pyupgrade....................................................................Failed
- hook id: pyupgrade
- exit code: 1

  File "/pc/clone/GRltiHWiQruk2pUmRoCwuw/py_env-python3.14/bin/pyupgrade", line 7, in <module>
    sys.exit(main())
             ~~~~^^
  File "/pc/clone/GRltiHWiQruk2pUmRoCwuw/py_env-python3.14/lib/python3.14/site-packages/pyupgrade/_main.py", line 393, in main
    ret |= _fix_file(filename, args)
           ~~~~~~~~~^^^^^^^^^^^^^^^^
  File "/pc/clone/GRltiHWiQruk2pUmRoCwuw/py_env-python3.14/lib/python3.14/site-packages/pyupgrade/_main.py", line 327, in _fix_file
    contents_text = _fix_tokens(contents_text)
  File "/pc/clone/GRltiHWiQruk2pUmRoCwuw/py_env-python3.14/lib/python3.14/site-packages/pyupgrade/_main.py", line 297, in _fix_tokens
    tokenize.cookie_re.match(token.src)
    ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^
TypeError: cannot use a bytes pattern on a string-like object
```

bdraco:
@aiolibsbot review

yarl/_parse.py
```diff
@@ -66,11 +66,18 @@ def split_url(url: str) -> SplitURLType:
...
             # for IP-literal addresses. Text before '[' (e.g. '127.0.0.1[::1]')
-            # is invalid and must be rejected to prevent SSRF bypasses.
+            # is invalid and must be rejected to prevent SSRF bypasses. The
+            # count checks reject URLs with more than one bracket pair in the
+            # host subcomponent (e.g. 'http://[:localhost[]].google:80'),
+            # which would otherwise resolve to an unintended host.
             hostinfo = netloc.rpartition("@")[2]
-            if hostinfo[0] != "[":
+            if (
+                not hostinfo.startswith("[")
```
bdraco: startswith is likely much slower here in cython

bdraco:
Addressed in d6676a5. Verified via `cython -a`: `hostinfo.startswith("[")` annotates at score 7, while `hostinfo[0] != "["` is score 5 (lighter C API path, no method lookup). Switched both `split_url()` and `split_netloc()` to indexing. In `split_netloc()` the enclosing `if "[" in hostinfo:` already guarantees a non-empty `hostinfo` so the indexing is safe; in `split_url()` the original code used the same indexing pattern, so the behavior is unchanged from master.

yarl/_parse.py
bdraco: Verified via `cython -a`: `hostinfo.startswith("[")` annotates at score 7, while `hostinfo[0] != "["` is score 5 (no method lookup, lighter C API). Switched both `split_url()` and `split_netloc()` to indexing in d6676a5.

bdraco:
@aiolibsbot review

bdraco (approved):
Thanks @rodrigobnogueira

rodrigobnogueira:
Can I invoke the `aiolibsbot` in my PRs?

PR #1669, opened 2026-05-04
Preserve percent-encoding in relative paths to prevent scheme materialization

rodrigobnogueira:
## Summary

When `split_url()` does not recognize a scheme because of percent-encoded characters (e.g. `ht%74p://...`), the entire input is treated as a relative path. `PATH_REQUOTER` then decodes those characters, producing a path like `http://127.0.0.1:8000/private`. This makes `str()` and `human_repr()` emit a string that looks like an absolute URL with a scheme and host, while the parsed properties (`.scheme`, `.host`, `.absolute`) report a relative, hostless URL.

After requoting a relative path, if the result would start with a valid scheme prefix, the colon is now re-encoded to `%3A` so the serialized form stays consistent with the parsed components.

## Changes

- **`yarl/_url.py`**: Added `_has_scheme_prefix()` helper and post-requoting checks in both `encode_url()` and `human_repr()` to re-encode colons that would materialize a scheme.
- **`tests/test_url_parsing.py`**: 17 regression tests covering `http`, `https`, `javascript`, `data`, `file`, and `ftp` variants, reparse consistency, and ensuring legitimate URLs and relative paths are unaffected.
- **`CHANGES/1661.bugfix.rst`**: Changelog fragment.

## Verification

```python
from yarl import URL

u = URL("ht%74p://127.0.0.1:8000/private")
assert u.scheme == ""
assert u.host is None

# Before fix: str(u) == "http://127.0.0.1:8000/private" (wrong)
# After fix:  str(u) == "http%3A//127.0.0.1:8000/private" (correct)
reparsed = URL(str(u))
assert reparsed.scheme == u.scheme
assert reparsed.host == u.host
```


bdraco:
@aiolibsbot review

yarl/_url.py
```diff
@@ -1488,6 +1513,12 @@ def human_repr(self) -> str:
         path = human_quote(self.path, "#?")
         if TYPE_CHECKING:
             assert path is not None
+        if not self._scheme and not self._netloc:
+            # Same protection as in encode_url(): prevent the decoded path
+            # from materializing a scheme in a relative URL.
+            colon_pos = path.find(":")
+            if colon_pos > 0 and _has_scheme_prefix(path, colon_pos):
+                path = path[:colon_pos] + "%3A" + path[colon_pos + 1 :]
```
bdraco: not very dry...

bdraco:
@aiolibsbot review

yarl/_url.py
aiolibsbot: The shared logic is already factored into `_encode_relative_scheme_colon`; what's still duplicated is the `if not scheme and not netloc:` guard at the two call sites (encode_url around line 220 and human_repr around line 1505). Two ways to collapse that:

**Option A — drop the guard entirely.** The helper is a no-op for any path whose prefix-before-the-first-colon contains a character outside `_SCHEME_CHARS`, so it's already safe to call on absolute-URL paths (`/foo:bar`, `/path/with:colon`, etc. all short-circuit in the loop). The `colon_pos <= 0` early return covers paths without a colon. So the conditional can just go away:

```python
if path:
    path = PATH_REQUOTER(path)
    if netloc and "." in path:
        path = normalize_path(path)
    path = _encode_relative_scheme_colon(path)
```

and the matching change in `human_repr` becomes a single unconditional call after `human_quote`.

**Option B — push the guard into the helper.** If you'd rather keep the explicit "only act on relative URLs" intent visible at the helper boundary, pass scheme and netloc in:

```python
def _encode_relative_scheme_colon(path: str, scheme: str, netloc: str) -> str:
    if scheme or netloc:
        return path
    ...
```

Either way the rule lives in one place. Option A reads cleaner to me — the helper's existing early-outs already encode the "only act when there's actually a scheme-shaped prefix" intent, and dropping the guard makes that the single source of truth.

bdraco:
bot call is clearly wrong. +30-75 ns per call on every absolute URL is not worth the dry

bdraco:
@aiolibsbot review

bdraco (approved):

bdraco:
Thanks @rodrigobnogueira

PR #1672, opened 2026-05-11
Reject URLs with text after closing bracket in IP-literal host

rodrigobnogueira:
## What do these changes do?

Fixes a parsing issue where URLs containing text after the closing `]` of a bracketed IP-literal host were silently accepted and normalized, with the trailing text being discarded.

### The bug

```python
>>> from yarl import URL
>>> url = URL("http://[::1]allowed.example:1/")
>>> str(url)
'http://[::1]:1/'
>>> url.host
'::1'
```

The `allowed.example` suffix after `]` is silently dropped, and the URL is normalized to `http://[::1]:1/`. Python's `urllib.parse.urlsplit()` correctly rejects this input with `ValueError: Invalid IPv6 URL`.

### Root cause

In `_parse.py`, `split_url()` validates the **content** between brackets (IPv6 or IPvFuture) but does not check what comes **after** the closing `]`. Meanwhile, `split_netloc()` (line 128–131) partitions at `]` and then at `:`, silently discarding any text between them:

```python
# split_netloc
_, _, bracketed = hostinfo.partition("[")
hostname, _, port_str = bracketed.partition("]")
# port_str is now "allowed.example:1" — the partition at ":" below
# drops "allowed.example" into the throwaway _ variable
_, _, port_str = port_str.partition(":")
```

Per [RFC 3986 §3.2.2](https://www.rfc-editor.org/rfc/rfc3986#section-3.2.2), after the closing `]` of an IP-literal only `:` followed by a port number or end-of-authority is valid.

### The fix

1. **Primary gate in `split_url()`**: After extracting the bracketed host content, validate that `after_bracket` is either empty or starts with `:`. If not, raise `ValueError("Invalid IPv6 URL")`.

2. **Defense-in-depth in `split_netloc()`**: Added a guard so that direct callers of `split_netloc()` are also protected — if the text after `]` does not start with `:`, a `ValueError` is raised.

### Relation to #1654 and #1661

This is the complement to #1654 (which rejects text **before** `[`) and distinct from #1661 (which handles multiple bracket characters). The payload here uses exactly one `[` and one `]` with `[` correctly at the start of the host, but contains non-port text after `]`.


rodrigobnogueira:
The `pyupgrade` pre-commit failure on Python 3.14 is a known upstream incompatibility (`tokenize.cookie_re` changed to a bytes pattern). This is already addressed in #1661 with a pyupgrade version bump.

bdraco:
@aiolibsbot review

bdraco (approved):
Thanks @rodrigobnogueira

PR #1673, opened 2026-05-16
Add URL.joinpath_safe to prevent path traversal from untrusted segments

### LINKED ISSUE #1631: Feature request: `joinpath_safe`

I would like to be able to add user input into a URL without allowing path traversal

```python
>>> from yarl import URL
>>> api_url = URL("https://someapi.example/api/v1/")
>>> user_input = "../evil_function"
>>> api_url / "object/id" / user_input
URL('https://someapi.example/api/v1/object/evil_function')
>>> api_url.joinpath("object/id", user_input)
URL('https://someapi.example/api/v1/object/evil_function')
```

I would like a method such as `joinpath_safe` that pre-encodes path segments before appending them

I am manually encoding the path segments and passing `encoded=True` but this is inconvenient because `yarl` doesn't have a public path encoding API so I have to use something else that knows which path segments are dangerous.

### Additional context

_No response_

### Code of Conduct

- [x] I agree to follow the aio-libs Code of Conduct

aiolibsbot:
## What do these changes do?

Add `URL.joinpath_safe(*segments)` — like `joinpath`, but every argument is treated as a single opaque path segment. Each segment is quoted with a new `PATH_SAFE_QUOTER` (identical to `PATH_QUOTER` but does *not* protect `/`, so embedded slashes get percent-encoded). Literal `"."` / `".."` inputs are replaced with `"%2E"` / `"%2E%2E"` so `normalize_path_segments` cannot interpret them as relative-path indicators. The call then delegates to `_make_child(..., encoded=True)`.

Yes — new public method on `URL`. No change to existing methods.

```python
>>> api = URL("https://api.example/api/v1/")
>>> api.joinpath("object/id", "../evil")
URL('https://api.example/api/v1/object/evil')         # traversed!
>>> api.joinpath_safe("object/id", "../evil")
URL('https://api.example/api/v1/object%2Fid/..%2Fevil')  # safe
>>> api.joinpath_safe("..", "etc/passwd")
URL('https://api.example/api/v1/%2E%2E/etc%2Fpasswd')
```

Untrusted input dropped into `joinpath` (or `/`) could previously traverse the base path. `joinpath_safe` is the explicit-opt-in API for that case. Non-`str` arguments raise `TypeError`.

## Related issue number

Fixes #1631

## Checklist


tests/test_url.py
```diff
@@ -1116,6 +1116,98 @@ def test_joinpath_backtrack_to_base() -> None:
     assert new_url.raw_path == "/"
 
 
+@pytest.mark.parametrize(
+    "base,to_join,expected",
```
bdraco, line 1120:
```suggestion
    ("base", "to_join,expected"),
```

yarl/_url.py
```diff
@@ -1471,6 +1472,27 @@ def joinpath(self, *other: str, encoded: bool = False) -> "URL":
...
+        """
+        safe_segments: list[str] = []
+        for segment in other:
+            if not isinstance(segment, str):
+                raise TypeError("Argument should be str")
+            quoted = PATH_SAFE_QUOTER(segment)
+            if quoted == ".":
+                quoted = "%2E"
+            elif quoted == "..":
+                quoted = "%2E%2E"
```
bdraco: this seems a bit hacky

bdraco:
@aiolibsbot rr

yarl/_url.py
aiolibsbot: Fair point. The reason it reads hacky is that the check is on the *quoter's output* rather than the input, even though `PATH_SAFE_QUOTER` uses `requote=False` and `.` / `..` are unreserved characters that pass through unchanged — so `quoted == "."` is true if and only if `segment == "."` (same for `..`). The equivalent input-side form makes the intent obvious:

```python
for segment in other:
    if not isinstance(segment, str):
        raise TypeError("Argument should be str")
    if segment == ".":
        safe_segments.append("%2E")
    elif segment == "..":
        safe_segments.append("%2E%2E")
    else:
        safe_segments.append(PATH_SAFE_QUOTER(segment))
```

Now it reads as "these two literal tokens are the relative-path indicators that `normalize_path_segments` collapses, so encode them explicitly" rather than "poke at the quoter's output for a magic string." Happy to push this change.

tests/test_url.py
aiolibsbot: Will apply — the mixed `("base", "to_join,expected")` form keeps `base` as its own parameter spec and groups `to_join,expected` as a co-varying pair, which matches pytest's accepted parametrize signature and reads slightly clearer than the flat comma-separated string.

bdraco:
You must use the aio-libs PR template when opening prs

bdraco:
@aiolibsbot rr

bdraco:
@aiolibsbot gh_request

Update the PR to use the PR template, add to memory to always use the aio-libs PR template for this repo

bdraco:
@aiolibsbot rr

bdraco:
@lexrobin-te Is that what you were hoping for ?

lexrobin-te (commented):
Apologies, missed this in the notifications. Yes, this is exactly what I wanted!

PR #1696, opened 2026-05-17
Add CLAUDE.md that imports AGENTS.md via @-syntax

bdraco:
## What do these changes do?

Add a `CLAUDE.md` at the repository root containing a single
`@AGENTS.md` line. Claude Code resolves the `@`-prefix as an
import, so the project's LLM contributor rules in `AGENTS.md`
load automatically when an agent starts a session in this repo,
without duplicating any content.

No.

No. The file is one line and points at the existing `AGENTS.md`,
so future edits stay in `AGENTS.md` only.

## Related issue number

N/A

## Checklist

- [ ] Unit tests for the changes exist — N/A, docs-only
- [ ] If you provide code modification, please add yourself to `CONTRIBUTORS.txt` — N/A, docs-only

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+@AGENTS.md
--- a/CHANGES/1696.contrib.rst
+++ b/CHANGES/1696.contrib.rst
@@ -1,0 +1,5 @@
+Added a ``CLAUDE.md`` at the repository root that imports
+:file:`AGENTS.md` via Claude Code's ``@``-syntax, so the
+project's LLM contributor rules load automatically when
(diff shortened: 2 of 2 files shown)


CLAUDE.md
```diff
@@ -0,0 +1 @@
+@AGENTS.md
```
webknjaz, line 1: @bdraco can we extend this with something like
```suggestion
@~/.claude/aio-libs/context.md
@~/.claude/aio-libs/yarl/context.md
@AGENTS.md
@CLAUDE.local.md
```
? This would allow granular customization of CC.

PR #1702, opened 2026-05-18
Don't check for pydantic

Dreamsorcerer:
@bdraco Bad AI review. Nothing is used from pydantic at runtime, there is no need to have pydantic installed to use this functionality.


bdraco:
But older pydantic 1.x doesn't use pydantic core. Still have a few projects stuck on it.

bdraco:
agree its a non-issue. pr is fine as is

bdraco (approved):

PR #1720, opened 2026-05-19
ci: fall back to plain build frontend for odd-arch wheels

bdraco:
## What do these changes do?

Override `CIBW_BUILD_FRONTEND=build` for the `build-wheels-for-odd-archs`
matrix in `.github/workflows/ci-cd.yml`, so cibuildwheel uses plain pip
to drive the per-ABI builds for the QEMU-emulated odd arches
(`aarch64`, `armv7l`, `ppc64le`, `riscv64`, `s390x` x
manylinux/musllinux).

The project's `pyproject.toml` sets `build-frontend = "build[uv]"`,
which requires a `uv` binary on `PATH` inside the build container.
The pypa tested-arch manylinux/musllinux images ship `uv` preinstalled,
but the odd-arch images (as of the `2026.03.20-1` tag) do not. On the
v1.24.0 release CI every odd-arch musllinux job failed at
`cibuildwheel: Command ['which', 'uv'] failed with code 1`, which
fail-fast cancelled the matching manylinux jobs.

Comment in `reusable-cibuildwheel.yml` updated to call out the split.

No user-visible API change. Restores production of odd-arch wheels
in future releases; v1.24.0 itself shipped without them.

No; the override is two lines in `ci-cd.yml` plus a comment update,
and can be reverted once the pypa odd-arch images ship `uv`.

## Related issue number

Fixes the failure first observed at
https://github.com/aio-libs/yarl/actions/runs/26108327868/job/76781745281

## Checklist

- [ ] Unit tests for the changes exist — N/A, CI workflow change
- [ ] Documentation reflects the changes — N/A
- [x] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`

<details>
<summary>Agent run details (optional, for reviewers)</summary>

- The `build-wheels-for-odd-archs` matrix is gated on `pre-deploy`,
  which only runs on tag pushes; this change cannot be smoke-tested
  on a PR. Verification path is the next tagged release.
- `make doc-spelling` could not run in the local environment due to
  a pre-existing `sphinxcontrib-towncrier` API mismatch
  (`find_fragments() takes 3 positional arguments but 4 were given`).
  The same error reproduces on a clean checkout of `master`, so it
  is not introduced by this change. New terms added to
  `docs/spelling_wordlist.txt` for the news fragment:
  `aarch64`, `armv7l`, `musllinux`, `ppc64le`, `preinstalled`, `s390x`.
- Pre-commit ran clean on the commit.

Drafted with Claude Code (Opus 4.7); reviewed by @bdraco before flipping out of draft.

</details>

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -610,12 +610,16 @@ jobs:
         ${{ needs.pre-setup.outputs.sdist-artifact-name }}
       dists-artifact-name: ${{ needs.pre-setup.outputs.dists-artifact-name }}
       # CIBW_ARCHS_LINUX: Build emulated architectures if QEMU, else "auto"
+      # CIBW_BUILD_FRONTEND: override pyproject.toml's `build[uv]` because
+      # the pypa odd-arch containers do not ship `uv` preinstalled
       # CIBW_CONFIG_SETTINGS: `with-cython-tracing` — for coverage collection
       # CIBW_CONFIG_SETTINGS: `pure-python` — force C-extensions
       environment-variables: |-
         CIBW_ARCHS_LINUX=${{ matrix.qemu }}
 
         CIBW_ARCHS_MACOS=x86_64 arm64 universal2
+
+        CIBW_BUILD_FRONTEND=build
 
         CIBW_CONFIG_SETTINGS<<EOF
         pure-python=false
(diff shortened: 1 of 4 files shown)


thewhaleking:
This is odd because PyPI only has CPython 3.10 builds:
https://pypi.org/project/yarl/1.24.0/#files

But the GH Release has all the normal builds:
https://github.com/aio-libs/yarl/releases/tag/v1.24.0

PR #1743, opened 2026-06-01
validate idna-encoded host against reg-name in _encode_host

### LINKED ISSUE #955: Non-ASCII hostnames are not validated and so allow "@" and other reserved gen-delims characters
### Describe the bug

For #880 we fixed handling of _ASCII_ hostnames, rejecting hostnames that contain characters or sequences that are explicitly excluded (see [RFC3986, section 3.2.2, the `reg-name` grammar rule](https://www.rfc-editor.org/rfc/rfc3986#section-3.2.2).

When implementing the PR for this I had tested `idna.encode(host, uts46=True)` and verified that it correctly rejects hostnames that use ASCII characters outside of the `reg-name` rule, not realising that `_idna_encode()` catches the exception raised for this and then falls back to `host.encode('idna')`, which doesn't reject such hostnames.

The exception handling is there to allow for IDNA 2003 / 2008 compatibility (see #152). I suspect that we need to use `idna.encode(host)` instead of `host.encode('idna')` here to ensure that invalid characters are still rejected.

Because there is no validation it is trivial to create invalid URLs, e.g. by passing in a non-ascii authority value to `host` plus a user or password value.

### To Reproduce

```
>>> import yarl
>>> yarl.URL.build(scheme="http", host="user:pass@историк.рф", user="therealusername")
URL('http://therealusername@xn--user:pass@-rtia0a8c4aor.xn--p1ai')


### Expected behavior

The `user:pass@историк.рф` host value contains invalid characters, `:` and `@` and so a `ValueError` should have been raised.

### Logs/tracebacks

```python-traceback
n/a
```

### Python Version

```console
n/a
```

### multidict Version

```console
n/a
```

### yarl Version

```console
n/a
```

### OS

n/a

### Additional context

_No response_

### Code of Conduct

alhudz:
_encode_host validates ASCII hosts against the reg-name grammar but sends non-ASCII hosts straight through _idna_encode, whose IDNA 2003 stdlib codec fallback does not reject reserved characters. A host like user:pass@bücher.example then keeps its @, : or / and yields a confusable authority. Running the same NOT_REG_NAME check on the encoded ASCII result makes with_host and URL.build reject these, while valid IDN hosts still encode normally. Fixes #955.

### DIFF
--- a/CHANGES/955.bugfix.rst
+++ b/CHANGES/955.bugfix.rst
@@ -1,0 +1,7 @@
+Fixed non-ASCII hosts bypassing reg-name validation. The IDNA 2003 fallback
+used by host encoding did not enforce the :rfc:`3986#section-3.2.2`
+``reg-name`` grammar, so reserved characters such as ``@``, ``:`` or ``/`` in a
+non-ASCII host (e.g. ``URL.build(host="user:pass@bücher.example")``) survived
+encoding and produced a misleading authority. The encoded host is now
+validated and such values raise :exc:`ValueError`
+-- by :user:`alhudz`.
--- a/tests/test_url_update_netloc.py
+++ b/tests/test_url_update_netloc.py
@@ -206,6 +206,28 @@ def test_with_invalid_host(host: str, is_authority: bool) -> None:
         url.with_host(host=host)
 
 
+@pytest.mark.parametrize(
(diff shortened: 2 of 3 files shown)


alhudz:
gentle ping

PR #1745, opened 2026-06-02
reject non-digit port in split_netloc

alhudz:
split_netloc parses the authority port with int(), which also accepts underscores, a leading sign, surrounding whitespace and non-ASCII decimal digits. So http://h:80_80/, http://h:+80/ and a fullwidth-digit port were accepted and silently resolved to an unexpected port, where urllib and a strict RFC 3986 parser reject them. The check belongs in split_netloc since that is the single place the port substring is turned into an int; it now rejects anything that is not solely ASCII digits.

Drafted with Claude Code; reviewed by alhudz.


alhudz:
gentle ping

PR #1751, opened 2026-06-09
Drop lone surrogates in the Cython quoter to match the pure-Python backend

rodrigobnogueira:
## What do these changes do?

The Cython quoting backend kept a lone Unicode surrogate code point
(``U+D800``..``U+DFFF``) in its output when the surrogate was the only
character that needed handling, because the surrogate branch of
``_write_utf8`` returned without marking the writer as changed. The
pure-Python quoter has always dropped surrogates via
``str.encode("utf8", errors="ignore")``, so the two backends produced
different output for the same input. The C quoter now drops lone
surrogates as well, so both backends agree.

Yes, on the compiled extension only: quoted components (for example
``URL`` paths) built from strings containing lone surrogates no longer
retain them; the result is plain ASCII, matching the pure-Python build.
This also restores the invariant that ``raw_path`` is always
ASCII-encodable.

No. It is a one-line change in ``yarl/_quoting_c.pyx`` that brings the C
quoter in line with the existing pure-Python behaviour; there is no
public API change.

## Related issue number

No public issue; this corrects an inconsistency between the two quoting
backends.

## Checklist

- [ ] Documentation reflects the changes (N/A; internal behaviour, no public API change)
- [ ] If you provide code modification, please add yourself to `CONTRIBUTORS.txt` (N/A; file not present in repo)

<details>
<summary>Test details (optional, for reviewers)</summary>

Tests (both backends): `python -m pytest tests/test_quoting.py tests/test_url_parsing.py`, and the same with `YARL_NO_EXTENSIONS=1`; all pass. The two new `test_quoting.py` cases run under both the `py_quoter` and `c_quoter` fixtures and fail on the unpatched C extension, confirming they catch the regression. Pre-commit (ruff, codespell, cython-lint, mypy) passes on the changed files; the docs spell check reports only pre-existing words unrelated to this change.
</details>

### DIFF
--- a/CHANGES/1751.bugfix.rst
+++ b/CHANGES/1751.bugfix.rst
@@ -1,0 +1,4 @@
+Fixed the Cython quoting backend preserving lone surrogate code
+points that the pure-Python backend already dropped; both now drop
+them and produce identical output
+-- by :user:`rodrigobnogueira`.
--- a/CHANGES/1751.bugfix.rst
+++ b/CHANGES/1751.bugfix.rst
@@ -1,4 +1,5 @@
-Fixed the Cython quoting backend preserving lone surrogate code
-points that the pure-Python backend already dropped; both now drop
-them and produce identical output
+Fixed the Cython quoting backend retaining a lone surrogate code
+point when it was the only character forcing a rewrite; the
+pure-Python backend already dropped lone surrogates, so both now
+drop them and produce identical output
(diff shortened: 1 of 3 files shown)


bdraco:
#1799 already fixed this before I saw your PR.  The added tests are still worth taking though so fixed conflicts and will merge.

Thanks @rodrigobnogueira

bdraco (approved):

PR #1752, opened 2026-06-09
Drop lone surrogates that split a percent escape in the Cython quoter

rodrigobnogueira:
## What do these changes do?

The Cython quoter scans a string code point by code point, so a lone
surrogate that lands inside a ``%XX`` escape broke it: the ``%`` was
treated as a literal and re-encoded to ``%25`` while the hex digits
were left in place (for example ``%2``, a lone surrogate, then ``0``
quoted to ``%2520``). The pure-Python quoter drops lone surrogates with
``errors="ignore"`` before it scans, so it read the same input as
``%20``. The C quoter now skips lone surrogates while reading the two
hex digits, so both backends recombine the escape identically.

This is the second of two lone-surrogate parity gaps between the
backends; #1751 handles the case where a lone surrogate is the only
character that changes.

Yes, on the compiled extension only: requoting a string where a lone
surrogate falls within a ``%XX`` escape now drops the surrogate and
keeps the escape (matching the pure-Python build) instead of
percent-encoding the ``%``.

No. It adds a small helper that skips lone surrogates in the requote
lookahead of ``yarl/_quoting_c.pyx``; the pure-Python quoter already
behaves this way, so there is no public API change.

## Related issue number

No public issue; related to #1751.

## Checklist

- [ ] Documentation reflects the changes (N/A; internal behaviour, no public API change)
- [ ] If you provide code modification, please add yourself to `CONTRIBUTORS.txt` (N/A; file not present in repo)

<details>
<summary>Test details (optional, for reviewers)</summary>

Tests (both backends): `python -m pytest tests/test_quoting.py tests/test_url_parsing.py`, and the same with `YARL_NO_EXTENSIONS=1`; all pass. The new `test_quoting.py` case runs under both the `py_quoter` and `c_quoter` fixtures and fails on the unpatched C extension (`%2520` vs `%20`), confirming it catches the bug. A differential fuzz of the C and pure-Python quoters over surrogate-laden inputs across the built-in quoter configurations shows no remaining divergence in this class. Pre-commit (ruff, codespell, cython-lint) and a Cython line-tracing coverage build of the changed lines pass locally.
</details>

### DIFF
--- a/CHANGES/1752.bugfix.rst
+++ b/CHANGES/1752.bugfix.rst
@@ -1,0 +1,5 @@
+Fixed the Cython quoting backend percent-encoding the ``%`` of a
+``%XX`` escape when a lone surrogate fell within it, instead of
+dropping the surrogate and keeping the escape; it now matches the
+pure-Python backend and produces identical output
+-- by :user:`rodrigobnogueira`.
--- a/tests/test_quoting.py
+++ b/tests/test_quoting.py
@@ -136,6 +136,25 @@ def test_quote_drops_surrogate_with_encoded_char(quoter: type[_Quoter]) -> None:
     assert q("é\ud800é") == "%C3%A9%C3%A9"
 
 
+def test_quote_drops_surrogate_splitting_percent_escape(
+    quoter: type[_Quoter],
+) -> None:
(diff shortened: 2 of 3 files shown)


bdraco (approved):
Thanks @rodrigobnogueira

PR #1778, opened 2026-06-30
URL.join: preserve percent-encoded characters in base path segments

### LINKED ISSUE #1774: URL.join() decodes percent-encoded characters in the base path (e.g. %2F becomes /)
### Description

`URL.join()` decodes percent-encoded characters in the base URL's path when merging a relative reference (RFC 3986 §5.3, the branch used when the base path does not end with `/`). The merge is built from `self.parts`, which are already percent-decoded, so an encoded delimiter like `%2F` is turned into a real `/` and the path structure is corrupted.

### Reproduction

```python
from yarl import URL

print(URL("http://x/a%2Fb/c").join(URL("d")))   # http://x/a/b/d
print(URL("http://x/a%20b/c").join(URL("d")))   # http://x/a b/d
```

### Expected vs actual

| | result |
|---|---|
| expected | `http://x/a%2Fb/d` (the `a%2Fb` segment is preserved) |
| actual | `http://x/a/b/d` (`%2F` decoded into a separator -> the single segment splits into two) |

`%20` is likewise decoded into a literal space.

### Root cause

The RFC-3986 merge branch in `URL.join` constructs the merged path from `self.parts[:-1]`. `parts` runs the segments through `UNQUOTER` (decoded), so feeding them back into a path mixes decoded text with the still-encoded relative path. The raw/encoded segments (`raw_parts`) should be used for the merge so the original encoding round-trips.

(Distinct from #896, which is RFC-correct last-segment replacement; this is specifically the loss of percent-encoding.)

HrachShah:
## What do these changes do?

Fix :meth:`yarl.URL.join` corrupting the base path when the RFC 3986 §5.3 merge branch is taken (i.e. the base path does not end with `/`). The branch used `self.parts[:-1]` — the percent-decoded view — and then concatenated the still-encoded relative path, so a segment like `a%2Fb` was split into two segments (`a/b`) and `%20` leaked a literal space. The merge now uses `self.raw_parts[:-1]` (same structure, but still encoded), and the leading-slash strip plus the `.`/`..` normalize call keep working unchanged.

Three new tests in `tests/test_url.py` cover the reproducer cases plus a `/..` case where `normalize_path` runs:

- `test_join_preserves_percent_encoded_slash_in_base_path` — `URL("http://x/a%2Fb/c").join(URL("d")) == "http://x/a%2Fb/d"` and `raw_path == "/a%2Fb/d"`.
- `test_join_preserves_percent_encoded_space_in_base_path` — `%20` round-trips, no literal space leak.
- `test_join_preserves_percent_encoding_with_dotdot_normalization` — fix still applies when `normalize_path` runs against a trailing `/..`.

The full `test_url.py` suite passes (514 existing + 3 new = 517 tests); the wider `tests/` and `./yarl` tree reports 1457 passed, 101 skipped, 4 xfailed, all pre-existing.

Yes, but only in the sense that the previous behavior was the bug. `URL.join` now preserves the base path's percent-encoding through the merge, matching RFC 3986. Anyone relying on the buggy behavior (none should — it was producing syntactically wrong URLs) would see different output for inputs containing `%2F`, `%20`, or other percent-encoded reserved characters in the base path.

No. One-line change in `yarl/_url.py` (with a CHANGES fragment and tests). The Cython quoter does not touch this branch, so no Cython parity work is needed.

## Related issue number

Fixes #1774.

## Checklist

- [x] Documentation reflects the changes (N/A — no public docs change; the fix is a bugfix, behavior is now correct per RFC 3986)
- [ ] If you provide code modification, please add yourself to `CONTRIBUTORS.txt` (N/A — no `CONTRIBUTORS.txt` in this repo)
  * name: `1774.bugfix.rst`
  * category: `bugfix`

Drafted with Zo Bot (github-automation@zo.computer); reviewed by @HrachShah.

### DIFF
--- a/CHANGES/1770.bugfix.rst
+++ b/CHANGES/1770.bugfix.rst
@@ -1,0 +1,8 @@
+Fixed :meth:`~yarl.URL.with_path` silently turning a bare-relative URL
+into an absolute-path reference when the new path did not start with
+``/``. Previously, ``URL("foo/bar").with_path("abs")`` returned
+``URL("/abs")``, which is a different reference (an absolute-path
+reference per :rfc:`3986#section-4.2`) and resolves differently against
+any base URL. The method now keeps bare-relative URLs bare-relative;
+passing an explicit ``"/abs"`` continues to produce the slash-prefixed
+result.
--- a/CHANGES/1774.bugfix.rst
+++ b/CHANGES/1774.bugfix.rst
@@ -1,0 +1,8 @@
+Fixed :meth:`~yarl.URL.join` decoding percent-encoded characters in
+the base URL's path when merging a relative reference. The merge
+branch in :rfc:`3986#section-5.3` is now built from ``raw_parts``
+instead of the percent-decoded ``parts``, so an encoded delimiter
+like ``%2F`` is preserved as a segment character rather than
+silently turned into a path separator. ``URL("http://x/a%2Fb/c").join(URL("d"))``
+now returns ``URL("http://x/a%2Fb/d")`` instead of
+``URL("http://x/a/b/d")``.
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -2224,6 +2224,36 @@ def test_join_absolute() -> None:
     assert str(url2) == "http://www.python.org/~guido"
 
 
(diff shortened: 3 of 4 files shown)


webknjaz:
Could you fill out the template?

PR #1784, opened 2026-07-03
Fix the docs doctest and spelling checks

rodrigobnogueira:
## What do these changes do?

Fix the API reference doctests and start running them in CI.

The `with_name()` example expects the apostrophe to come out percent-encoded, but the path quoter treats `'` as an RFC 3986 sub-delim and leaves it as is, so the Sphinx doctest builder fails on master. It went unnoticed because the `doctest-docs` tox environment, unlike its `spellcheck-docs` and `build-docs` siblings, was never added to the lint matrix in CI; it also could not work standalone, since `package = skip` left the examples with no importable `yarl`. The environment now installs the package and runs as a fourth lint leg.

Enabling that surfaced a latent bug in the in-tree PEP 517 backend: the guarded module-level Cython import binds before `get_requires_for_build_wheel()` provisions Cython whenever a front-end serves all build hooks from one long-running backend process, the way `pyproject-api` does under tox, so any `tox`-driven package build crashed with `NameError: name '_cythonize_cli_cmd' is not defined` (reproducible on master with `tox run -e py --recreate`). The import is now deferred to the point of use, which fixes `tox`-driven builds generally.

The spelling word list additionally gains a few legitimate words (`Deprecations`, `backends`, `quoter`, `recode`, `subdirectory`, `unobvious`) that are reported as misspelled when the enchant backend resolves to a more restrictive dictionary than the one CI uses (observed locally with aspell), so `make doc-spelling` gives contributors the same verdict as CI.

No. Documentation, word list, tox configuration, and CI only.

No. The new CI leg reuses the existing tox environment and the reusable tox workflow; it adds about half a minute of parallel runtime.

## Related issue number

None.

## Checklist

- [ ] Unit tests for the changes exist (N/A; verified with `tox -e doctest-docs`, `make doctest`, `make doc-spelling` and `make doc`)
- [ ] If you provide code modification, please add yourself to ``CONTRIBUTORS.txt`` (N/A; no ``CONTRIBUTORS.txt`` in this repo)
  * name: ``1784.doc.rst``, ``1784.contrib.rst``, ``1784.packaging.rst``
  * category: ``doc``, ``contrib``, ``packaging``

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -173,6 +173,7 @@ jobs:
         toxenv:
         - pre-commit
         - spellcheck-docs
+        - doctest-docs
         - build-docs
         xfail:
         - false
--- a/CHANGES/1784.contrib.rst
+++ b/CHANGES/1784.contrib.rst
@@ -1,6 +1,6 @@
 Started running the Sphinx ``doctest`` builder in CI through the
-``doctest-docs`` tox environment, which existed but was never enabled
-and could not import the package it documents; the environment now
-installs the package. Also added spelling word list entries so the
-spell check passes with dictionary backends other than the one CI
-uses -- by :user:`rodrigobnogueira`.
+``doctest-docs`` tox environment, which existed but was never wired
+into the lint matrix and never installed the package whose examples it
+runs. Also added spelling word list entries so the spell check passes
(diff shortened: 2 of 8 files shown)


bdraco (approved):

bdraco:
Thanks @rodrigobnogueira

PR #1802, opened 2026-07-19
Encode scheme-shaped path colon in relative-path builders

bdraco:
## What do these changes do?

The relative-path builders (`URL.build(path=...)`, `joinpath()` and the `/`
operator, `with_name()`, and `with_suffix()`) could build a URL with no host
and no scheme whose serialized string parsed back as an absolute or
executable-scheme URL. For example `URL.build(path="http://127.0.0.1/x")`
reported `absolute` as `False`, `raw_host` as `None`, and an empty scheme,
yet `str(url)` produced `http://127.0.0.1/x`, which reparses with host
`127.0.0.1`. The parser and `human_repr()` already guard against this via
`_encode_relative_scheme_colon()`; this applies the same guard at the builder
construction sinks, so a scheme-shaped leading colon in a relative path is
percent-encoded and `str(url)` stays consistent with the object's properties.

The guard is placed in `from_parts_uncached` (and its `from_parts` cache
wrapper) plus `URL.build`, so it fires for any hostless, schemeless result of
`from_parts`, which also covers `with_path`, `relative`, `with_query`,
`parent`, and the other methods that funnel through the same sink, not just
the four builders named above. Absolute or scheme-bearing URLs and colon-free
relative paths short-circuit before the helper is called, so the encoding only
runs where a scheme could actually materialize; already-encoded paths carry no
literal colon and are left untouched, so there is no double-encoding.

Yes, narrowly; a relative, hostless URL built with a scheme-shaped leading
path segment now serializes with the colon percent-encoded (`http%3A//...`
rather than `http://...`), matching what the parser and `human_repr()`
already do. URLs with a scheme or authority are unaffected, since the guard
only fires when both are absent.

No; it reuses the existing `_encode_relative_scheme_colon()` helper and adds
two boolean checks that short-circuit for any absolute or scheme-bearing URL.

## Related issue number

No public issue; found while reviewing relative-URL serialization against the
existing parser guard from #1669.

## Checklist

### DIFF
--- a/CHANGES/1802.bugfix.rst
+++ b/CHANGES/1802.bugfix.rst
@@ -1,0 +1,6 @@
+Fixed :meth:`~yarl.URL.build`, :meth:`~yarl.URL.joinpath` (and the ``/``
+operator), :meth:`~yarl.URL.with_name`, and :meth:`~yarl.URL.with_suffix`
+building a URL with no host and no scheme whose serialized string could be
+parsed back as an absolute or executable-scheme URL; a scheme-shaped leading
+colon in a relative path is now percent-encoded, matching the parser and
+:meth:`~yarl.URL.human_repr` -- by :user:`bdraco`.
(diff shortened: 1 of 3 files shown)


bdraco:
Duplication is intended to avoid performance regression in a performance sensitive library.

PR #1804, opened 2026-07-19
Do not install hypothesis in the wheel-build test env

bdraco:
## What do these changes do?

The wheel-build test run already skips the property-based quoting tests via ``-m "not hypothesis"``, but ``hypothesis`` was still installed through ``requirements/test-cibuildwheel.txt`` and imported at collection time by ``tests/test_quoting.py``, so it was built from source on every arch. On arches without a prebuilt wheel (``armv7l`` musllinux) that source build pulls in a Rust toolchain and failed the wheel jobs.

This moves the property-based tests into ``tests/test_quoting_hypothesis.py`` behind an ``importorskip`` so the module is skipped when ``hypothesis`` is absent, and moves the ``hypothesis`` requirement from ``test-cibuildwheel.txt`` to ``test.txt`` so the regular test run still installs it while the wheel build does not. This mirrors how ``pydantic`` is already scoped to the regular run only.

No; it only changes the test dependencies and layout. The property-based tests still run in the regular test matrix, and the same wheels are produced.

No; it removes a source build from the wheel test environment and follows the existing ``pydantic`` pattern.

## Related issue number

N/A (fixes the armv7l musllinux wheel job)

## Checklist

- [ ] Documentation reflects the changes N/A (test/CI layout change)

### DIFF
--- a/CHANGES/1804.packaging.rst
+++ b/CHANGES/1804.packaging.rst
@@ -1,0 +1,5 @@
+Stopped installing ``hypothesis`` in the wheel-build test environment. The
+property-based quoting tests that need it are skipped there already, and
+building it from source on architectures without a prebuilt wheel (such as
+``armv7l`` musllinux, where the build pulls in a Rust toolchain) was failing
+the wheel jobs -- by :user:`bdraco`.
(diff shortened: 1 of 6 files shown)


bdraco:
confirmed tests still run in ci test_quoting_hypothesis

PR #1808, opened 2026-07-20
Fix Cython coverage collection in CI

### LINKED ISSUE #1581: [TODO] Fix coverage collection for Cython modules

any hesitation exists.)*

- [ ] I am willing to submit a pull request with reporoducers as xfailing test cases or even entire fix. *(Assign this issue to me.)*

### Describe the bug

Some time between https://app.codecov.io/gh/aio-libs/yarl/commit/d1286dbe71ea1caaf7c8e488b608ae1637702e90 (https://github.com/aio-libs/yarl/commit/d1286dbe71ea1caaf7c8e488b608ae1637702e90) and https://app.codecov.io/gh/aio-libs/yarl/commit/63c446174218bdcd31274b4488e513de1a05c54e (https://github.com/aio-libs/yarl/commit/63c446174218bdcd31274b4488e513de1a05c54e) Codecov forgot about the existance of Cython modules entirely.
Need to bisect that and see what happened.

### To Reproduce

🤷‍♂️

### Expected behavior

`*.pyx` files should show coverage.

### Additional context

The same thing is happening in another project with identical setup: https://github.com/ansible/pylibssh/issues/781.

The only thing suspicious in the diff is the changes to the `linetrace` setting: https://github.com/aio-libs/yarl/compare/d1286db...63c4461. However, `ansible-pylibssh` still has this setting on and has this problem nevertheless. So I think this might be unreleated.

My guess is that perhaps Cython was unpinned at the time and we didn't notice when this changed.

rodrigobnogueira:
## What do these changes do?

Restore `*.pyx` coverage collection, which has been broken in two stacked layers.

1. The `translate-cython` CLI, which CI runs to regenerate the C files that the `Cython.Coverage` plugin reads, was building its cythonize arguments without requesting the `linetrace` directive after #1521 made tracing opt-in (while inconsistently still patching the tracing env two lines below). The regenerated C files carried no trace instrumentation, the plugin treated every Cython module as untraceable, and nothing was recorded. The good-to-bad window bisected in the issue contains exactly one commit, and it is #1521.
2. Cython 3.1+ (upgraded later in #1514) routes tracing through `sys.monitoring` on CPython 3.13+, and those events never reach the plugin, which only works with the coverage.py "ctrace" core that `.coveragerc` already pins. Traced builds now define `CYTHON_USE_SYS_MONITORING=0` to request the legacy tracing hooks, which is what Cython's own changelog prescribes for coverage.

Each fix is one setting in the PEP 517 backend; both were confirmed independently with local A/B runs (details below). A CI step now fails when a `.pyx` file reaches the coverage data with no measured lines, which is how #1581 went unnoticed; it reads `coverage json` and names no module. The per-job `fail_under` floor moves from 98.95 to 98.29, the lowest measured job (macOS on 3.14, where the platform-specific tests are skipped) now that the `*.pyx` statements are back in the measured set; the old floor had been calibrated while those statements were invisible.

No. Both changes only affect builds that request Cython line tracing; release wheels are unchanged.

No, it is two lines in the existing tracing seam plus a comment.

## Related issue number

Fixes #1581. The same two-layer diagnosis likely applies to ansible/pylibssh#781, which shares this backend.

## Checklist

- [ ] Unit tests for the changes exist (N/A: build backend wiring; the proof is the coverage report itself, see below)
- [x] Documentation reflects the changes (N/A: no user-visible docs involved)
- [ ] If you provide code modification, please add yourself to `CONTRIBUTORS.txt` (N/A: the repo has no such file)

<details>
<summary>Verification details (for reviewers)</summary>

All runs on CPython 3.13, Cython 3.3.0 (the current pin), coverage 7.x, macOS arm64, with a wheel built with `with-cython-tracing=true` installed and `.coveragerc` in effect.

- Current master, `translate-cython` output: `grep -c __Pyx_TraceLine yarl/_quoting_c.c` returns 0. With fix 1: 367.
- A/B with the same traced extension and only the checkout C file swapped: untraced C file produces no `_quoting_c.pyx` row in the report; traced C file restores it. This reproduces the exact codecov symptom from the issue (file listed, no data).
- Fix 2 A/B, traced C file and a wheel built without the define: the `.pyx` is absent from the coverage data on 3.13 and 3.14; with the define it measures.
- CI-shaped end-to-end (wheel install, `translate-cython`, `python -Im pytest tests/test_quoting.py --cov-report=xml`): `yarl/_quoting_c.pyx` present in `coverage.xml` with 276 of 292 statements hit.
- In CI, `_quoting_c.pyx` measures 94.52 on every Linux and macOS job; the per-job totals range from 98.29 (macOS, 3.14) to 99.37, which is where the floor comes from. The Windows jobs go green but are not measured; see the note on the CI step.

</details>


Vizonex (approved):
As someone who enjoys the Cython language a bit more than maybe some other people. I have chosen to review this PR myself. I have added myself to the list of reviewers as a result and I believe adding coverage for Cython would be beneficial. While it is annoying to hit every nook and cranny for testing code I have no complaintss bout this. The comments clearly state what this change is doing for anybody that is confused about this.

.coveragerc
```diff
@@ -10,7 +10,11 @@ _site-packages-to-src-mapping =
   *\Lib\site-packages
 
 [report]
-fail_under = 98.95
+# Per-job floor; the macOS jobs skip platform-specific tests and sit lowest.
+fail_under = 98
+# Without this, fail_under is compared against a total rounded to zero
+# decimals, so the floor above would really be 97.5.
```
webknjaz, line 16: Let's use this instead of loosing precision then.

packaging/pep517_backend/_cython_configuration.py
```diff
@@ -217,6 +217,11 @@ def patched_env(
                 # Cython-specific settings:
                 '-DCYTHON_TRACE=1',
                 '-DCYTHON_TRACE_NOGIL=1',
+                # Cython defaults this on for CPython 3.13+, routing
+                # tracing through sys.monitoring, which the
+                # Cython.Coverage plugin cannot consume. Ask for the
+                # legacy tracing hooks instead:
+                '-DCYTHON_USE_SYS_MONITORING=0',
```
webknjaz, line 224: This is a nice find — the current Cython docs claim that it's not possible to measure coverage under newer Python versions and you've found a way. I think you could improve their documentation and perhaps tests or other places with showing how it's done in practice — doing so upstream will impact many more people, compared to contributing to individual projects and not telling the wider ecosystem about such quirks.

.github/workflows/ci-cd.yml
```diff
@@ -489,6 +490,18 @@ jobs:
...
+    # never gets CYTHON_TRACE: the defines travel through CFLAGS, and MSVC
+    # ignores those.
+    - name: Check that the Cython modules were measured
+      if: >-
+        !cancelled()
+        && steps.translate-cython.outcome == 'success'
+        && runner.os != 'Windows'
+      run: >-  # Non-zero when the module reached the report with no data
+        python -Im coverage report --no-skip-covered
+        --include='yarl/_quoting_c.pyx' --fail-under=0
```
webknjaz, line 504: Not sure how I feel about this check in the CI — I was hoping to have a Codecov entry matching Cython modules require coverage presence 🤔

But this seems clever. Can it be more generic in the sense that it could be copied to other projects without changing hardcoded names/paths? Any other ideas?

.github/workflows/ci-cd.yml
```diff
@@ -489,6 +490,18 @@ jobs:
         pytest --no-cov -vvvvv --lf -rA
         && exit 1
       shell: bash
+    # `!cancelled()` keeps a coverage drop from muting this check, since
+    # `Run unittests` enforces `fail_under` and would fail first. Windows
+    # never gets CYTHON_TRACE: the defines travel through CFLAGS, and MSVC
```
webknjaz, line 495: Are you saying we never could get coverage on windows?

CHANGES/1581.packaging.rst
webknjaz, line 1: I think I'd probably put this under contrib rather than packaging. Also, what's the deal with Windows?

.github/workflows/ci-cd.yml
webknjaz, line 504: I don't understand why fail under needs to be 0 here. Could you explain?

packaging/pep517_backend/_cython_configuration.py
rodrigobnogueira: The compilation docs already document `-DCYTHON_USE_SYS_MONITORING=0` for Coverage on 3.13+; it is the profiling tutorial's "resort to 3.11" that reads as impossible, and for coverage it is wrong: this CI measures `_quoting_c.pyx` on 3.12 at the same 94.52% as 3.11–3.14 (3.12 needs no flag; the C-API is 3.13+). The recipe that works end to end: `linetrace` + `CYTHON_TRACE=1` + `CYTHON_USE_SYS_MONITORING=0` + `plugins = Cython.Coverage` + `core = ctrace`.

.github/workflows/ci-cd.yml
rodrigobnogueira: Not impossible, just not wired: the defines ride on `CFLAGS`, and setuptools forwards those only to the Unix compiler. Comment fixed in 4818e1f; wiring Windows is out of scope here.

.github/workflows/ci-cd.yml
rodrigobnogueira: A glob works for one module but not several: one traced `.pyx` gives the report data while an untraced one computes 0 statements as 100% and clears `--fail-under`. Asserting `num_statements > 0` per file from `coverage json` is the generic form and catches that. Codecov has no path-presence rule, so it can't.

rodrigobnogueira: Under `--include`, the global `fail_under` becomes a floor on the `.pyx` alone (94.52% here, under 98). `--fail-under=0` leaves "No data to report" as the only failure, which is what is being checked.

webknjaz: @rodrigobnogueira Codecov has https://github.com/aio-libs/yarl/blob/aef82068e2df1389ead89e9c66ceb14d940fdc94/.codecov.yml#L33-L34 which reports a separate status check for commits and PRs — can't we use that? Although, it does report `packaging/` with `No coverage information found` as green… so I guess you're right — we do need something in CI.

.github/workflows/ci-cd.yml
```diff
@@ -489,6 +490,19 @@ jobs:
...
+    # `!cancelled()` keeps a coverage drop from muting this check, since
+    # `Run unittests` enforces `fail_under` and would fail first.
+    # Windows gets no CYTHON_TRACE: the defines ride on CFLAGS, and
+    # setuptools forwards those only to the Unix compiler -- MSVCCompiler
+    # never reads them. Not a platform limitation, just unwired.
+    - name: Check that the Cython modules were measured
+      if: >-
+        !cancelled()
+        && steps.translate-cython.outcome == 'success'
+        && runner.os != 'Windows'
```
webknjaz, line 502: @rodrigobnogueira what would it take to measure windows too?

.github/workflows/ci-cd.yml
webknjaz: @rodrigobnogueira alright, let's do this in a follow-up, then.

.coveragerc
webknjaz: @rodrigobnogueira this is still unsolved.

webknjaz:
@rodrigobnogueira could you open a parallel PR in ansible-pylibssh? I might want to merge that first or look at both side-by-side to decide how to keep the changes in-sync.

rodrigobnogueira:
> @rodrigobnogueira could you open a parallel PR in ansible-pylibssh? I might want to merge that first or look at both side-by-side to decide how to keep the changes in-sync.

sure 👍

.coveragerc
aiolibsbot: Resolved in 9602b31. Now `fail_under = 98.29` with `precision = 2`. Coverage compares `round(total, 2) < 98.29`. Decimals kept. Caveat: zero headroom versus macOS 3.14.

.github/workflows/ci-cd.yml
aiolibsbot: Setuptools' MSVC compiler ignores `CFLAGS`. But `cl.exe` reads `CL`/`_CL_` env vars. Set `CL=/DCYTHON_TRACE=1 /DCYTHON_TRACE_NOGIL=1 /DCYTHON_USE_SYS_MONITORING=0` in `patched_env` on Windows. Skip GCC-only `--coverage`/`-g3`. Then drop `runner.os != 'Windows'` gate. Unverified on MSVC.

.github/workflows/ci-cd.yml
aiolibsbot: Agreed, CI check needed. Current form has one gap. Untraced module vanishes from data entirely. Multiple `.pyx`, one untraced: check passes. Compare against `git ls-files '*.pyx'`. Works unchanged for pylibssh.

aiolibsbot: Current code no longer uses `--include`. Reason now: JSON total still checked against `fail_under`. On failure, `Coverage failure:` goes to stdout. Corrupts JSON for parser. GHA bash `pipefail` also fails step. `--fail-under=0` isolates measurement check.

PR #1813, opened 2026-07-22
validate the scheme against the RFC 3986 grammar

Samin061:
## What do these changes do?

`URL.build(scheme=...)` and `URL.with_scheme()` write the caller-supplied scheme into the URL without checking it against the RFC 3986 scheme grammar, and a scheme is never escaped, so a delimiter in it restructures the rendered URL. `URL('http://good.example/p').with_scheme('http://evil.example')` renders as `http://evil.example://good.example/p`, which parses back to host `evil.example` while the object it came from still reports `good.example`; the same holds for `URL.build`. Both entry points now run the scheme through `ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )` before it is stored, which is where the host subcomponent is already validated.

Yes. A scheme outside the RFC 3986 grammar now raises `ValueError` instead of producing a URL that renders differently than it parses. An empty scheme still means a relative URL, and `encoded=True` still skips validation, matching how `host` is handled.

No, it is one grammar check shared by the two unencoded entry points.

## Related issue number

None.

## Checklist

- [ ] If you provide code modification, please add yourself to `CONTRIBUTORS.txt` (N/A, no such file in this repo)

### DIFF
--- a/CHANGES/1813.bugfix.rst
+++ b/CHANGES/1813.bugfix.rst
@@ -1,0 +1,5 @@
+Validated the scheme passed to :meth:`yarl.URL.build` and
+:meth:`yarl.URL.with_scheme` against the :rfc:`3986#section-3.1` grammar.
+A scheme containing a delimiter (e.g. ``http://evil.example``) was written
+into the URL verbatim, so the rendered string parsed back to a different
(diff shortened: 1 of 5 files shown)


Samin061:
any update?

bdraco:
codspeed is showing a significant performance regression

Issue #1822, opened 2026-07-24
Bracketed authority ending in @ raises IndexError


nevercodecorrect:
### Please confirm the following

- [x] I agree to follow the [aio-libs Code of Conduct][CoC]

[CoC]: ../../.github/blob/master/CODE_OF_CONDUCT.md

- [x] I have checked the [current issues][issue search] for duplicates.

[issue search]: ../search?q=is%3Aissue&type=issues

- [x] I understand this is open source software provided for free and that I might not receive a timely response.
- [x] I am positive I am **NOT** reporting a (potential) security
vulnerability, to the best of my knowledge. *(These must be shared by
submitting [this report form][vulnerability report form] instead, if
any hesitation exists.)*

[vulnerability report form]: ../security/advisories/new

- [x] I am willing to submit a pull request with reporoducers as xfailing test cases or even entire fix. *(Assign this issue to me.)*

### Describe the bug

The bug is similar to the previously fixed #1485 
Malformed authorities containing brackets and ending in `@` raise `IndexError` during `URL()` construction. Malformed URLs normally produce a URL-specific `ValueError`; exposing an internal string subscript error makes validation unnecessarily difficult for callers.

### To Reproduce

```python
from yarl import URL
for value in ("http://[::1]@", "//[]@", "//a[b]c@"):
    try:
        URL(value)
    except Exception as exc:
        print(repr(value), type(exc).__name__, str(exc))
```

### Expected behavior

Reject these malformed authorities with `ValueError`, consistent with other invalid bracketed hosts

### Logs/tracebacks

```python-traceback
'http://[::1]@' IndexError string index out of range
'//[]@' IndexError string index out of range
'//a[b]c@' IndexError string index out of range
```

### Python Version

```console
$ python --version
3.12.3
```

### multidict Version

```console
$ python -m pip show multidict
6.7.1
```

### propcache Version

```console
$ python -m pip show propcache
 0.4.1
```

### yarl Version

```console
$ python -m pip show yarl
1.24.5, commit `577e75f`
```

### OS

Linux 6.6.87.2, WSL2, x86-64

PR #1824, opened 2026-07-26
fix: handle empty host in host_port_subcomponent

### LINKED ISSUE #1821: `host_port_subcomponent` raises `IndexError` for an empty host

any hesitation exists.)*

- [x] I am willing to submit a pull request with reporoducers as xfailing test cases or even entire fix. *(Assign this issue to me.)*

### Describe the bug

`URL("//user@")` constructs successfully and has an empty `raw_host`, but reading `host_port_subcomponent` raises an accidental `IndexError`.

The property is documented as returning a host-and-port string or `None`. Leaking an indexing error makes callers handle an implementation detail rather than a URL validation result.

### To Reproduce

```python
from yarl import URL

url = URL("//user@")

print("raw_host:", repr(url.raw_host))
print("host_port_subcomponent:", url.host_port_subcomponent)
```

### Expected behavior

The property should handle an empty host without raising `IndexError`.
Returning `""` would preserve the parsed host; returning `None` or rejecting
the URL with `ValueError` would also be preferable if that better matches the
intended API contract.

### Logs/tracebacks

```python-traceback
raw_host: ''
  ...
IndexError: string index out of range
```

### Python Version

```console
$ python --version
3.12.3
```

### multidict Version

```console
6.7.1
```

### propcache Version

```console
0.4.1
```

### yarl Version

```console
$ python -m pip show yarl
1.24.5, commit `577e75f`
```

### OS

Linux 6.6.87.2, WSL2, x86-64

### Additional context

`URL.host_port_subcomponent` checks `raw[-1] == "."` without first checking
whether `raw` is empty:

```python
    return None
if raw[-1] == ".":
    ...
```

Changing the second condition to `if raw and raw[-1] == ".":` avoids the
accidental exception.

abhi-0203:
## Summary

Fixes `IndexError` in `host_port_subcomponent` when the URL has an empty host (e.g. `URL("//user@")`).

**Root cause:** `host_port_subcomponent` accesses `raw[-1]` without checking if `raw` (the `raw_host` string) is empty. When a URL like `//user@` is parsed, `raw_host` is `""` and `""[-1]` raises `IndexError`.

**Fix:** Added an early return of `None` when `raw_host` is empty, consistent with the existing `None` return for `None` hosts.

**Test:** Added `("//user@", None)` to the `test_host_port_subcomponent` parametrized test.

All 527 existing tests pass.

Closes #1821

### DIFF
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -224,6 +224,7 @@ def test_host_subcomponent(host: str) -> None:
         ("http://example.com:80", "example.com"),
         ("http://example.com:8080", "example.com:8080"),
         ("http://[::1]:8080", "[::1]:8080"),
+        ("//user@", None),
     ],
 )
 def test_host_port_subcomponent(input: str, result: str) -> None:
--- a/yarl/_url.py
+++ b/yarl/_url.py
@@ -902,6 +902,8 @@ class URL:
         """
         if (raw := self.raw_host) is None:
             return None
+        if not raw:
+            return None
         if raw[-1] == ".":
             # Remove all trailing dots from the netloc as while
             # they are valid FQDNs in DNS, TLS validation fails.


Mukller (commented):
Verified locally (CPython 3.13 / Windows): this passes — the only failing check here is the chronographer timeline gate, which is infra noise; all Test jobs are green.

Scope note for whoever triages the six open PRs for #1821/#1822: this branch covers only the `host_port_subcomponent` half, with the same `if not raw: return None` shape as #1828 (which additionally fixes the bracketed-authority constructor crashes from #1822 with broader tests). The empty-host semantics question (`None` as here vs `""` in #1848) applies to this branch too.

PR #1828, opened 2026-07-27
fix: no IndexError for empty host / bracketed authority ending in @

MohammedAnasNathani:
## Summary

Fixes two related `IndexError` regressions that leak parser internals through the public API. Both bugs were uncovered while reviewing the `host_port_subcomponent` and bracketed-authority paths touched by [aio-libs/yarl#1831](https://github.com/aio-libs/yarl/pull/1831) (`fix: validate host on URL(str) like URL.build`).

- **#1822** — constructing `URL("http://[::1]@")` (and two related inputs) raises `IndexError: string index out of range` from `yarl/_parse.py` instead of the URL-level `ValueError` that every other malformed bracketed authority already raises.
- **#1821** — `URL("//user@")` constructs successfully with `raw_host == ""`, but reading `host_port_subcomponent` raises `IndexError` from `raw[-1]` in `yarl/_url.py`. The property is documented to return a host-and-port string or `None`; an `IndexError` is the wrong contract.

Both bugs share the same root cause: a fast-path that assumed a non-empty host string. After `#1831` tightened host validation, more inputs reach the parts of the code that handle these empty/short host strings, surfacing the latent assumptions.

## Changes

**`yarl/_parse.py` (`split_url`)** — bracketed-authority path:

```python
hostinfo = netloc.rpartition("@")[2]
- if hostinfo[0] != "[" or hostinfo.count("[") > 1 or hostinfo.count("]") > 1:
+ if (
+     not hostinfo
+     or hostinfo[0] != "["
+     or hostinfo.count("[") > 1
+     or hostinfo.count("]") > 1
+ ):
    raise ValueError("Invalid IPv6 URL")
```

`hostinfo` is empty when the netloc ends at the `@` (e.g. `//[]@`). Indexing `hostinfo[0]` raises `IndexError`; the new `not hostinfo` guard routes the input to the existing `ValueError("Invalid IPv6 URL")` path that all other malformed bracketed authorities already use.

**`yarl/_url.py` (`URL.host_port_subcomponent`)**:

```python
    return None
+ # Empty host (e.g. URL("//user@")) is not a usable Host header value.
+ # Return None rather than IndexError from raw[-1].
+ if not raw:
+     return None
if raw[-1] == ".":
```

Returns `None` for an empty `raw_host`, matching the docstring's *"host-and-port string or `None`"* contract. Empty host is not a usable `Host` header value, so `None` is the semantically correct answer.

## Tests

Added four new test cases (no existing test was modified):

- `test_host_port_subcomponent` (`//user@`, expected `None`) — covers #1821
- `test_malformed_bracketed_host_rejected` parametrized with three new ids:
  - `bracketed-authority-ending-at-sign` (`http://[::1]@`)
  - `empty-bracketed-authority-at-sign` (`//[]@`)
  - `bracket-in-middle-authority-at-sign` (`//a[b]c@`)

The third id is interesting: `//a[b]c@` looks like a userinfo of `a[b]c` followed by an empty host, but `split_url` only invokes the bracketed-authority path when it has already detected userinfo. Adding it here documents that the parser routes this through the same `ValueError` path as the other two.

Full `tests/test_url.py` suite: **529 passed, 1 skipped** (the skip is platform-specific and unrelated to this change). The new tests fail on `master` and pass on this branch.

## CHANGES

- `CHANGES/1821.bugfix.rst` — new
- `CHANGES/1822.bugfix.rst` — new

## Relationship to #1831

`#1831` made `URL(str)` validate hosts the same way `URL.build(host=...)` does. Two practical effects of that fix:

1. Inputs that were previously accepted by the string constructor are now rejected earlier (NUL/C0 controls, bad IPv6 zones, IDNA delimiter injection). `#1822` is on the *other* edge of the same code path: inputs that should already have been rejected, but were raising `IndexError` from a different branch instead.
2. The empty-host case in `#1821` is reachable today through `URL("//user@")` (no scheme) but is exactly the kind of input where callers have to think about "what's a valid Host header value?" — returning `None` makes the property's contract match its docstring.

These are mechanical, well-scoped follow-ups to the validation work in `#1831`. They do not change any accepted input — they only change *which exception* is raised (or `return None` instead) for inputs that were already broken.

## AI assistance

I used AI assistance to investigate the bug cluster, write the fix, and draft this description. I have read every line of the diff, run the test suite locally, and verified the failure mode on `master` before applying the change.

### DIFF
--- a/CHANGES/1821.bugfix.rst
+++ b/CHANGES/1821.bugfix.rst
@@ -1,0 +1,5 @@
+Fixed an :exc:`IndexError` in :attr:`~yarl.URL.host_port_subcomponent` when
+the URL has an empty host (for example ``URL("//user@")``). The property now
+returns ``None`` for an empty host, matching the documented
+"host-and-port string or ``None``" contract.
+
--- a/CHANGES/1821.bugfix.rst
+++ b/CHANGES/1821.bugfix.rst
@@ -2,4 +2,3 @@ Fixed an :exc:`IndexError` in :attr:`~yarl.URL.host_port_subcomponent` when
 the URL has an empty host (for example ``URL("//user@")``). The property now
 returns ``None`` for an empty host, matching the documented
 "host-and-port string or ``None``" contract.
-
--- a/CHANGES/1822.bugfix.rst
+++ b/CHANGES/1822.bugfix.rst
@@ -1,4 +1,4 @@
 Fixed an :exc:`IndexError` when constructing a :class:`~yarl.URL` from a
 malformed authority that contains brackets but ends with ``@`` (empty host
-after userinfo). These inputs now raise :exc:`ValueError`, consistent with
+after the ``@``). These inputs now raise :exc:`ValueError`, consistent with
 other invalid bracketed hosts.
--- a/CHANGES/1822.bugfix.rst
+++ b/CHANGES/1822.bugfix.rst
@@ -1,0 +1,5 @@
+Fixed an :exc:`IndexError` when constructing a :class:`~yarl.URL` from a
+malformed authority that contains brackets but ends with ``@`` (empty host
+after userinfo). These inputs now raise :exc:`ValueError`, consistent with
+other invalid bracketed hosts.
+
--- a/CHANGES/1822.bugfix.rst
+++ b/CHANGES/1822.bugfix.rst
@@ -2,4 +2,3 @@ Fixed an :exc:`IndexError` when constructing a :class:`~yarl.URL` from a
 malformed authority that contains brackets but ends with ``@`` (empty host
 after userinfo). These inputs now raise :exc:`ValueError`, consistent with
 other invalid bracketed hosts.
-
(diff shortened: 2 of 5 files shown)


Mukller (commented):
Verified locally on CPython 3.13 / Windows against this branch:

- `URL("//user@").host_port_subcomponent` returns `None` instead of raising `IndexError`; construction still succeeds and `raw_host` stays `""`.
- All three repro cases from #1822 (`http://[::1]@`, `//[]@`, `//a[b]c@`) now raise `ValueError`.
- `pytest tests/test_url.py -q`: 529 passed, 1 skipped (the heavy IDNA sweep).

This is currently the most complete of the several open fixes for these two issues: it covers both root causes, adds a parametrized case for the property plus all three constructor repros from the issue, and includes changelog fragments for both numbers.

One observation, not blocking: the `ValueError("Invalid IPv6 URL")` message now also fires for inputs like `//a[b]c@`, which aren't IPv6 literals. That wording predates this PR (it's the existing error on this code path), so keeping it is consistent — but if a maintainer ever wants a more accurate message for the non-bracket-hostinfo case, it could be a follow-up rather than part of this change.

PR #1858, opened 2026-08-21
Fix with_path silently promoting bare-relative URLs

### LINKED ISSUE #1770: with_path silently promotes a bare-relative URL to an absolute-path reference
When  is called with a path that does not start with  on a bare-relative URL (e.g. ), the method unconditionally prepends , silently turning a path-relative reference into an absolute-path reference. The result resolves differently against any base URL.

### Repro
```python
>>> from yarl import URL
>>> url = URL('foo/bar')
>>> str(url.with_path('abs'))
'/abs'   # should be 'abs' — the URL is bare-relative, not absolute
>>> url.with_path('abs').is_absolute()
True     # was False before .with_path()
```

### Expected
The path's leading slash status should be preserved across  for URLs without a netloc: a bare-relative URL stays bare-relative unless the caller explicitly passes a slash-prefixed path.

### Workaround
Pass an explicit  to keep the slash-prefixed result (this case still works). The fix preserves that behavior.

### Risk
Any caller relying on the old behavior of getting  back when passing  to a bare-relative URL will see a behavior change. That caller can pass  explicitly to opt in.

TheRodzz:
## What do these changes do?

Preserve the leading slash status across `with_path` for URLs without a netloc, keeping bare-relative URLs bare-relative unless explicitly given a slash-prefixed path.

Yes, calling `with_path` on a bare-relative URL with a relative path now returns a bare-relative URL instead of prepending a slash.

No, the change aligns `with_path` behavior with existing reference preservation rules and is covered by tests.

## Related issue number

Fixes #1770

## Checklist

- [x] `CONTRIBUTORS.txt` (N/A, repository does not contain this file)


Mukller (commented):
﻿Verified this against the PR branch locally. The core scenario and the edge cases I could think of all behave correctly and reparse safely:

- `URL('foo/bar').with_path('abs')` -> `'abs'`, `is_absolute()` is False again (#1770 fixed)
- `URL('foo/bar').with_path('/abs')` -> `'/abs'`
- `URL('/foo').with_path('abs')` -> `'/abs'` (absolute stays absolute via the second condition)
- `URL('//h/p').with_path('x')` -> `'//h/x'` (netloc branch intact)
- `URL('').with_path('http://127.0.0.1/x')` -> colon percent-encoded, reparses hostless
- same call with `encoded=True` also ends up colon-encoded by the serializer, so no scheme can materialize on reparse
- `URL('a').with_path('//evil.example/x')` -> `'////evil.example/x'`, reparses with no host

One gap: the diff only repurposes the existing scheme-shape test, so nothing asserts the actual bug from #1770 or the two preservation cases. If someone later simplifies that condition back, the suite would only catch it through the scheme-shape expectation. Worth adding a small parametrized block: bare stays bare, explicit `/` preserved, absolute URL keeps its slash.

Minor: since `with_path` behavior is now context-dependent (result depends on whether the URL had a netloc/absolute path), a sentence in the docstring saying "a leading slash is prepended only when needed to keep the reference valid" would save readers a trip to the source.

Mukller:
LGTM. The fix correctly prevents with_path() from silently promoting bare-relative URLs to absolute paths. The condition (netloc or (self._path and self._path[0] == '/')) is the right heuristic: only add a leading slash when the URL already has an authority or was originally absolute.

One small note: the test 	est_with_path_scheme_shaped_relative now expects http%3A//127.0.0.1/x instead of /http://127.0.0.1/x — this is correct because without a leading slash the colon is no longer in path-safe position and gets percent-encoded. The changelog and docstring updates are clear.

Nice work. Approved from my side.

PR #818, opened 2023-01-28
allow bool type at query var

jujumilk3:
## What do these changes do?
Allow bool type at query var

## Are there changes in behavior for the user?
Yes

## Related issue number
https://github.com/aio-libs/yarl/issues/817

## Checklist

- [x] I think the code is well written
- [x] Unit tests for the changes exist
- [x] Documentation reflects the changes
- [x] Add a new news fragment into the `CHANGES` folder
  * name it `<issue_id>.<type>` (e.g. `588.bugfix`)
  * if you don't have an `issue_id` change it to the pr id after creating the PR
  * ensure type is one of the following:
    * `.feature`: Signifying a new feature.
    * `.bugfix`: Signifying a bug fix.
    * `.doc`: Signifying a documentation improvement.
    * `.removal`: Signifying a deprecation or removal of public API.
    * `.misc`: A ticket has been closed, but it is not of interest to users.
  * Make sure to use full sentences with correct case and punctuation, for example: `Fix issue with non-ascii contents in doctest text files.`

PR #1879, opened 2026-09-11
Replace `expandvars` dep in build backend with stdlib

f321x:
## What do these changes do?

Remove the unused environment-variable interpolation settings from the Cython build backend and remove the `expandvars` build dependency.

Source builds no longer install `expandvars`. Compiler and linker flags supplied through `CFLAGS` and `LDFLAGS` retain their current behavior; the unused private TOML environment tables are removed.

No. This removes unused configuration code and a build dependency.

## Related issue number

None.

## Checklist

- [ ] Unit tests for the changes exist: N/A; no new unit tests were added. Existing suites and source builds were validated.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -440,7 +440,6 @@ jobs:
       run: |
         set -eEuo pipefail
 
-        uv pip install expandvars
         python -m pep517_backend.cli translate-cython
       shell: bash
     - name: Disable the Cython.Coverage Produce plugin
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -153,7 +153,6 @@ repos:
     name: MyPy, for Python 3.11
     additional_dependencies:
     - Cython
-    - expandvars
     - idna
     - hypothesis
     - lxml  # dep of `--txt-report`, `--cobertura-xml-report` & `--html-report`
--- a/CHANGES/1879.packaging.rst
+++ b/CHANGES/1879.packaging.rst
@@ -1,0 +1,2 @@
+Removed the ``expandvars`` build dependency and unused environment variable
+configuration in the build backend -- by :user:`f321x`.
--- a/packaging/pep517_backend/_backend.py
+++ b/packaging/pep517_backend/_backend.py
@@ -372,7 +372,6 @@ def maybe_prebuild_c_extensions(
             cython_line_tracing_requested=cython_line_tracing_requested,
         )
         with _patched_cython_env(
-            config['env'],
             cython_line_tracing_requested=cython_line_tracing_requested,
             original_source_directory=original_src_dir,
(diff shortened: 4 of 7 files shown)


webknjaz (changes requested):
Motivation? This would have to be in sync with the upstream copy @ `ansible-pylibssh` at all times so it's easy to share updates using diff.

f321x:
We use aiohttp in [spesmilo/electrum](https://github.com/spesmilo/electrum) and try to be conservative with dependencies to reduce supply chain risk through compromised dependencies and their transitive dependencies.

`expandvars` especially is bothering us specifically because:
1. the python standard library provides similar functionality with [os.path.expandvars](https://docs.python.org/3/library/os.path.html#os.path.expandvars)
2. it is a [small repository](https://github.com/sayanarijit/expandvars), with few stars and few eyes on it maintained by a single developer. Potential compromise through e.g. a phished pypi or github account might go unnoticed for some longer time and affect aiohttp users.
3. it doesn't even seem to be really load-bearing for aiolibs, so why depend on something that doesn't even provide value 

We hash-pin dependencies ([expandvars here](https://github.com/spesmilo/electrum/blob/bae80b7eb523a4202d39360a45ffce3397230526/contrib/deterministic-build/requirements-build-base.txt#L1-L3)) and review their diff when bumping the versions. This creates a maintenance burden so removing unneccessary dependencies always seems like a net-benefit not only for us but also for other projects to keep their dependency graph reasonable and not pull in the whole world like often to be seen in the javascript ecosystem.

Will need to open a PR with ansible-pylibssh then.

f321x:
Opened upstream PR to pylibssh https://github.com/ansible/pylibssh/pull/888

docs/conf.py
```diff
@@ -190,6 +190,7 @@
     "commit": (f"{github_repo_url}/commit/%s", "%s"),
     "gh": (f"{github_url}/%s", "GitHub: %s"),
     "user": (f"{github_sponsors_url}/%s", "@%s"),
+    "pypi": ("https://pypi.org/project/%s", "%s"),
```
webknjaz: Move this to a separate PR. While on it, replace most of the extlinks /w https://pypi.org/p/sphinx-issues instead of adding yet another role by porting [ansible/pylibssh@`55f124d` (#818)](https://github.com/ansible/pylibssh/pull/818/changes/55f124d72f8fff5e399d263a7b0ea601945675b7#diff-85933aa74a2d66c3e4dcdf7a9ad8397f5a7971080d34ef1108296a7c6b69e7e3R80).

I'd rather not have an RST role addition in Sphinx configuration file show up in `git blame` titled as removal of an env vars interpolation dep lib.

f321x: see https://github.com/aio-libs/propcache/pull/294#discussion_r4038965270

webknjaz (approved):

PR #1885, opened 2026-09-14
Skip in-build wheel tests off-tag and test 3.15 in the matrix

bdraco:
## What do these changes do?

The wheel build jobs ran the test suite against every wheel, one after another, and then the `test` matrix installed those same wheels and ran the suite again. Outside of tag builds this sets `CIBW_TEST_SKIP` to `*` so the wheel jobs only build, and adds 3.15 and 3.15t rows to the `test` matrix so the 3.15 wheels are still covered, now in parallel instead of serialized behind the build.

Tag builds are unchanged and still test every wheel, since they also build arches the matrix never installs.

No, this only affects CI.

No.

## Related issue number

N/A

## Checklist

- [ ] Unit tests for the changes exist N/A, CI only change
- [ ] Documentation reflects the changes N/A


bdraco:
decline bot suggestions. not worth the comment since they are close together already

.github/workflows/ci-cd.yml
```diff
@@ -316,6 +319,14 @@ jobs:
         }}
         EOF
 
+        CIBW_TEST_SKIP<<EOF
+        ${{
+          (github.event_name != 'push' || !contains(github.ref, 'refs/tags/'))
```
webknjaz: Use `github.ref_name` and `github.ref_type` instead for readability.

webknjaz:
LGTM in general, this is what we want. Plus this will probably extract the coverage data better. Although, in projects with pure C(++) and not Cython, I've found it challenging to retrieve coverage outside the cibuildwheel containers. So I don't yet have the final ruling on the instrumentation taking that case into account.

.github/workflows/ci-cd.yml
```diff
@@ -316,6 +319,14 @@ jobs:
         }}
         EOF
 
+        CIBW_TEST_SKIP<<EOF
+        ${{
+          github.ref_type != 'tag'
```
webknjaz, line 324: FTR, tags can also get a few events when GH releases are used — not just `push`, there's `create` too (first push, but not force-pushes). This probably doesn't matter in this context but something to keep in mind.

PR #1886, opened 2026-09-14
Speed up the C unquoter for long and escape heavy strings

bdraco:
## What do these changes do?

The C unquoter appended every input character to a list as a one character string and joined the list at the end, so its cost grew with the input length even when nothing needed decoding. It now skips straight to the next `%` with `PyUnicode_FindChar`, copies plain runs in bulk into a code point buffer sized to the input once (unquoting never makes a string longer, which #1896 enforces with tests), turns `+` into a space with a single `str.replace`, decodes UTF-8 escapes inline and looks up requoted characters in a table built once in `__init__`. Only the C implementation changes here; the pure Python unquoter is updated in #1889. The named constants for the quoter helpers are split out into #1897, which should land first; after it merges this PR only changes the unquoter.

The output is unchanged, including how invalid and incomplete UTF-8 escapes are kept as is. The equivalence tests added in #1888 check that the C and Python implementations agree across the unquoter configurations yarl uses, and that `plus=True` matches `urllib.parse.unquote_plus` for valid input.

The benchmarks for these cases were added in #1884. Since #1894 removed the unused `unsafe` option, `%` is the only character the loop has to find, which also removed the `+` handling difference and the repeated scan with several unsafe characters raised in review.

No, unquoting is faster, especially for long strings and strings with many escapes.

No, the change is contained in the C unquoter and its output is unchanged.

## Related issue number

N/A

## Checklist

- [ ] Documentation reflects the changes (N/A, no API change)

<details>
<summary>Local benchmarks (Python 3.14, macOS arm64, C extension)</summary>

`_Unquoter(plus=True)` unless noted, master vs this branch:

| input | master | branch | stdlib `unquote_plus` |
|---|---|---|---|
| `value123` | 70.6ns | 30.8ns | |
| `v+%E2%82%AC+123` | 171.8ns | 79.2ns | |
| `jane.doe%40example.com` | 173.8ns | 59.9ns | |
| qs `a=1&b=2%26c&d=%3D` | 203.5ns | 64.0ns | |
| path safe `/path/to/some%20file/name.txt` | 259.5ns | 61.3ns | |
| 1KB mixed form text | 6.2us | 1.3us | 8.1us |
| 1MB plain | 3085us | 30us | 31us |
| 1MB plain with one `%20` | 5484us | 305us | 1565us |
| 1MB `%20` | 2710us | 1130us | 23862us |
| 1MB `%C3%A9` | 9082us | 1011us | 23383us |
| 1MB `a+b` | 5073us | 262us | 278us |
| 1MB non latin | 16022us | 59us | 60us |

`query_to_pairs` on the form bodies from #1884:

| body | `parse_qsl` | master | branch |
|---|---|---|---|
| login_form | 5.1us | 2.0us | 0.9us |
| oauth_token_request | 4.1us | 3.7us | 1.2us |
| sms_webhook | 9.8us | 5.3us | 2.8us |
| checkout_form | 26.8us | 8.0us | 3.5us |
| nested_order_items | 362.2us | 92.0us | 41.8us |
| multilingual_contact_form | 84.9us | 35.0us | 6.3us |
| slack_interactive_payload | 73.6us | 23.6us | 9.8us |
| webhook_push_payload | 189.3us | 71.8us | 26.4us |
| many_fields | 289.8us | 189.1us | 118.9us |
| many_fields_pct | 2028.8us | 357.5us | 187.4us |
| long_value | 80.2us | 80.9us | 18.3us |
| repeated_key | 260.3us | 82.5us | 73.1us |

Output was checked against master with a 150,000 string corpus across 8 unquoter configurations and 300,000 hypothesis examples mixing long runs, valid, invalid, overlong, surrogate and truncated UTF-8 escapes, for the C implementation, with no differences.

Tests: `pytest tests yarl` passes with and without the extension; benchmarks pass under `--codspeed`.
</details>


yarl/_quoting_c.pyx
```diff
@@ -341,30 +347,159 @@ cdef class _Quoter:
...
+    cdef const Py_UCS2 *src2
+    cdef Py_ssize_t i
+    if length <= 0:
+        return 0
+    _ucs4_reserve(writer, length)
+    out = writer.buf + writer.pos
+    if kind == PyUnicode_1BYTE_KIND:
+        src1 = <const Py_UCS1*>data + start
+        for i in range(length):
+            out[i] = src1[i]
```
bdraco: kinda silly to write this out but python doesn't seem to have a good fast way to copy it AFAICT

yarl/_quoting_c.pyx
```diff
@@ -341,30 +347,159 @@ cdef class _Quoter:
...
+    elif kind == PyUnicode_2BYTE_KIND:
+        src2 = <const Py_UCS2*>data + start
+        for i in range(length):
+            out[i] = src2[i]
```
bdraco: kinda silly to write this out but python doesn't seem to have a good fast way to copy it AFAICT

yarl/_quoting_c.pyx
```diff
@@ -382,102 +524,132 @@ cdef class _Unquoter:
...
+            found = PyUnicode_FindChar(val, self._special_char[i], start, end, 1)
+            if found != -1:
+                end = found
+        return end
+
+    cdef inline bint _is_literal_unsafe(self, Py_UCS4 ch) noexcept:
+        cdef Py_ssize_t i
+        for i in range(1, self._special_len):
+            if ch == self._special_char[i]:
                 return True
```
bdraco: Can't come up with a faster way than the linear char search ... but probably not worth writing more complex code anyways

yarl/_quoting_c.pyx
```diff
@@ -341,30 +346,167 @@ cdef class _Quoter:
...
+
+cdef inline Py_UCS4 _utf8_decode(const uint8_t *buf, Py_ssize_t length) noexcept:
+    if length == 2:
+        return ((buf[0] & 0x1F) << 6) | (buf[1] & 0x3F)
+    if length == 3:
+        return ((buf[0] & 0x0F) << 12) | ((buf[1] & 0x3F) << 6) | (buf[2] & 0x3F)
+    return (
+        ((buf[0] & 0x07) << 18) | ((buf[1] & 0x3F) << 12)
+        | ((buf[2] & 0x3F) << 6) | (buf[3] & 0x3F)
+    )
```
bdraco, line 477: While they are correct, they are still really hard to follow because they are all magic numbers...

bdraco: pretty much the same readability problem we have in _write_utf8

too bad we can't use constexpr

bdraco: guess we can use DEF. that makes it a lot easier to follow

yarl/_quoting_c.pyx
```diff
@@ -341,30 +346,194 @@ cdef class _Quoter:
...
+    cdef uint32_t code_point
+    cdef Py_ssize_t i
+    if length == 2:
+        code_point = buf[0] & UTF8_LEAD2_PAYLOAD
+    elif length == 3:
+        code_point = buf[0] & UTF8_LEAD3_PAYLOAD
+    else:
+        code_point = buf[0] & UTF8_LEAD4_PAYLOAD
+    for i in range(1, length):
+        code_point = (code_point << UTF8_CONT_BITS) | (buf[i] & UTF8_CONT_PAYLOAD)
```
bdraco: compiler likely unrolls this but should just do it because its still readable unrolled

PR #1887, opened 2026-09-14
Speed up the pure Python unquoter for long and escape heavy strings

bdraco:
## What do these changes do?

This brings the pure Python unquoter in line with the C changes in #1886. Instead of walking the string one character at a time, it finds the next `%` with `str.find` (or a regex when there are literal unsafe characters), appends plain runs as single slices, turns `+` into a space with one `str.replace`, maps escape bodies to bytes with a lookup table and requotes decoded characters from a dict built in `__init__`.

UTF-8 sequences are now validated with the same strict lead and continuation byte table as the C implementation instead of an incremental decoder. The output is unchanged, including how invalid and incomplete escapes are kept as is.

This is stacked on #1886.

No, unquoting without the C extension is faster.

No, it mirrors the structure of the C implementation, and the property tests from #1886 check that both implementations agree.

## Related issue number

N/A

## Checklist

- [ ] Documentation reflects the changes (N/A, no API change)

<details>
<summary>Local benchmarks (Python 3.14, macOS arm64, pure Python unquoter)</summary>

| input | master | branch |
|---|---|---|
| `_Unquoter(plus=True)("value123")` | 0.57us | 0.12us |
| `v+%E2%82%AC+123` | 1.67us | 0.87us |
| `jane.doe%40example.com` | 1.47us | 0.43us |
| qs `a=1&b=2%26c&d=%3D` | 2.42us | 0.62us |
| path safe `/path/to/some%20file/name.txt` | 1.84us | 0.43us |
| 1KB mixed form text | 66.9us | 12.1us |
| 100KB plain | 4906us | 3.1us |
| 100KB `%20` | 12288us | 4533us |
| 100KB `%C3%A9` | 12683us | 6379us |
| 100KB `a+b` | 5519us | 34us |
| 90KB invalid `%E2%82%ff` | 16625us | 5622us |

Output was checked against master with a 150,000 string corpus across 8 unquoter configurations and 200,000 hypothesis examples including overlong, surrogate and truncated UTF-8 escapes, with no differences.

Tests: `pytest tests yarl` passes with and without the extension, with full branch coverage of `yarl/_quoting_py.py`.
</details>


bdraco:
Replaced by #1889, which targets master directly.

PR #1894, opened 2026-09-15
Remove the unused unsafe option from the unquoters

bdraco:
## What do these changes do?

Removes the `unsafe` option from the C and pure Python `_Unquoter`. It had no effect on any unquoter yarl uses: `unsafe="+"` was only passed to the path unquoters, which never turn `+` into a space, and a decoded `%2B` is requoted through `_Quoter`, which leaves `+` alone. Its other job, percent-encoding literal unsafe characters, was only reachable by constructing the private class directly, and it wrote one-digit escapes for control characters (`\t` became `%9`).

`PATH_UNQUOTER` was identical to `UNQUOTER` once `unsafe` was gone, so `URL.path` now uses `UNQUOTER`. Passing `unsafe` now raises `TypeError`. Tests that went through the decoded requote path with `unsafe` now use `ignore`, which is the same path. The multiple unsafe character benchmarks from #1891 are removed since there is nothing left for them to measure.

This also removes the literal unsafe checks from the unquoting loops, which is where the `+` handling issue in #1886 and its quadratic scan came from.

No. Output of every unquoter yarl uses is unchanged; only direct use of the private `_Unquoter` with `unsafe` is affected.

No, it removes code.

## Related issue number

N/A

## Checklist

- [ ] Documentation reflects the changes (N/A, private classes)

<details>
<summary>Local verification</summary>

Output of `UNQUOTER`, the old `PATH_UNQUOTER` against the new `UNQUOTER`, `PATH_SAFE_UNQUOTER`, `QS_UNQUOTER` and `UNQUOTER_PLUS` was compared against master for both implementations over 100,000 hypothesis examples mixing text, long runs, `+`, and valid, invalid and truncated escapes, with no differences. The C and Python unquoters also agree for random `ignore`, `qs` and `plus` over 100,000 examples. `pytest tests yarl` passes with and without the extension. The Cython annotation score for `_quoting_c.pyx` goes from 821 to 737.
</details>


yarl/_quoters.py
```diff
@@ -15,8 +15,7 @@
 FRAGMENT_REQUOTER = _Quoter(safe="?/:@")
 
 UNQUOTER = _Unquoter()
-PATH_UNQUOTER = _Unquoter(unsafe="+")
```
bdraco, line 18: These always produced identical output so unsafe was truely unneeded

PR #1906, opened 2026-09-15
Skip the in-build test run for riscv64 wheels

bdraco:
## What do these changes do?

The riscv64 wheel job in the 1.25.0 release run hit the 60 minute timeout with two Python versions still to go. There is no native riscv64 runner, so building and testing both happen under QEMU; per version the build took about 3 minutes while installing the test dependencies and running the suite took about 5 more.

`CIBW_TEST_SKIP` is now set to `*` for riscv64, so those wheels are only built. The other emulated arches are unchanged and still run the suite.

No, CI only. riscv64 wheels are no longer tested at build time.

No.

## Related issue number

N/A

## Checklist

- [ ] Unit tests for the changes exist (N/A, CI configuration)
- [ ] Documentation reflects the changes (N/A)


bdraco:
I was tempted to drop the wheels since I think they were added because the goal was to get HA to support riscv64 but HA decided not to .

We have no platform specific code here so we would just be testing CPython and Cython internals so not running the same tests on riscv64 doesn't give up much.

